PR #2, opened 2021-10-21
Add result functions

cvybhu:
Adds some basic functions needed for result handling

there are now two declarations of size_t - will fix later

### DIFF
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -1,11 +1,15 @@
 use crate::argconv::*;
 use crate::cass_error::{self, CassError};
+use crate::query_result::CassResult;
 use crate::RUNTIME;
 use oneshot::{channel, Receiver};
+use scylla::QueryResult;
 use std::future::Future;
+use std::sync::Arc;
 
 pub enum CassResultValue {
     Empty,
+    QueryResult(Arc<QueryResult>),
 }
 
 pub type CassFutureResult = Result<CassResultValue, CassError>;
@@ -94,4 +98,19 @@ pub unsafe extern "C" fn cass_future_error_code(future_raw: *mut CassFuture) -> 
 #[no_mangle]
 pub unsafe extern "C" fn cass_future_free(future_raw: *mut CassFuture) {
     free_boxed(future_raw);
+}
+
+pub unsafe extern "C" fn cass_future_get_result(future_raw: *mut CassFuture) -> *const CassResult {
+    let future: &mut CassFuture = ptr_to_ref_mut(future_raw);
+    let result: &CassResultValue = match future.wait_for_result() {
+        Ok(res) => res,
+        Err(_) => return std::ptr::null(),
+    };
+
+    let query_result: Arc<QueryResult> = match result {
+        CassResultValue::QueryResult(qr) => qr.clone(),
+        _ => return std::ptr::null(), // TODO other code?
+    };
+
+    Box::into_raw(Box::new(query_result))
 }
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1,27 +1,24 @@
+use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::types::*;
-use scylla::frame::response::result::CqlValue;
+use scylla::frame::response::result::{CqlValue, Row};
 use scylla::QueryResult;
 use std::sync::Arc;
-use crate::argconv::*;
 
 pub type CassResult = Arc<QueryResult>;
 
 pub struct CassIterator {
-    #[allow(unused)]
     result: Arc<QueryResult>,
-    #[allow(unused)]
     position: usize,
 }
 
-pub struct CassRow {
-    #[allow(unused)]
-    row: scylla::frame::response::result::Row,
-}
+pub type CassRow = Row;
 
 pub type CassValue = Option<CqlValue>;
 
-pub unsafe extern "C" fn cass_iterator_from_result(result_raw: *const CassResult) -> *mut CassIterator {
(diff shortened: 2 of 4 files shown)


scylla-rust-wrapper/src/future.rs
```diff
@@ -97,3 +101,18 @@ pub extern "C" fn cass_future_free(future_raw: *mut CassFuture) {
         drop(ptr); // Explicit drop, to make function clearer
     }
 }
+
+pub extern "C" fn cass_future_get_result(future_raw: *mut CassFuture) -> *const CassResult {
+    let future: &mut CassFuture = unsafe { future_raw.as_mut().unwrap() };
```
piodul: You can use `deref_ptr_mut` here

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -0,0 +1,444 @@
...
+    let iter: &mut CassIterator = crate::utils::deref_ptr_mut(iterator);
+    iter.position += 1;
+
+    match &iter.result.rows {
+        Some(rs) => {
+            if iter.position < rs.len() {
+                1
+            } else {
+                0
+            }
```
piodul: Nit: bools should convert to integers with an `as` conversion:

```rust
(iter.position < rs.len()) as cass_bool_t
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -0,0 +1,444 @@
...
+    crate::cass_error::OK
+}
+
+pub extern "C" fn cass_value_is_null(value: *const CassValue) -> cass_bool_t {
+    let val: &CassValue = crate::utils::deref_ptr(value);
+    if val.is_none() {
+        1
+    } else {
+        0
+    }
```
piodul: Ditto

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -0,0 +1,444 @@
...
+    let val: &CassValue = crate::utils::deref_ptr(value);
+    let out: &mut cass_int32_t = crate::utils::deref_ptr_mut(output);
+
+    let cql_val: &CqlValue = match val {
+        Some(v) => v,
+        None => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
+
+    match cql_val {
+        CqlValue::Int(i) => *out = *i,
```
Lorak-mmk, line 95: Don't we need to check for integer size also?

scylla-rust-wrapper/src/utils.rs
```diff
@@ -0,0 +1,15 @@
...
+    }
+}
+
+pub fn deref_ptr<T>(ptr: *const T) -> &'static T {
+    unsafe { ptr.as_ref().unwrap() }
+}
+
+pub fn deref_ptr_mut<T>(ptr: *mut T) -> &'static mut T {
+    unsafe { ptr.as_mut().unwrap() }
+}
```
Lorak-mmk: Rebase on top of Piotr's PR to avoid duplicating util functions

Lorak-mmk (changes requested):
Please rebase on Piotr's PR, and mark user-facing functions as unsafe.

One question regarding iterators: how would they work with paged queries? In cppdriver, iterator will iterate over current page, not fetch next page automatically - you need to fetch next pages manually, there are functions for that,

scylla-rust-wrapper/src/query_result.rs
cvybhu: CqlValue has separate variants for all integer types, an int64 would be `CqlValue::BigInt` or something like that

cvybhu:
New version ready, rebased and cleaned

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    let iterator = CassIterator {
+        result: result.clone(),
+        position: 0,
+    };
+
+    Box::into_raw(Box::new(iterator))
+}
+
+// This was const for some reason, seems like a mistake in cpp driver
```
Lorak-mmk, line 33: It's not a mistake - dec_ref is marked as const, because ref counter is marked as mutable, so in C it can be const. Here we probably need to remove this const as you did.

Lorak-mmk (approved):

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -64,10 +65,18 @@ pub unsafe extern "C" fn cass_iterator_get_row(iterator: *const CassIterator) ->
...
+    index: size_t,
 ) -> *const CassValue {
-    unimplemented!();
+    let row: &CassRow = ptr_to_ref(row_raw);
+
+    let index_usize: usize = index.try_into().unwrap();
+    let column_value: &Option<CqlValue> = match row.columns.get(index_usize) {
+        Some(val) => val,
+        None => return std::ptr::null(),
+    };
```
piodul, line 77: Super tiny nit: this could be written as follows:

```rust
row.columns.get(index_usize).unwrap_or(std::ptr::null())
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -37,16 +34,33 @@ pub unsafe extern "C" fn cass_result_free(result_raw: *mut CassResult) {
...
+        .result
+        .rows
+        .as_ref()
+        .and_then(|rs| rs.get(iter.position))
+    {
+        Some(row) => row,
+        None => return std::ptr::null(),
+    };
+
+    row
```
piodul, line 64: Super tiny nit: rewrite this as follows:

```rust
iter.result.rows.as_ref().and_then(|rs| rs.get(iter.position)).unwrap_or(std::ptr::null());
```

piodul (approved):
Left some very tiny nits, looks good otherwise.

PR #3, opened 2021-10-21
Utility functions for converting C arguments

piodul:
- Adds utility functions which should simplify converting arguments from C to arguments usable by Rust
- Marks all extern functions as unsafe so that less `unsafe` blocks are needed
- Modifies existing functions so that they use the new utility functions

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -1,0 +1,26 @@
+use crate::size_t;
+use std::ffi::CStr;
+use std::os::raw::c_char;
+
+pub unsafe fn ptr_to_ref<T>(ptr: *const T) -> &'static T {
+    ptr.as_ref().unwrap()
+}
+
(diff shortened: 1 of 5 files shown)


cvybhu:
Looks good, but I don't like using `ptr_to_opt_box` to free boxed values
Maybe something like `free_boxed`? It would be clearer what's going on

cvybhu:
let's merge this before my PR, I will use those utility functions instead of my own

Lorak-mmk (approved):

piodul:
> Looks good, but I don't like using `ptr_to_opt_box` to free boxed values Maybe something like `free_boxed`? It would be clearer what's going on

OK, will change.

piodul:
@cvybhu done, changed `ptr_to_opt_box` into `free_boxed`

cvybhu (approved):

cvybhu:
LGTM, let's merge

PR #6, opened 2021-10-22
Initial support for binding values to statements (only int32)

constantfold:
First, switch to our fork of Rust Driver, as some `Clone` traits were missing for necessary types and this version supports serializing `CqlValue`.

Second, I unwrap `CassStatement` from `Arc`, as this data type doesn't need to be thread-safe. This unwrapping makes it also easier to mutate this structure.

Finally, I implemented basic version of binding values to statements. On each `CassStatement` a `Vec` of bound values is stored (initially it is `Unset`). Newly implemented `cass_statement_bind_int32` allows for changing those values. Finally `cass_session_execute` now uses those bound values.

### DIFF
--- a/scylla-rust-wrapper/Cargo.lock
+++ b/scylla-rust-wrapper/Cargo.lock
@@ -865,8 +865,7 @@ checksum = "d29ab0c6d3fc0ee92fe66e2d99f700eab17a8d57d1c1d3b748380fb20baa78cd"
 [[package]]
 name = "scylla"
 version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c19ff835deaf406d42020180b948ee5d8991d43999e10a7f1ebdbf5fe08be407"
+source = "git+https://github.com/hackathon-rust-cpp/scylla-rust-driver.git?branch=main#e5eed4bb020c363ba7aaff748ecee85ba04a6637"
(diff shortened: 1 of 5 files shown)


Lorak-mmk:
Do we really need this RWLock? CassStatement doesn't have to be thread-safe (only CassSession and CassFuture do).

constantfold:
> Do we really need this RWLock? CassStatement doesn't have to be thread-safe (only CassSession and CassFuture do).

How would I achieve it in Rust? I couldn't find a way to modify the contents of `Arc` other way.

piodul:
> > Do we really need this RWLock? CassStatement doesn't have to be thread-safe (only CassSession and CassFuture do).
> 
> How would I achieve it in Rust? I couldn't find a way to modify the contents of `Arc` other way.

First, you don't have to do `Box<Arc<CassStatement>>`. You can convert an `Arc<CassStatement>` directly to a pointer and convert it back by using `Arc::from_raw` and `Arc::to_raw`.

Second, if you don't need thread-safety of the `CassStatement`, you could just accept a `*mut CassStatement` and modify the contents.

constantfold:
v2: Unwrapped `CassStatement` from both `Arc` and `RwLock`.

Lorak-mmk (approved):

havaker (approved):

PR #8, opened 2021-10-22
Implement future callbacks

piodul:
This PR adds support for setting callbacks for the CassFutures. In order to be able to do this, the implementation of CassFuture had to be changed to use mutexes and condition variables instead of oneshot channels.


cvybhu:
Please rebase onto master to enable the CI

cvybhu (approved):
LGTM, although it got a little complicated. Keeping oneshot + mutex would probably be easier to understand, but probably less efficient.

piodul:
Rebased

PR #11, opened 2021-10-22
Prepared statements

havaker:
Implemented
```c
cass_session_prepare
cass_future_get_prepared
cass_prepared_bind
cass_prepared_free
```
Modified `cass_session_execute` to support running prepared queries.

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -2,6 +2,7 @@ use crate::argconv::*;
 use crate::cass_error;
 use crate::future::{CassFuture, CassResultValue};
 use crate::statement::CassStatement;
+use crate::statement::Statement;
 use scylla::{QueryResult, Session, SessionBuilder};
 use std::sync::Arc;
 use tokio::sync::RwLock;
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/src/session.rs
```diff
@@ -40,23 +44,60 @@ pub unsafe extern "C" fn cass_session_execute(
...
+#[no_mangle]
+pub unsafe extern "C" fn cass_session_prepare(
+    session: *mut CassSession,
+    query: *const c_char,
+) -> *const CassFuture {
+    // TODO: error handling
+    let query_str = ptr_to_cstr(query).unwrap();
+    let query_length = query_str.len();
+
+    cass_session_prepare_n(session, query, query_length as size_t)
```
cvybhu, line 74: `try_into.unwrap()` would be more correct

cvybhu (approved):
lgtm

piodul (approved):

scylla-rust-wrapper/src/session.rs
Lorak-mmk: PiotrD said "as" is better, as we don't handle errors anyway.

Lorak-mmk (approved):

PR #12, opened 2021-10-22
Improved error handling

Lorak-mmk:
Changed CassError to be enum
Implemented some error-checking instead of unwrapping.
Template implementation of error conversion - now we need to actually create correct mappings.
Implemented functions related to CassErrorResult

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -37,3 +37,8 @@ pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
 pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static str> {
     std::str::from_utf8(std::slice::from_raw_parts(ptr as *const u8, size as usize)).ok()
 }
+
+pub unsafe fn write_str_to_c(s: &str, c_str: *mut *const c_char, c_strlen: *mut size_t) {
+    *c_str = s.as_ptr() as *const i8;
+    *c_strlen = s.len() as u64;
+}
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -1,64 +1,335 @@
-pub type CassError = ::std::os::raw::c_uint;
+use scylla::transport::errors::*;
 
-pub const OK: CassError = 0;
-pub const LIB_BAD_PARAMS: CassError = 16777217;
-pub const LIB_NO_STREAMS: CassError = 16777218;
-pub const LIB_UNABLE_TO_INIT: CassError = 16777219;
-pub const LIB_MESSAGE_ENCODE: CassError = 16777220;
-pub const LIB_HOST_RESOLUTION: CassError = 16777221;
-pub const LIB_UNEXPECTED_RESPONSE: CassError = 16777222;
-pub const LIB_REQUEST_QUEUE_FULL: CassError = 16777223;
-pub const LIB_NO_AVAILABLE_IO_THREAD: CassError = 16777224;
-pub const LIB_WRITE_ERROR: CassError = 16777225;
-pub const LIB_NO_HOSTS_AVAILABLE: CassError = 16777226;
-pub const LIB_INDEX_OUT_OF_BOUNDS: CassError = 16777227;
-pub const LIB_INVALID_ITEM_COUNT: CassError = 16777228;
-pub const LIB_INVALID_VALUE_TYPE: CassError = 16777229;
-pub const LIB_REQUEST_TIMED_OUT: CassError = 16777230;
-pub const LIB_UNABLE_TO_SET_KEYSPACE: CassError = 16777231;
-pub const LIB_CALLBACK_ALREADY_SET: CassError = 16777232;
-pub const LIB_INVALID_STATEMENT_TYPE: CassError = 16777233;
-pub const LIB_NAME_DOES_NOT_EXIST: CassError = 16777234;
-pub const LIB_UNABLE_TO_DETERMINE_PROTOCOL: CassError = 16777235;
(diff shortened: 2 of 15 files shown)


scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -1,64 +1,127 @@
...
-pub const SSL_NO_PEER_CERT: CassError = 50331651;
-pub const SSL_INVALID_PEER_CERT: CassError = 50331652;
-pub const SSL_IDENTITY_MISMATCH: CassError = 50331653;
-pub const SSL_PROTOCOL_ERROR: CassError = 50331654;
-pub const SSL_CLOSED: CassError = 50331655;
-pub const LAST_ENTRY: CassError = 50331656;
+#[allow(non_camel_case_types)]
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub enum CassError {
```
constantfold: Shouldn't we specify the size of CassError explicitly (that it's `c_uint`)?

constantfold: Something like this? `#[repr(C, c_uint)]`

Lorak-mmk: I dont't think you can set 2 reprs on one enum. I can change it to `#[repr(u32)]`, but I think that `#[repr(C)]` is better - it will use best-guess representation of the same enum in C.

constantfold: That’s correct - I was under false assumption that CassError was not an enum in the C header file.

scylla-rust-wrapper/src/future.rs
```diff
@@ -139,10 +142,29 @@ pub unsafe extern "C" fn cass_future_wait(future_raw: *const CassFuture) {
...
+) -> () {
+    let message = ptr_to_ref_mut(message);
+    let message_length = ptr_to_ref_mut(message_length);
+    ptr_to_ref(future).with_waited_result(|r: &mut CassFutureResult| match r {
+        Ok(_) => {
+            *message = "".as_ptr() as *const i8;
+            *message_length = 0;
+        },
+        Err((_, s)) => {
+            *message = s.as_str().as_ptr() as *const i8;
```
constantfold: What’s the guarantee that s won’t be deallocated after this function returns?

Lorak-mmk: As far as I can see, the same as in original cpp-driver: `s` lives as long as the future.
```C
void cass_future_error_message(CassFuture* future, const char** message, size_t* message_length) {
  const Future::Error* error = future->error();
  if (error != NULL) {
    const String& m = error->message;
    *message = m.data();
    *message_length = m.length();
  } else {
    *message = "";
    *message_length = 0;
  }
}
```

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -1,64 +1,219 @@
...
+    fn from(error: &DbError) -> Self {
+        match error {
+            DbError::ServerError => CassError::SERVER_SERVER_ERROR,
+            DbError::ProtocolError => CassError::SERVER_PROTOCOL_ERROR,
+            DbError::AuthenticationError => CassError::SERVER_BAD_CREDENTIALS,
+            DbError::Unavailable {
+                consistency: _,
+                required: _,
+                alive: _,
+            } => CassError::SERVER_UNAVAILABLE,
```
piodul: You don't have to specify all those fields, the following should suffice:

```rust
DbError::Unavailable { .. } => CassError::SERVER_UNAVAILABLE,
```

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -1,64 +1,219 @@
...
+            DbError::SyntaxError => CassError::SERVER_SYNTAX_ERROR,
+            DbError::Unauthorized => CassError::SERVER_UNAUTHORIZED,
+            DbError::Invalid => CassError::SERVER_INVALID_QUERY,
+            DbError::ConfigError => CassError::SERVER_CONFIG_ERROR,
+            DbError::AlreadyExists {
+                keyspace: _,
+                table: _,
+            } => CassError::SERVER_ALREADY_EXISTS,
+            DbError::Unprepared => CassError::SERVER_UNPREPARED,
+            DbError::Other(num) => unsafe { mem::transmute(*num) },
```
piodul: I know we are generally using unsafe, but I think that using transmute here is too much. I think that we should either:

1. drop precise information about the error and use an existing variant in its place (e.g. `SERVER_SERVER_ERROR`),
1. or introduce corresponding `Other(i32)` variant in the `CassError` enum - and maybe implement `Into<c_int>` etc. and use it every time we return the error.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -32,3 +32,8 @@ pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
 pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static str> {
     std::str::from_utf8(std::slice::from_raw_parts(ptr as *const u8, size as usize)).ok()
 }
+
+pub unsafe fn write_str_to_c(s: &'static str, c_str: *mut *const c_char, c_strlen: *mut size_t) {
+    *c_str = s.as_ptr() as *const i8;
+    *c_strlen = s.len() as u64;
+}
```
piodul: A tiny nitpick: I think it would be better if you specify a lifetime explicitly here, like this:

```rust
pub unsafe fn write_str_to_c<'a>(s: &'a str, c_str: *mut *const c_char, c_strlen: *mut size_t) {
    *c_str = s.as_ptr() as *const i8;
    *c_strlen = s.len() as u64;
}
```

It will allow using this function with string slices with temporary lifetime (which is what we do, but we also give them `'static` lifetime, which is something that we might try to fix in the future)

piodul (changes requested):

piodul (approved):

PR #13, opened 2021-10-22
First portion of cass_cluster_* implementations

constantfold:
#### Properly implement setting contact points
The previous implementation only supported adding a single contact 
point. This commit allows to pass multiple contact points (delimited
by ',' as in Cpp Driver). cass_cluster_set_contact_points_n is 
implemented additionally.

In order to support Cpp Driver's setting of port (which can happen
anytime), the contact points are stored in CassCluster. Just before
the session is built, build_session_builder "merges" the contact point
list together with last set port (I believe this is how Cpp Driver 
works).

Some minor type problems are fixed in session.rs: previously it
reinterpreted CassCluster pointer as SessionBuilder pointer (which
worked as CassCluster consisted only of SessionBuilder).

#### Add support for setting load balancing strategies
Implement most of the methods related to setting different load 
balancing strategies.

#### Add cass_cluster_set_credentials

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -3,35 +3,97 @@ use crate::cass_error::{self, CassError};
 use scylla::SessionBuilder;
 use std::os::raw::c_char;
 
-pub struct CassCluster(pub SessionBuilder);
+pub struct CassCluster {
+    session_builder: SessionBuilder,
 
-#[no_mangle]
-pub unsafe extern "C" fn cass_cluster_new() -> *mut CassCluster {
-    Box::into_raw(Box::new(CassCluster(SessionBuilder::new())))
+    contact_points: Vec<String>,
+    port: u16,
+}
+
+pub fn build_session_builder(cluster: &CassCluster) -> SessionBuilder {
+    let known_nodes: Vec<_> = cluster
+        .contact_points
+        .clone()
+        .into_iter()
+        .map(|cp| format!("{}:{}", cp, cluster.port))
+        .collect();
+    cluster
+        .session_builder
+        .clone()
+        .known_nodes(&known_nodes)
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn cass_cluster_free(session_builder_raw: *mut CassCluster) {
-    free_boxed(session_builder_raw);
+pub unsafe extern "C" fn cass_cluster_new() -> *mut CassCluster {
+    Box::into_raw(Box::new(CassCluster {
+        session_builder: SessionBuilder::new(),
+        port: 9042,
+        contact_points: Vec::new(),
+    }))
+}
+
+#[no_mangle]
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1,37 +1,250 @@
...
+        // set of contact points, the contact points should be cleared.
+        cluster.contact_points.clear();
+        return Ok(());
+    }
+
+    // cass_cluster_set_contact_points() will append
+    // in subsequent calls, not overwrite.
+    cluster
+        .contact_points
+        .extend(contact_points.map(|cp| cp.to_string()));
```
Lorak-mmk: If we want to have the same behaviour as cpp-driver, the the strings should have whitespace stripped. Relevant code from cpp-driver:
```
void explode(const String& str, Vector<String>& vec, const char delimiter /* = ',' */) {
  IStringStream stream(str);
  while (!stream.eof()) {
    String token;
    std::getline(stream, token, delimiter);
    if (!trim(token).empty()) {
      vec.push_back(token);
    }
  }
}
```

Lorak-mmk: Also, only non-empty strings should be added to contact points.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1,37 +1,250 @@
...
-    session_builder_raw: *mut CassCluster,
-    uri_raw: *const c_char,
+    cluster_raw: *mut CassCluster,
+    contact_points_raw: *const c_char,
+    contact_points_length: size_t,
 ) -> Result<(), CassError> {
-    let session_builder = ptr_to_ref_mut(session_builder_raw);
-    let uri = ptr_to_cstr(uri_raw).ok_or(cass_error::LIB_BAD_PARAMS)?;
-    session_builder.0.config.add_known_node(uri);
+    // FIXME: validate contact points (whether they are valid inets)
```
Lorak-mmk: I don't think we have to do that - cpp-driver doesn't. The Address constructor called for every part there looks like this:
```
Address::Address(const String& hostname, int port, const String& server_name)
    : server_name_(server_name)
    , family_(UNRESOLVED)
    , port_(port) {
  char addr[16];
  if (uv_inet_pton(AF_INET, hostname.c_str(), addr) == 0) {
    hostname_or_address_.assign(addr, addr + 4);
    family_ = IPv4;
  } else if (uv_inet_pton(AF_INET6, hostname.c_str(), addr) == 0) {
    hostname_or_address_.assign(addr, addr + 16);
    family_ = IPv6;
  } else {
    hostname_or_address_ = hostname;
  }
}
```
So no matter what is the string, no error will be returned, and addresses will be validated at connect time, which I think is the same behaviour as with our current code.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cluster.rs
constantfold: Added trimming of split strings and skipping those that are empty.

PR #14, opened 2021-10-22
Make clippy CI actually fail

constantfold:
With `--examples` flag, even though clippy found some warnings, it didn't fail the build as it was only executed on "examples" (This seems like a mistake?).


constantfold:
@cvybhu Is this what you actually intended to do? The way it was previously, it was pretty useless...

Lorak-mmk (approved):

PR #15, opened 2021-10-22
Add *_by_name and *_by_name_n variants of statement bind functions

constantfold:
Add *_by_name and *_by_name_n variants of statement bind functions, which work for prepared statements and allow to set a value based on the column name. 

Fix cass_prepared_bind to correctly initialize bound_values based on the received metadata. 

Small refactoring around CassCollection to CqlValue conversion.

### DIFF
--- a/scylla-rust-wrapper/src/collection.rs
+++ b/scylla-rust-wrapper/src/collection.rs
@@ -5,6 +5,7 @@ use scylla::frame::response::result::CqlValue;
 use scylla::frame::response::result::CqlValue::*;
 use std::os::raw::c_char;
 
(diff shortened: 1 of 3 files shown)


constantfold:
On the second look, it probably would be nicer using Rust macros rather than copy-pasta.

piodul:
I agree that using macros for reducing boilerplate would be desirable here. I can try doing that.

piodul:
Pushed an updated version with reduced boilerplate via macros

piodul:
@avelanarius @Lorak-mmk please review

Lorak-mmk (approved):

PR #16, opened 2021-10-22
First portion of cass_statement_set_* functions

constantfold:
Implement a few easy to write cass_statement_set_* functions. Modify main.c with an example use-case.

### DIFF
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -8,7 +8,7 @@ use scylla::frame::value::MaybeUnset;
 use scylla::frame::value::MaybeUnset::{Set, Unset};
 use scylla::query::Query;
 use scylla::statement::prepared_statement::PreparedStatement;
-use std::os::raw::c_char;
+use std::os::raw::{c_char, c_int};
 use std::sync::Arc;
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


Lorak-mmk (approved):

havaker:
Rust driver's prepared statements API was designed to fit a use case, in which user configures a PreparedStatement, and then executes it multiple times without modifying its config.  Cloning PreparedStatement before each execution could unnecessarily degrade this wrapper's performance, because [PreparedStatement](https://github.com/scylladb/scylla-rust-driver/blob/bd74bff4db234d9d6b7580791870281a59e49759/scylla/src/statement/prepared_statement.rs#L14) can be quite heavy (it contains [PreparedMetadata](https://github.com/scylladb/scylla-rust-driver/blob/bd74bff4db234d9d6b7580791870281a59e49759/scylla/src/frame/response/result.rs#L310)).

I suggest leaving PreparedStatement wrapped in Arc, and using [`Arc::make_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) method to create a clone-on-write mechanism:
```rust
Statement::Prepared(inner) => {
    let inner_cow: &mut PreparedStatement = Arc::make_mut(inner);
    inner_cow.set_is_idempotent(is_idempotent != 0)
}
```
This solution would reduce cloning to situations, when a statement is currently being executed and  `cass_statement_set*` is concurrently called on it.

Lorak-mmk:
> Rust driver's prepared statements API was designed to fit a use case, in which user configures a PreparedStatement, and then executes it multiple times without modifying its config. Cloning PreparedStatement before each execution could unnecessarily degrade this wrapper's performance, because [PreparedStatement](https://github.com/scylladb/scylla-rust-driver/blob/bd74bff4db234d9d6b7580791870281a59e49759/scylla/src/statement/prepared_statement.rs#L14) can be quite heavy (it contains [PreparedMetadata](https://github.com/scylladb/scylla-rust-driver/blob/bd74bff4db234d9d6b7580791870281a59e49759/scylla/src/frame/response/result.rs#L310)).
> 
> I suggest leaving PreparedStatement wrapped in Arc, and using [`Arc::make_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) method to create a clone-on-write mechanism:
> 
> ```rust
> Statement::Prepared(inner) => {
>     let inner_cow: &mut PreparedStatement = Arc::make_mut(inner);
>     inner_cow.set_is_idempotent(is_idempotent != 0)
> }
> ```
> 
> This solution would reduce cloning to situations, when a statement is currently being executed and `cass_statement_set*` is concurrently called on it.

Am I correct in assuming that the metadata, statement, page_size and id are never changed? If so, then would it be a breaking change to the driver if those fields were an Arc'ed struct, allowing faster clones? I can see that those fields are not pub, so the user doesn't have access to them?

constantfold:
v2: removed commit removing `Arc`. Implemented setters now use `Arc::make_mut()`.

havaker (approved):

PR #17, opened 2021-10-22
Implement collection types

constantfold:
Implement collection types: `MAP`, `LIST`, `SET` (`CassCollection` type).

Implement `cass_statement_bind_collection()`.

### DIFF
--- a/scylla-rust-wrapper/src/collection.rs
+++ b/scylla-rust-wrapper/src/collection.rs
@@ -1,0 +1,153 @@
+use crate::argconv::*;
+use crate::cass_error::CassError;
+use crate::types::*;
+use scylla::frame::response::result::CqlValue;
+use scylla::frame::response::result::CqlValue::*;
+use std::os::raw::c_char;
+
+pub struct CassCollection {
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/collection.rs
```diff
@@ -0,0 +1,153 @@
...
+        // Maps consist of a key and a value, so twice
+        // the number of CqlValue will be stored.
+        CassCollectionType::CASS_COLLECTION_TYPE_MAP => item_count * 2,
+        _ => item_count,
+    };
+
+    Box::into_raw(Box::new(CassCollection {
+        collection_type,
+        capacity: item_count as usize,
+        items: Vec::new(),
```
Lorak-mmk: If the user has to supply item count, then maybe we could use it to create vector with correct capacity?

Lorak-mmk (approved):

scylla-rust-wrapper/src/collection.rs
constantfold: Changed to `Vec::with_capacity()`.

PR #18, opened 2021-10-22
Implement CassDataType, CassValueType and CassUserType

constantfold:
The first commit implements `CassDataType`, `CassValueType` and a portion of related `cass_data_type*` functions.

The second commit implements `CassUserType` and a portion of related `cass_user_type*` functions.

The third commit implements `cass_statement_bind_user_type` and adds an example UDT insert to `main.c`.

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -1,0 +1,230 @@
+use crate::argconv::*;
+use crate::cass_error::CassError;
+use crate::types::*;
+use std::os::raw::c_char;
+
+#[allow(non_camel_case_types)]
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub enum CassValueType {
+    CASS_VALUE_TYPE_UNKNOWN = 65535,
+    CASS_VALUE_TYPE_CUSTOM = 0,
+    CASS_VALUE_TYPE_ASCII = 1,
+    CASS_VALUE_TYPE_BIGINT = 2,
+    CASS_VALUE_TYPE_BLOB = 3,
+    CASS_VALUE_TYPE_BOOLEAN = 4,
+    CASS_VALUE_TYPE_COUNTER = 5,
+    CASS_VALUE_TYPE_DECIMAL = 6,
+    CASS_VALUE_TYPE_DOUBLE = 7,
+    CASS_VALUE_TYPE_FLOAT = 8,
+    CASS_VALUE_TYPE_INT = 9,
+    CASS_VALUE_TYPE_TEXT = 10,
+    CASS_VALUE_TYPE_TIMESTAMP = 11,
+    CASS_VALUE_TYPE_UUID = 12,
+    CASS_VALUE_TYPE_VARCHAR = 13,
+    CASS_VALUE_TYPE_VARINT = 14,
+    CASS_VALUE_TYPE_TIMEUUID = 15,
+    CASS_VALUE_TYPE_INET = 16,
+    CASS_VALUE_TYPE_DATE = 17,
+    CASS_VALUE_TYPE_TIME = 18,
+    CASS_VALUE_TYPE_SMALL_INT = 19,
+    CASS_VALUE_TYPE_TINY_INT = 20,
+    CASS_VALUE_TYPE_DURATION = 21,
+    CASS_VALUE_TYPE_LIST = 32,
+    CASS_VALUE_TYPE_MAP = 33,
+    CASS_VALUE_TYPE_SET = 34,
+    CASS_VALUE_TYPE_UDT = 48,
+    CASS_VALUE_TYPE_TUPLE = 49,
+    CASS_VALUE_TYPE_LAST_ENTRY = 50,
+}
+
+#[derive(Clone)]
+pub struct UDTDataType {
+    pub field_count: usize,
+
+    // Vec to preserve the order of types
+    pub field_types: Vec<(String, CassDataType)>,
+
+    pub keyspace: String,
+    pub name: String,
+}
+
+#[derive(Clone)]
+pub enum CassDataType {
+    ValueDataType(CassValueType),
+    UDTDataType(UDTDataType),
+}
+
+#[no_mangle]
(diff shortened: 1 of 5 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+#[derive(Clone)]
+pub enum CassDataType {
+    ValueDataType(CassValueType),
+    UDTDataType(UDTDataType),
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_data_type_new(value_type: CassValueType) -> *mut CassDataType {
```
piodul, line 59: The `CassValueType` used here looks a bit scary. I don't know if it is defined behavior if e.g. the user passes `123456` here, which does not correspond to any valid variant of the enum. I would recommend explicitly checking if the integral value passed to Rust corresponds to a valid `CassValueType`. You can use the [`num_enum`](https://crates.io/crates/num_enum) crate to generate the boilerplate for you.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -0,0 +1,230 @@
...
+    CASS_VALUE_TYPE_TIME = 18,
+    CASS_VALUE_TYPE_SMALL_INT = 19,
+    CASS_VALUE_TYPE_TINY_INT = 20,
+    CASS_VALUE_TYPE_DURATION = 21,
+    CASS_VALUE_TYPE_LIST = 32,
+    CASS_VALUE_TYPE_MAP = 33,
+    CASS_VALUE_TYPE_SET = 34,
+    CASS_VALUE_TYPE_UDT = 48,
+    CASS_VALUE_TYPE_TUPLE = 49,
+    CASS_VALUE_TYPE_LAST_ENTRY = 50,
```
piodul, line 38: I don't think we need this variant - it's not a real value type.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -232,6 +232,26 @@ pub unsafe extern "C" fn cass_statement_bind_collection(
...
+    // FIXME: implement _by_name and _by_name_n variants
+    let user_type = ptr_to_ref(user_type_raw);
+
+    cass_statement_bind_cql_value(
+        statement,
+        index,
+        CqlValue::UserDefinedType {
+            keyspace: user_type.udt_data_type.keyspace.clone(),
+            type_name: user_type.udt_data_type.name.clone(),
+            fields: user_type.field_values.clone().into_iter().collect(),
```
piodul, line 250: With the master version of scylla-rust-driver, just cloning should be sufficient (`fields` has the same type as `user_type.field_values`).

scylla-rust-wrapper/src/cass_types.rs
piodul: Besides, I think that the original cpp driver returns a null pointer in case `value_type` is not a valid type (in case of UNKNOWN it does it, too).

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -0,0 +1,320 @@
...
+            // of the correct type as far as I checked.
+            *field_value = value;
+            return crate::cass_error::OK;
+        }
+    }
+
+    crate::cass_error::LIB_NAME_DOES_NOT_EXIST
+}
+
+unsafe fn cass_user_type_set_cql_value_by_name(
```
piodul, line 73: I think we could reduce the boilerplate of the `cass_user_type_set_xyz` functions with the help of macros, but it can be done in a followup.

piodul (changes requested):

scylla-rust-wrapper/src/cass_types.rs
havaker: I think it's safer to leave it. This enum has a `repr(C)`, it assumes that enum on a C side has same variants, and chooses enum's representation based on it.

scylla-rust-wrapper/src/statement.rs
constantfold: Yes, this is remnant of my older version...

scylla-rust-wrapper/src/cass_types.rs
piodul: Removing this field won't affect the enum's representation because it's not the variant with the biggest value (the biggest is `CASS_VALUE_TYPE_UNKNOWN`). The enum size is chosen based on the biggest variant.

That said, I'm not sure if we should put rust enums on a FFI boundary directly because somebody could pass an invalid enum variant and it would be undefined behavior in rust:

> Fieldless enums with repr(C) or repr(u*) still may not be set to an integer value without a corresponding variant, even though this is permitted behavior in C or C++. It is undefined behavior to (unsafely) construct an instance of an enum that does not match one of its variants. (This allows exhaustive matches to continue to be written and compiled as normal.)

Source: https://doc.rust-lang.org/nomicon/other-reprs.html#reprc

I suggested using `num_enum` crate in a previous comment in order to avoid this issue.

scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk: I opened a PR that generates enum at compile time - and those enums are implemented as wrappers for integer types, with associated constants - then we can then accept such enum as argument, and any match on it simply won't be exhaustive until we handle case of invalid value. So I suggest merging my PR, then rebasing this one to use the same approach.

scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk: This variant will remain, as enum is now auto generated. Also, even if it was copied, it is actually used in cpp driver, so we probably would to also use it at some point:
```
    default:
      if (type < CASS_VALUE_TYPE_LAST_ENTRY) {
        data_type = new DataType(type);
        data_type->inc_ref();
      }
      break;
```

Lorak-mmk:
Closing in favour of https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/60

PR #19, opened 2021-10-24
Implement a portion of UUID-related functions

constantfold:
Implement the `CassUuid` type, a portion of `cass_uuid*` functions (some directly ported from Cpp Driver code), `cass_value_get_uuid` and `cass_statement_bind_uuid`.

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1,6 +1,7 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::types::*;
+use crate::uuid::CassUuid;
 use scylla::frame::response::result::{CqlValue, Row};
 use scylla::QueryResult;
 use std::convert::TryInto;
@@ -113,6 +114,27 @@ pub unsafe extern "C" fn cass_value_get_int32(
 
     match cql_val {
         CqlValue::Int(i) => *out = *i,
+        _ => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
+
+    crate::cass_error::OK
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_get_uuid(
+    value: *const CassValue,
+    output: *mut CassUuid,
+) -> CassError {
+    let val: &CassValue = ptr_to_ref(value);
+    let out: &mut CassUuid = ptr_to_ref_mut(output);
+
+    let cql_val: &CqlValue = match val {
+        Some(v) => v,
+        None => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
+
+    match cql_val {
+        CqlValue::Uuid(uuid) => *out = (*uuid).into(),
         _ => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
     };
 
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -119,6 +120,27 @@ pub unsafe extern "C" fn cass_value_get_int32(
...
+
+    let cql_val: &CqlValue = match val {
+        Some(v) => v,
+        None => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
+
+    match cql_val {
+        CqlValue::Uuid(uuid) => *out = (*uuid).into(),
+        _ => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
```
piodul, line 139: You could make it shorter, with only one `match`:

```rust
match val {
    Some(CqlValue::Uuid(uuid)) => *out = (*uuid).into(),
    _ => return crate::cass_error::LIB_BAD_PARAMS,
}
```

scylla-rust-wrapper/src/uuid.rs
```diff
@@ -0,0 +1,145 @@
...
+        // employs. "Recovered" and validated it empirically...
+        let msb = ((upper_u64 & 0xFFFFFFFF) as u32).rotate_left(16) as u64;
+        let lsb = (upper_u64 & 0xFFFFFFFF00000000) >> 32;
+
+        CassUuid {
+            time_and_version: (msb << 32) | lsb,
+            clock_seq_and_node: lower_u64,
+        }
+    }
+}
```
piodul: The logic for UUID serialization/deserialization in cpp driver can be found in the `src/serialization.hpp` file. It's a bit different than what you wrote here - but it seems to do the same thing. In the future we could consider converting `Uuid` to bytes and then parsing with a very similar logic, and doing the reverse in case of conversion in the other direction.

scylla-rust-wrapper/src/uuid.rs
```diff
@@ -0,0 +1,145 @@
...
+        let msb = ((time_and_version_msb << 32) | time_and_version_lsb) as u128;
+        let lsb = uuid.clock_seq_and_node as u128;
+
+        Uuid::from_u128((msb << 64) | lsb)
+    }
+}
+
+impl From<Uuid> for CassUuid {
+    fn from(uuid: Uuid) -> Self {
+        let u128_representation = u128::swap_bytes(uuid.to_u128_le());
```
piodul: I think this is equivalent to `let u128_representation = uuid.as_u128();` ([link to the relevant source](https://docs.rs/uuid/0.8.2/src/uuid/lib.rs.html#455-513)).

piodul (approved):
Left some minor comments, LGTM otherwise

scylla-rust-wrapper/src/query_result.rs
constantfold: This is a copy-pasta from the previous function, it will be addressed by @jul-stas when he implements `cass_value_get*` family (and hopefully copy-pastas the right "clean" version).

scylla-rust-wrapper/src/uuid.rs
constantfold: I wanted to rely as much as possible on uuid library implementation - not reimplementing serialization/deserialization myself (apart from those one-liner functions above). Unfortunately, the internal representation of driver needs a few adjustments compared to uuid library representation.

scylla-rust-wrapper/src/uuid.rs
constantfold: Fixed

PR #20, opened 2021-10-24
Implement CassInet and related functions

constantfold:
Implement `CassInet`, all `cass_inet*` functions and statement binding of `CassInet`.

### DIFF
--- a/scylla-rust-wrapper/src/inet.rs
+++ b/scylla-rust-wrapper/src/inet.rs
@@ -1,0 +1,144 @@
+use crate::argconv::*;
+use crate::cass_error::CassError;
+use crate::types::*;
+use std::convert::TryInto;
+use std::net::IpAddr;
+use std::os::raw::c_char;
+use std::slice::from_raw_parts;
+use std::str::FromStr;
+
+#[repr(u8)] // address_length field in CassInet is cass_uint8_t
+#[allow(non_camel_case_types)]
+#[derive(Debug, Copy, Clone)]
+pub enum CassInetLength {
+    CASS_INET_V4 = 4,
+    CASS_INET_V6 = 16,
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct CassInet {
+    pub address: [cass_uint8_t; 16],
+    pub address_length: CassInetLength,
+}
+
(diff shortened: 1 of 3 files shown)


scylla-rust-wrapper/src/inet.rs
```diff
@@ -0,0 +1,144 @@
...
+    let input = ptr_to_cstr_n(input_raw, input_length);
+    if input.is_none() {
+        return crate::cass_error::LIB_BAD_PARAMS;
+    }
+
+    let input_str = input.unwrap();
+    let ip_addr = IpAddr::from_str(input_str);
+
+    match ip_addr {
+        Result::Ok(ip_addr) => {
```
piodul: No need to use qualified `Result::Ok` or `Result::Err`, those are imported by Rust by default. Use `Ok` and `Err` instead.

scylla-rust-wrapper/src/inet.rs
```diff
@@ -0,0 +1,144 @@
+use crate::argconv::*;
+use crate::cass_error::CassError;
```
piodul, line 2: You can change it to the following:

```rust
use crate::cass_error::{self, CassError};
```

Then you will be able to change occurrences of `crate::cass_error::ERROR_NAME` to `cass_error::ERROR_NAME` in this file.

piodul (approved):

scylla-rust-wrapper/src/inet.rs
constantfold: I was just too cautious as when I tried this:

```
enum MyEnum {
A, B
}
...
match(my_enum_instance) {
  A => {},
  B => {},
}
```

it did not do what I expect and had to write `MyEnum::A`, `MyEnum::B` in the `match`.

scylla-rust-wrapper/src/inet.rs
constantfold: Purposefully did not make those occurrences "clean", as #12 changes the error type and all those instances will be changed either way after that PR is merged.

scylla-rust-wrapper/src/inet.rs
constantfold: Fixed

PR #21, opened 2021-10-24
Implement a few cass_cluster_set functions

constantfold:
Just another batch of easy-to-implement `cass_cluster_set*` functions.

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -5,6 +5,7 @@ use core::time::Duration;
 use scylla::load_balancing::{
     DcAwareRoundRobinPolicy, LoadBalancingPolicy, RoundRobinPolicy, TokenAwarePolicy,
 };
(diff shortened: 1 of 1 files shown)


Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -22,16 +25,22 @@ pub struct CassCluster {
...
-    let known_nodes: Vec<_> = cluster
+    let mut known_nodes: Vec<_> = cluster
         .contact_points
         .clone()
         .into_iter()
         .map(|cp| format!("{}:{}", cp, cluster.port))
         .collect();
 
+    if cluster.randomize_contact_points {
+        known_nodes.shuffle(&mut thread_rng());
```
piodul: The rust driver does this internally, and there is no way to switch it off. Maybe the `cass_cluster_set_use_randomized_contact_points` function should be a no-op?

constantfold: Per discussion on Slack, for the time being, I removed this function.

PR #23, opened 2021-10-25
Remove reduntant Box in CassSession and CassResult

Lorak-mmk:
Currently CassSession and CassResult are Arc<...> but are still returned to user using `Box::into_raw(Box::new(...))`. This PR removes this redundant Box and instead uses `Arc::into_raw`.

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -12,11 +12,13 @@ use std::os::raw::c_char;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
-type CassSession = Arc<RwLock<Option<Session>>>;
+type CassSession = RwLock<Option<Session>>;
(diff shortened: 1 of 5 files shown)


scylla-rust-wrapper/src/session.rs
```diff
@@ -25,11 +27,11 @@ pub unsafe extern "C" fn cass_session_connect(
     cluster_raw: *const CassCluster,
 ) -> *const CassFuture {
     let session_opt = ptr_to_ref(session_raw);
-    let cluster = ptr_to_ref(cluster_raw);
+    let cluster: CassCluster = (*ptr_to_ref(cluster_raw)).clone();
```
piodul, line 30: This clones the whole CassCluster object, right? As a future improvement, maybe we could keep it behind an Arc and call Arc::make_mut every time we want to modify it.

piodul (approved):

PR #24, opened 2021-10-25
CassValue: add getters for integer and float types

jul-stas:
I also prepared indentation for other getters' prototypes.

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -101,21 +101,128 @@ pub unsafe extern "C" fn cass_row_get_column(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn cass_value_get_float(
+    value: *const CassValue,
+    output: *mut cass_float_t,
+) -> CassError {
+    let val: &CassValue = ptr_to_ref(value);
+    let out: &mut cass_float_t = ptr_to_ref_mut(output);
+    match val {
+        Some(CqlValue::Float(f)) => *out = *f,
+        Some(_) => return crate::cass_error::LIB_INVALID_VALUE_TYPE,
+        None => return crate::cass_error::LIB_NULL_VALUE,
+    };
+
+    crate::cass_error::OK
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_get_double(
+    value: *const CassValue,
+    output: *mut cass_double_t,
+) -> CassError {
+    let val: &CassValue = ptr_to_ref(value);
+    let out: &mut cass_double_t = ptr_to_ref_mut(output);
+    match val {
+        Some(CqlValue::Double(d)) => *out = *d,
+        Some(_) => return crate::cass_error::LIB_INVALID_VALUE_TYPE,
+        None => return crate::cass_error::LIB_NULL_VALUE,
+    };
+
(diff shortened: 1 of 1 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -100,6 +100,132 @@ pub unsafe extern "C" fn cass_row_get_column(
...
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_get_float(
+    value: *const CassValue,
+    output: *mut cass_float_t,
+) -> CassError {
+    let val: &CassValue = ptr_to_ref(value);
+    let out: &mut cass_float_t = ptr_to_ref_mut(output);
+
+    let cql_val: &CqlValue = match val {
```
constantfold: This could be made shorter as in: https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/19#discussion_r735494920

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -100,6 +100,132 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+        Some(v) => v,
+        None => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
```
constantfold: Maybe now's the time to finally check what's the correct error code? (Since we are doing copy-pasta)

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -100,6 +100,132 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+    let val: &CassValue = ptr_to_ref(value);
+    let out: &mut cass_bool_t = ptr_to_ref_mut(output);
+
+    let cql_val: &CqlValue = match val {
+        Some(v) => v,
+        None => return crate::cass_error::LIB_BAD_PARAMS, // TODO: other error code?
+    };
+
+    match cql_val {
+        CqlValue::Boolean(b) => *out = if *b { 1 } else { 0 },
```
constantfold: In other code I used `*b != 0` - this is what Rust error message suggested

scylla-rust-wrapper/src/query_result.rs
jul-stas: Nice, but I left it as is to stay true to the cpp-driver's error messages (see below)

scylla-rust-wrapper/src/query_result.rs
jul-stas: Done

constantfold (approved):

scylla-rust-wrapper/src/query_result.rs
piodul: How about: `*out = *b as cass_bool_t`

jul-stas: done

scylla-rust-wrapper/src/query_result.rs
constantfold: I believe it could written this way:

```rust
match val {
    Some(CqlValue::Uuid(uuid)) => *out = (*uuid).into(),
    Some(_) => return crate::cass_error::LIB_INVALID_VALUE_TYPE,
    None => return crate::cass_error::LIB_NULL_VALUE,
}
```

PR #25, opened 2021-10-25
Add support for paging

constantfold:
Implement `cass_result_has_more_pages` and `cass_statement_set_paging_state`. In `cass_session_execute` invoke `_paged` versions of `execute` and `query`. Fix the default paging setting for queries (it's disabled in Cpp Driver). Add an example of paging to `main.c`.


piodul (commented):
LGTM, but needs a rebase first.

PR #27, opened 2021-10-25
Implement cass_value_get_string

Lorak-mmk:


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -244,6 +244,27 @@ pub unsafe extern "C" fn cass_value_get_uuid(
     CassError::CASS_OK
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_get_string(
+    value: *const CassValue,
+    output: *mut *const ::std::os::raw::c_char,
```
constantfold:
```suggestion
    output: *mut *const c_char,
```

constantfold (approved):

jul-stas:
I piggybacked with `cass_value_get_inet`

scylla-rust-wrapper/src/query_result.rs
jul-stas: done

PR #28, opened 2021-10-25
Set default consistency level to ONE

constantfold:
Set the default consistency level for queries to ONE, as is the case in the Cpp Driver.


Lorak-mmk (approved):

Lorak-mmk:
LGTM

PR #29, opened 2021-10-25
Fixed error code in futures

Lorak-mmk:
Return correct error code from futures


constantfold (approved):
You should probably fix the commit message: "fized lifetime issue in get_error_message"

piodul (approved):

PR #31, opened 2021-10-25
Overhaul of CassResult* structures for cass_row_get_column_by_name*

constantfold:
PR code-complete, needs polishing

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -8,6 +8,7 @@ use scylla::Bytes;
 use std::convert::TryInto;
 use std::os::raw::c_char;
 use std::sync::Arc;
+use scylla::frame::response::result::ColumnSpec;
 
 pub struct CassResult {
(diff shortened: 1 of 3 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -100,14 +113,49 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+        name_str = name_str.strip_suffix("\"").unwrap().to_string();
+    }
+
+    let row = ptr_to_ref(row);
+    let col_specs = &row.result_metadata.col_specs;
+
+    println!("{:?} query: {:?}", col_specs, name_str);
+
+    col_specs.iter().enumerate().filter(|(_, spec)| spec.name.to_lowercase() == name_str).next().map(|(index, _)| {
+        cass_row_get_column(row, index as size_t)
```
Lorak-mmk, line 155: So on each call to cass_row_get_column_by_name we iterate over all columns and search for right column name? I think it would be better to create map str -> index when creating CassResult, and get right index using this map.

PR #32, opened 2021-10-26
Initial batches support

piodul:


scylla-rust-wrapper/src/batch.rs
```diff
@@ -0,0 +1,189 @@
...
+pub unsafe extern "C" fn cass_batch_free(batch: *mut CassBatch) {
+    free_boxed(batch)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_batch_set_keyspace(
+    _batch: *mut CassBatch,
+    _keyspace: *const c_char,
+) -> CassError {
+    // XXX: Not supported by v4 of protocol
```
Lorak-mmk, line 50: Then maybe it would be best to return some error, not CASS_OK?

constantfold: In my opinion, this function should be left unimplemented. (The same reasoning as before:) When some user uses this function they will get a linking error and will have to modify their program. But when it is implemented, even returning a proper error code, there's a high probability that it will go unnoticed (even DataStax's example programs ignore a lot of errors of these kinds of functions).

scylla-rust-wrapper/src/batch.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub batch: Batch,
+    pub bound_values: Vec<Vec<MaybeUnset<Option<CqlValue>>>>,
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_batch_new(type_: CassBatchType) -> *mut CassBatch {
+    if let Some(batch_type) = types::make_batch_type(type_) {
+        Box::into_raw(Box::new(CassBatch {
+            state: Arc::new(CassBatchState {
+                batch: Batch::new(batch_type),
```
constantfold, line 30: You should also set the consistency level of the created batch to `ONE` (this is what Cpp Driver does, while Rust Driver default is to use QUORUM).

constantfold: See https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/28

constantfold:
Closing as newer version was merged (https://github.com/scylladb/cpp-rust-driver/pull/71)

PR #56, opened 2021-11-19
Generate enums

### LINKED ISSUE #26: Use bindgen to generate Cass* enums instead of copying them maunally.
Currently, we used bindgen to generate Cass* enums definitions, then copied them to our code. After hackathon, we should use definitions generated by bindgen by importing them

Lorak-mmk:
Fixes #26 
Use bindgen to generate cppdriver enums at compile time (instead of copying bindgen results into source code).


scylla-rust-wrapper/src/inet.rs
```diff
@@ -97,21 +95,22 @@ pub unsafe extern "C" fn cass_inet_from_string_n(
...
             }
-            CassInetLength::CASS_INET_V6 => {
+            Some(CassInetLength::CASS_INET_V6) => {
                 let addr_bytes: [cass_uint8_t; CassInetLength::CASS_INET_V6 as usize] = inet
                     .address[0..(CassInetLength::CASS_INET_V6 as usize)]
                     .try_into()
                     .unwrap();
                 IpAddr::V6(addr_bytes.into())
             }
+            None => panic!("Encountered CassInet with unknown type!"),
```
piodul: What is the behavior of the cpp driver in case address_length is invalid? I see that this field is not hidden in any way in the cassandra.h header, so I think we should be prepare for such a situation.

You can use the [`std::convert::TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) trait to implement fallible conversions.

scylla-rust-wrapper/src/collection.rs
```diff
@@ -171,6 +164,7 @@ impl From<CassCollection> for CqlValue {
                 Map(grouped_items)
             }
             CassCollectionType::CASS_COLLECTION_TYPE_SET => CqlValue::Set(collection.items),
+            _ => panic!("Collection with invalid type encountered!"),
```
piodul: I'd suggest changing this conversion function to use TryFrom, too.

scylla-rust-wrapper/src/inet.rs
Lorak-mmk: > What is the behavior of the cpp driver in case address_length is invalid?

It doesn't check this at all, and will happily use CassInet with invalid length in statement... I don't think that's the approach we want to take (even if we did, I don't think Rust driver supports this), so I'll change this to use TryFrom, and return some error in cass_statement_bind_ if length is invalid.

piodul (approved):

PR #57, opened 2021-12-01
Cass types v2

Lorak-mmk:
This is a continuation of https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/18
Changes from original PR:
- Derive PartialOrd and PartialEq for generated enums
- Reorder functions in cass_types to the same order as in spreadsheet.
- Reimplement cass_data_type_new - it's behaviour should depend on argument, as in cppdriver
- Add strlen helper function
- Implement missing variants of CassDataType
- Implement missing function of CassDataType - apart from is_frozen (I can't see a scenario where it returns true)
- CassDataType is now Arc, otherwise it's impossible to be semantically compatible with cppdriver.
- Overhaul of CassUserType (use macros to generate binding functions, binders for more values)

### DIFF
--- a/scylla-rust-wrapper/src/user_type.rs
+++ b/scylla-rust-wrapper/src/user_type.rs
@@ -22,7 +22,7 @@ pub unsafe extern "C" fn cass_user_type_new_from_data_type(
     let data_type = ptr_to_ref(data_type_raw);
 
     match data_type {
-        CassDataType::UDTDataType(udt_data_type) => {
+        CassDataType::UDT(udt_data_type) => {
             let field_values = udt_data_type
                 .field_types
                 .iter()
--- a/scylla-rust-wrapper/src/user_type.rs
+++ b/scylla-rust-wrapper/src/user_type.rs
@@ -1,36 +1,100 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::cass_types::CassDataType;
-use crate::cass_types::UDTDataType;
+use crate::cass_types::{CassDataTypeArc, UDTDataType};
+use crate::collection::CassCollection;
+use crate::inet::CassInet;
 use crate::types::*;
+use crate::uuid::CassUuid;
 use scylla::frame::response::result::CqlValue;
 use scylla::frame::response::result::CqlValue::*;
+use std::convert::TryInto;
 use std::os::raw::c_char;
+use std::sync::Arc;
 
 #[derive(Clone)]
 pub struct CassUserType {
-    pub udt_data_type: UDTDataType,
+    pub data_type: CassDataTypeArc,
 
     // Vec to preserve the order of fields
-    pub field_values: Vec<(String, Option<CqlValue>)>,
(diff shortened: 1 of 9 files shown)


scylla-rust-wrapper/src/user_type.rs
```diff
@@ -0,0 +1,315 @@
...
+        if !is_compatible_type(&*self.get_udt_type().field_types[index].1, &value) {
+            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+        }
+        self.field_values[index] = value;
+        CassError::CASS_OK
+    }
+
+    //TODO: some hashtable for name lookup?
+    unsafe fn set_option_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
+        let mut indices = Vec::new();
```
piodul, line 50: I understand that the borrow done by `get_udt_type` does not let you borrow `self` as mutable again in this `for` loop, but let's not introduce unnecessary allocations - the `indices` vector can be avoided with some reorganization. To avoid the problem you can start by converting `get_udt_type` to be a method of `CassDataType` and then inlining some of the logic of `set_option_by_index` inside the loop.

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -0,0 +1,315 @@
...
+pub struct CassUserType {
+    pub data_type: CassDataTypeArc,
+
+    // Vec to preserve the order of fields
+    pub field_values: Vec<Option<CqlValue>>,
+}
+
+fn is_compatible_type(_data_type: &CassDataType, _value: &Option<CqlValue>) -> bool {
+    // TODO: cppdriver actually checks types.
+    // TODO: this function should probably somewhere else
```
piodul, line 25: When is this planned to be implemented?

piodul (changes requested):
Please do not submit PRs which introduces large changes in one commit and then modifies everything several times in later commits. It really wastes reviewers' time when one expects changes to be logically split into commits and then reviews the changes commit by commit, not to mention that it makes both maintaining the repo and reviewing the git history much harder.

I know that this PR is an expanded version of another PR, but we aren't going to merge the old one so there is no need to preserve the old commits.

Judging by how nice the PR looks in the "Files changed" view, I think that most of the commits in this PR could be squashed together. Please clean it up a little.

scylla-rust-wrapper/src/user_type.rs
Lorak-mmk: In one of the future PR (probably sooner rather than later). This should be too hard to implement, as it's most likely going to just be direct translation of https://github.com/scylladb/cpp-driver/blob/master/src/data_type.hpp#L433-L564

Lorak-mmk:
Closing, refactored changes in https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/59 and https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/60

PR #58, opened 2021-12-04
Implement CassTuples, improved binding macros

Lorak-mmk:
1. Implemented CassTuple and all associated functions.
2. Created new macros that replace old version of `make_binders` - with the new version we can be sure that binding to each structure works the same (there is no code repetition), the code is shorter and less error-prone.

### DIFF
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -69,6 +69,25 @@ macro_rules! make_name_n_binder {
     }
 }
 
+macro_rules! make_appender {
+    ($this:ty, $consume_v:expr, $fn_append:ident, $e:expr, [$($arg:ident @ $t:ty), *]) => {
+        #[no_mangle]
+        #[allow(clippy::redundant_closure_call)]
+        pub unsafe extern "C" fn $fn_append(
+            this: *mut $this,
+            $($arg: $t), *
+        ) -> CassError {
+            // For some reason detected as unused, which is not true
+            #[allow(unused_imports)]
+            use scylla::frame::response::result::CqlValue::*;
+            match ($e)($($arg), *) {
+                Ok(v) => $consume_v(ptr_to_ref_mut(this), v),
+                Err(e) => e,
+            }
+        }
+    }
+}
+
 macro_rules! binders_maker_types {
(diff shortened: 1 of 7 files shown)


scylla-rust-wrapper/src/binding.rs
```diff
@@ -0,0 +1,261 @@
+use crate::cass_types::CassDataType;
```
piodul, line 1: This module really needs better documentation explaining how all of these macros work and how to use them. These macros are really hard to follow (it's not every day you see Rust code using macros to generate other macros), and it's unclear what they do without knowing the context in which they are supposed to be used.

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -248,8 +243,8 @@ make_binders!(@multi
     cass_user_type_set_string,
     cass_user_type_set_string_by_name,
     cass_user_type_set_string_by_name_n,
-    |v, n| Ok(Some(Text(ptr_to_cstr_n(v, n).unwrap().to_string()))),
-    [v @ *const c_char, n @ size_t],
+    |v| Ok(Some(Text(ptr_to_cstr(v).unwrap().to_string()))),
```
piodul: Could this be fixed in the "Cass types v2" PR instead?

scylla-rust-wrapper/src/binding.rs
```diff
@@ -216,6 +216,20 @@ macro_rules! binders_maker_types {
...
+    (tuple, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
+        $macro_name!(
+            $this,
+            $consume_v,
+            $fn,
+            |p: *const crate::tuple::CassTuple| {
+                match std::convert::TryInto::try_into(ptr_to_ref(p)) {
+                    Ok(v) => Ok(Some(v)),
+                    Err(e) => Err(e),
+                }
```
piodul, line 252: This is just `std::convert::TryInto::try_into(ptr_to_ref(p)).map(Some)` (not sure if `Some` can serve as a function; if not then change it to a closure: `|v| Some(v)`).

Lorak-mmk:
Closing, refactored changes in https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/59 and https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/60

PR #59, opened 2021-12-10
Refactor binding

Lorak-mmk:
This PR introduces macros to create binder functions that work with each use case from cpp-driver and allow to reduce code-duplication and amount of code.
Apart from that, there are few small fixes:
 - updating Cargo.lock to use current version of our Rust-driver fork
 - replacing unnecessary string decodings with strlen calls
 - Changed TryFrom in CassCollection to avoid unnecessary copies

### DIFF
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -1,0 +1,379 @@
+//! This module implements macros that generate functions used to
+//! bind/append data to driver's structures (statement, UDT, tuple, collection)
+//!
+//! For examples on how to use this module, see documentation of `prepare_binders_macro`
+//!
+//! This module is a bit complicated, mostly because we can't concatenate identifiers
+//! in macros, but wanted to avoid code duplication anyway (so for instance having code
+//! that converts function arguments to CqlValue in only one place).
+//! `make_*_binder` and `make_appender` are macros that directly declare new functions.
+//! `invoke_binder_maker_macro_with_type` is responsible for converting function arguments to `Result<Option<CqlValue>, CassError>`.
+//! It basically takes a macro name (`make_*_binder` / `make_appender`), some arguments,
+//! and forwards those arguments to the macro, along with the description of additional arguments
+//! and a function that converts them to `Result<Option<CqlValue>, CassError>`.
+//! `prepare_binders_macro` declares `make_binders` macro, that is then directly called
+//! by module user (more on `make_binders` in documentation of `prepare_binders_macro`).
+//! Generally, `make_binders` will just use `invoke_binder_maker_macro_with_type` underneath,
+//! which in turn uses `make_*_binder` / `make_appender`, which declare desired function.
+
+//! # make_*_binder
+//! make_*_binder and make_appender generate target function.
+//! They should not be used directly, and are present only for internal usage by other macros.
+//!
+//! ## Arguments for make_*_binder and make_appender
+//!
+//! * `$this:ty` - type to which we are going to bind (e.g. `CassTuple`)
+//! * `$consume_v:expr` - function that accepts structure reference (`&mut $this`),
+//!     then optionally some flavour-specific arguments (described futher down),
+//!     value (`Option<CqlValue>`), and binds value to the structure.
+//! * `$fn_by_*` - name of the generated function.
+//! * `$e:expr` - takes remaining function arguments and produces `Result<Option<CqlValue>, CassError>`
+//! * `[$($arg:ident @ $t:ty), *]` - list of remaining function arguments and their types.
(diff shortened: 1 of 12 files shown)


scylla-rust-wrapper/src/binding.rs
```diff
@@ -0,0 +1,379 @@
...
+//! in macros, but wanted to avoid code duplication anyway (so for instance having code
+//! that converts function arguments to CqlValue in only one place).
+//! `make_*_binder` and `make_appender` are macros that directly declare new functions.
+//! `binders_maker_types` is responsible for converting function arguments to `Result<Option<CqlValue>, CassError>`.
+//! It basically takes a macro name (`make_*_binder` / `make_appender`), some arguments,
+//! and forwards those arguments to the macro, along with the description of additional arguments
+//! and a function that converts them to `Result<Option<CqlValue>, CassError>`.
+//! `prepare_binders_macro` declares `make_binders` macro, that is then directly called
+//! by module user (more on `make_binders` in documentation of `prepare_binders_macro`).
+//! Genrally, `make_binders` will just use `binders_maker_types` underneath,
```
piodul: Nit: typo: `Genrally` -> `Generally`

scylla-rust-wrapper/src/binding.rs
```diff
@@ -0,0 +1,379 @@
...
+            use scylla::frame::response::result::CqlValue::*;
+            match ($e)($($arg), *) {
+                Ok(v) => $consume_v(ptr_to_ref_mut(this), v),
+                Err(e) => e,
+            }
+        }
+    }
+}
+
+macro_rules! binders_maker_types {
```
piodul: I don't understand this macro's name (what is "binder maker types" supposed to mean?). How about e.g. `invoke_binder_maker_macro_with_type`?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -30,6 +25,31 @@ pub struct CassStatement {
     pub paging_state: Option<Bytes>,
 }
 
+impl CassStatement {
```
piodul, line 28: The last commit mixes up multiple things which include 1) rewrite of the setters to use macros 2) moving functions lower on higher within the same file 3) addition of the CassStatement methods. This makes comparing old and new code hard. Please split the changes apart (not necessarily in that order).

scylla-rust-wrapper/src/binding.rs
Lorak-mmk: Fixed

scylla-rust-wrapper/src/binding.rs
Lorak-mmk: You're right, this name was bad, fixed.

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: Done

scylla-rust-wrapper/src/statement.rs
```diff
@@ -30,6 +30,31 @@ pub struct CassStatement {
...
+            CassError::CASS_OK
+        }
+    }
+
+    fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
+        match &self.statement {
+            Statement::Prepared(prepared) => {
+                for (i, col) in prepared.get_metadata().col_specs.iter().enumerate() {
+                    if col.name == name {
+                        return self.bind_cql_value(i, value);
```
piodul, line 43: The original cpp driver binds to all values with given name, not only to the first one. Although I can't really imagine a useful case for this behavior (I might have poor imagination), it IS possible to have multiple values with the same name in the same prepared statement, e.g. in this one:

```cql
SELECT pk, ck, v FROM ks.t WHERE pk = ? AND ck > ? AND ck < ?
```

(there are three bind markers here, and the last two have name `ck`, I checked by printing the rust driver's metadata)

For the sake of compatibility, let's keep the old behavior.

piodul (commented):
Left one comment, otherwise LGTM

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: I noticed that, and I agree that we should keep compatibility in this regard. Would it be ok to fix this in one of the next PR's? After merging this one and https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/60 I wanted to work on https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/31 - and in my local version I already have this problem fixed.

piodul: OK, I'm fine with fixing it in later PRs. If the fix is unrelated/orthogonal to the work in #31 then I'd recommend sending it in a separate PR.

piodul (approved):

PR #60, opened 2021-12-10
Implement CassDataType, CassUserType, CassTuple

Lorak-mmk:
Based on https://github.com/hackathon-rust-cpp/cpp-rust-driver/pull/58
Implements CassDataType
Implements CassUserType, CassTuple and all bindings related to them, using new macros.

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -1,0 +1,468 @@
+use crate::argconv::*;
+use crate::cass_error::CassError;
+use crate::types::*;
+use std::os::raw::c_char;
+use std::ptr;
+use std::sync::Arc;
+
+include!(concat!(env!("OUT_DIR"), "/cppdriver_data_types.rs"));
+
+#[derive(Clone)]
+pub struct UDTDataType {
+    // Vec to preserve the order of types
+    pub field_types: Vec<(String, CassDataTypeArc)>,
+
+    pub keyspace: String,
+    pub name: String,
+}
+
+impl UDTDataType {
+    pub fn new() -> UDTDataType {
+        UDTDataType {
+            field_types: Vec::new(),
+            keyspace: "".to_string(),
+            name: "".to_string(),
+        }
+    }
+
+    pub fn with_capacity(capacity: usize) -> UDTDataType {
+        UDTDataType {
+            field_types: Vec::with_capacity(capacity),
+            keyspace: "".to_string(),
+            name: "".to_string(),
+        }
+    }
+
+    pub fn add_field(&mut self, name: String, field_type: CassDataTypeArc) {
+        self.field_types.push((name, field_type));
+    }
+
+    pub fn get_field_by_name(&self, name: &str) -> Option<&CassDataTypeArc> {
+        self.field_types
+            .iter()
+            .find(|(field_name, _)| field_name == name)
+            .map(|(_, t)| t)
+    }
+
+    pub fn get_field_by_index(&self, index: usize) -> Option<&CassDataTypeArc> {
+        self.field_types.get(index).map(|(_, b)| b)
+    }
+}
+
+impl Default for UDTDataType {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Clone)]
+pub enum CassDataType {
+    Value(CassValueType),
+    UDT(UDTDataType),
+    List(Option<CassDataTypeArc>),
+    Set(Option<CassDataTypeArc>),
+    Map(Option<CassDataTypeArc>, Option<CassDataTypeArc>),
+    Tuple(Vec<CassDataTypeArc>),
+    Custom(String),
+}
+
+pub type CassDataTypeArc = Arc<CassDataType>;
+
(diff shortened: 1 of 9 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -0,0 +1,470 @@
...
+            name: "".to_string(),
+        }
+    }
+
+    pub fn add_field(&mut self, name: String, field_type: CassDataTypeArc) {
+        self.field_types.push((name, field_type));
+    }
+
+    pub fn get_field_by_name(&self, name: &str) -> Option<&CassDataTypeArc> {
+        for (field_name, field_type) in self.field_types.iter() {
```
piodul: Nit: not sure if you like terse, iterator-base style, but if you do this should work, too:
```rust
self.field_types
    .iter()
    .find(|(field_name, _)| field_name == name)
    .map(|(_, t)| t)
```

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(Clone)]
+pub struct CassUserType {
+    pub data_type: CassDataTypeArc,
+
+    // Vec to preserve the order of fields
+    pub field_values: Vec<Option<CqlValue>>,
+}
+
+impl CassUserType {
+    unsafe fn set_option_by_index(&mut self, index: usize, value: Option<CqlValue>) -> CassError {
```
piodul: Why is this function unsafe?

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+        if !is_compatible_type(&self.data_type.get_udt_type().field_types[index].1, &value) {
+            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+        }
+        self.field_values[index] = value;
+        CassError::CASS_OK
+    }
+
+    //TODO: some hashtable for name lookup?
+    unsafe fn set_option_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
```
piodul: Ditto?

scylla-rust-wrapper/src/tuple.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::cass_types::CassDataTypeArc;
+use crate::types::*;
+use scylla::frame::response::result::CqlValue;
+use scylla::frame::value::MaybeUnset;
+use scylla::frame::value::MaybeUnset::{Set, Unset};
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct CassTuple {
+    pub data_type: CassDataTypeArc,
```
piodul: It was a bit surprising for me to find out that the tuple in cpp driver actually only validates the first N types specified in its data type (and if it is a completely new tuple, it doesn't validate anything). Until I realized this, the logic looked incorrect to me and I was thinking how it should be fixed.

Please describe this behavior in this file so that other people reading/modifying this code in the future will understand what is going on quicker.

piodul (changes requested):

scylla-rust-wrapper/src/tuple.rs
Lorak-mmk: I don't fully understand your comment, but it made me look at cppdriver's tuple code again, and it seems my implementation here is surprisingly incorrect in few places. I'll fix it and leave some comments.

Lorak-mmk:
@piodul I fixed all issues pointed by you. Could you check if current tuple implementation is consistent with cpp-driver? I don't trust myself - I was sure it was correct before, but today it turned out to be wrong.

haaawk:
@piodul could you help @Lorak-mmk with tuples question?

scylla-rust-wrapper/src/tuple.rs
piodul: I guess the most confusing part for me was the `inner_types.len() > index` check. It is always true if the tuple was created with `cass_tuple_new_from_data_type` and always false when with tuples created by `cass_tuple_new`. It does not depend on the value of `index` at all.

To make things worse, I didn't know that either `inner_types.len() == 0` or `inner_types.len() == self.items.len()`, so I was wondering what would be the meaning of the original check if `0 < inner_types.len() < self.items.len()`.

In addition to the comment with description (which I can see you have already added) I would suggest expressing intent more clearly in the code. Maybe we could change `data_type` into an `Option<CassDataTypeArc>`? It would be `None` if created with `cass_tuple_new`, and `Some` in the other case.

piodul:
@Lorak-mmk I'm not too familiar with the cpp-driver's logic, but from what I managed to understand, your current logic checks out. In the future, it would be worthwhile to write differential tests in order to help enforce the same behavior as in the cpp-driver.

piodul (commented):
Please address the issue of confusing `data_type` for untyped tuples and I think it will be OK to merge.

Lorak-mmk:
Fixed. I had to use global variable to handle `cass_tuple_data_type` for tuples created by `cass_tuple_new` @piodul

piodul (approved):

PR #63, opened 2022-07-07
Basic tests setup

Gor027:
This PR adds an implementation of necessary things for the BasicTests setup. The BasicTests do not pass yet, as row and column iterators are yet to be implemented. Additionally, minimal validations are added in cluster configuration in order to pass ClusterTests.

PR #64, opened 2022-07-08
Add RowIterator implementation

Gor027:
This PR is dependent upon #63

As the C++ driver interface requires functions `cass_iterator_from_result` and `cass_iterator_from_row` to return `CassIterator`, this PR modifies already existing `CassIterator` struct to be enum and to contain `CassResultIterator` and `CassRowIterator`.
Additionally, some necessary functions were implemented in order to have at least 1 passing integration test. So, it should pass a basic test called `BasicsTests.Integration_Cassandra_RowsInRowsOut`.

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -1,6 +1,7 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::types::*;
+use scylla::frame::response::result::ColumnType;
 use std::os::raw::c_char;
 use std::ptr;
 use std::sync::Arc;
@@ -125,6 +126,59 @@ impl CassDataType {
             CassDataType::UDT(udt) => udt,
             _ => panic!("Can get UDT out of non-UDT data type"),
         }
+    }
+}
+
+pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
+    match column_type {
+        ColumnType::Custom(s) => CassDataType::Custom((*s).clone()),
+        ColumnType::Ascii => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_ASCII),
+        ColumnType::Boolean => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BOOLEAN),
+        ColumnType::Blob => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BLOB),
+        ColumnType::Counter => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_COUNTER),
+        ColumnType::Decimal => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_DECIMAL),
+        ColumnType::Date => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_DATE),
+        ColumnType::Double => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_DOUBLE),
+        ColumnType::Float => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_FLOAT),
+        ColumnType::Int => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_INT),
+        ColumnType::BigInt => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BIGINT),
+        ColumnType::Text => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TEXT),
+        ColumnType::Timestamp => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TIMESTAMP),
+        ColumnType::Inet => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_INET),
+        ColumnType::List(boxed_type) => {
+            CassDataType::List(Some(Arc::new(get_column_type(boxed_type.as_ref()))))
+        }
+        ColumnType::Map(key, value) => CassDataType::Map(
+            Some(Arc::new(get_column_type(key.as_ref()))),
+            Some(Arc::new(get_column_type(value.as_ref()))),
+        ),
+        ColumnType::Set(boxed_type) => {
+            CassDataType::Set(Some(Arc::new(get_column_type(boxed_type.as_ref()))))
+        }
+        ColumnType::UserDefinedType {
+            type_name,
+            keyspace,
+            field_types,
+        } => CassDataType::UDT(UDTDataType {
+            field_types: field_types
+                .iter()
+                .map(|(name, col_type)| ((*name).clone(), Arc::new(get_column_type(col_type))))
+                .collect(),
+            keyspace: (*keyspace).clone(),
+            name: (*type_name).clone(),
+        }),
+        ColumnType::SmallInt => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_SMALL_INT),
+        ColumnType::TinyInt => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TINY_INT),
+        ColumnType::Time => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TIME),
+        ColumnType::Timeuuid => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TIMEUUID),
+        ColumnType::Tuple(v) => CassDataType::Tuple(
+            v.iter()
+                .map(|col_type| Arc::new(get_column_type(col_type)))
+                .collect(),
+        ),
+        ColumnType::Uuid => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_UUID),
+        ColumnType::Varint => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_VARINT),
+        _ => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_UNKNOWN),
     }
 }
 
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -3,13 +3,22 @@ use crate::cass_error::CassError;
 use crate::inet::CassInet;
 use crate::types::*;
 use crate::uuid::CassUuid;
-use scylla::frame::response::result::{CqlValue, Row};
-use scylla::QueryResult;
+use scylla::frame::response::result::{ColumnSpec, CqlValue};
+use scylla::Bytes;
 use std::convert::TryInto;
 use std::os::raw::c_char;
 use std::sync::Arc;
 
-pub type CassResult = QueryResult;
+pub struct CassResult {
+    pub rows: Option<Vec<CassRow>>,
+    pub metadata: Arc<CassResultData>,
+}
+
+pub struct CassResultData {
+    pub paging_state: Option<Bytes>,
+    pub col_specs: Vec<ColumnSpec>,
+}
+
 pub type CassResult_ = Arc<CassResult>;
 
 pub struct CassIterator {
@@ -17,7 +26,10 @@ pub struct CassIterator {
     position: Option<usize>,
 }
 
-pub type CassRow = Row;
+pub struct CassRow {
+    pub columns: Vec<CassValue>,
+    pub result_metadata: Arc<CassResultData>,
+}
 
 pub type CassValue = Option<CqlValue>;
 
@@ -44,7 +56,7 @@ pub unsafe extern "C" fn cass_result_free(result_raw: *mut CassResult) {
 #[no_mangle]
 pub unsafe extern "C" fn cass_result_has_more_pages(result: *const CassResult) -> cass_bool_t {
     let result = ptr_to_ref(result);
-    result.paging_state.is_some() as cass_bool_t
+    result.metadata.paging_state.is_some() as cass_bool_t
 }
 
 #[no_mangle]
@@ -79,7 +91,7 @@ pub unsafe extern "C" fn cass_iterator_get_row(iterator: *const CassIterator) ->
         None => return std::ptr::null(),
     };
 
-    let row: &Row = match iter
+    let row: &CassRow = match iter
         .result
         .rows
         .as_ref()
@@ -311,7 +323,7 @@ pub unsafe extern "C" fn cass_result_row_count(result_raw: *const CassResult) ->
 pub unsafe extern "C" fn cass_result_column_count(result_raw: *const CassResult) -> size_t {
     let result = ptr_to_ref(result_raw);
 
-    result.col_specs.len() as size_t
+    result.metadata.col_specs.len() as size_t
 }
 
 #[no_mangle]
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1,5 +1,6 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
+use crate::cass_types::{cass_data_type_type, CassDataType, CassValueType};
 use crate::inet::CassInet;
 use crate::types::*;
 use crate::uuid::CassUuid;
@@ -21,42 +22,31 @@ pub struct CassResultData {
 
 pub type CassResult_ = Arc<CassResult>;
 
-pub struct CassIterator {
-    result: CassResult_,
-    position: Option<usize>,
-}
+pub type CassRow_ = &'static CassRow;
 
 pub struct CassRow {
     pub columns: Vec<CassValue>,
     pub result_metadata: Arc<CassResultData>,
 }
 
-pub type CassValue = Option<CqlValue>;
-
-#[no_mangle]
-pub unsafe extern "C" fn cass_iterator_from_result(
-    result_raw: *const CassResult,
-) -> *mut CassIterator {
-    let result: CassResult_ = clone_arced(result_raw);
-
-    let iterator = CassIterator {
-        result,
-        position: None,
-    };
-
-    Box::into_raw(Box::new(iterator))
+pub struct CassValue {
+    pub value: Option<CqlValue>,
+    pub value_type: CassDataType,
 }
 
-// This was const for some reason, seems like a mistake in cpp driver
-#[no_mangle]
-pub unsafe extern "C" fn cass_result_free(result_raw: *mut CassResult) {
-    free_arced(result_raw);
+pub struct CassResultIterator {
+    result: CassResult_,
+    position: Option<usize>,
 }
 
-#[no_mangle]
-pub unsafe extern "C" fn cass_result_has_more_pages(result: *const CassResult) -> cass_bool_t {
-    let result = ptr_to_ref(result);
-    result.metadata.paging_state.is_some() as cass_bool_t
+pub struct CassRowIterator {
+    row: CassRow_,
+    position: Option<usize>,
+}
+
+pub enum CassIterator {
+    CassResultIterator(CassResultIterator),
+    CassRowIterator(CassRowIterator),
 }
 
 #[no_mangle]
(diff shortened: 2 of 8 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -55,41 +44,117 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
 pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass_bool_t {
-    let iter: &mut CassIterator = ptr_to_ref_mut(iterator);
+    let mut iter = ptr_to_ref_mut(iterator);
+
+    match &mut iter {
+        CassIterator::CassResultIterator(result_iterator) => {
+            let new_pos: usize = match result_iterator.position {
+                Some(prev_pos) => prev_pos + 1,
+                None => 0,
+            };
```
piodul: Tip: you can use `map_or` to make is shorter:

```rust
let new_pos = result_iterator.map_or(0, |x| x + 1);
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -12,39 +13,27 @@ use std::sync::Arc;
...
-) -> *mut CassIterator {
-    let result: CassResult_ = clone_arced(result_raw);
-
-    let iterator = CassIterator {
-        result,
-        position: None,
-    };
-
-    Box::into_raw(Box::new(iterator))
+#[repr(C)]
```
piodul: `repr(C)` is not needed here. It's only important if you have a corresponding definition in a C header and you want both Rust and C to use the same struct. If you return an opaque pointer to C then `repr` doesn't matter and it's better to leave it as default.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -12,39 +13,27 @@ use std::sync::Arc;
...
+pub struct CassResultIterator {
+    result: CassResult_,
+    position: Option<usize>,
 }
 
-// This was const for some reason, seems like a mistake in cpp driver
-#[no_mangle]
-pub unsafe extern "C" fn cass_result_free(result_raw: *mut CassResult) {
-    free_arced(result_raw);
+#[repr(C)]
```
piodul: No need for `repr(C)`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -12,39 +13,27 @@ use std::sync::Arc;
...
+pub struct CassRowIterator {
+    row: CassRow_,
+    position: Option<usize>,
 }
 
-#[no_mangle]
-pub unsafe extern "C" fn cass_result_has_more_pages(result: *const CassResult) -> cass_bool_t {
-    let result = ptr_to_ref(result);
-    result.paging_state.is_some() as cass_bool_t
+#[repr(C)]
```
piodul: No need for `repr(C)`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -55,41 +44,117 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+
+    // Defined only for row iterator, for other types should return null
+    if let CassIterator::CassRowIterator(row_iterator) = iter {
+        let iter_position = match row_iterator.position {
+            Some(pos) => pos,
+            None => return std::ptr::null(),
+        };
 
-    iter.position = Some(new_pos);
+        let column: &CassValue = row_iterator.row.columns.get(iter_position).unwrap();
```
piodul: Is this panic here intentional? If the iterator goes past the last column, there will be a panic here. How does the original cpp driver code behave?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -108,6 +173,78 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+        Some(CqlValue::TinyInt(_)) => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TINY_INT),
+        Some(CqlValue::Time(_)) => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TIME),
+        Some(CqlValue::Timeuuid(_)) => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_TIMEUUID),
+        Some(CqlValue::Tuple(_)) => CassDataType::Tuple(vec![]),
+        Some(CqlValue::Uuid(_)) => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_UUID),
+        Some(CqlValue::Varint(_)) => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_VARINT),
+        _ => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_UNKNOWN),
+    };
+
+    Arc::into_raw(Arc::new(cass_data_type))
```
piodul: The docstring next to `cass_value_data_type` contains the following text:

```
 * @return Returns a reference to the data type of the value.
 * Do not free this reference as it is bound to the lifetime of the value.
```

We cannot allocate and return new objects because users won't be freeing them and this will lead to memory leaks.

You need to store a `CassDataType` along with `CassValue` somehow.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -55,41 +44,117 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+            match &result_iterator.result.rows {
+                Some(rs) => (new_pos < rs.len()) as cass_bool_t,
+                None => false as cass_bool_t,
+            }
+        }
+        CassIterator::CassRowIterator(row_iterator) => {
+            let new_pos: usize = match row_iterator.position {
+                Some(prev_pos) => prev_pos + 1,
+                None => 0,
+            };
```
piodul: Use `map_or` here, too.

piodul (changes requested):

scylla-rust-wrapper/src/query_result.rs
Gor027: If the iterator goes past the last column then `cass_iterator_next` should return `false` and `cass_iterator_get_column` will not get executed.

scylla-rust-wrapper/src/query_result.rs
Gor027: The only way I could think of is to change the type `CassValue` to either a tuple or a struct to contain `CassDataType`, but that would cause a lot more other changes. Perhaps, you have other suggestions?

scylla-rust-wrapper/src/query_result.rs
piodul: Nothing prevents users from calling `cass_iterator_get_column` after `cass_iterator_next` returns false. In such case we should behave as the original driver does.

Besides, you already implemented a similar function `cass_iterator_get_row` and there you return `std::ptr::null()` when it is called on an exhausted iterator. We have two similar functions which behave inconsistently. Please check how the original driver behaves and align those functions with this behavior (I'd expect that they return null).

scylla-rust-wrapper/src/query_result.rs
piodul: I'm afraid changing `CassValue` to a struct that contains `CassDataType` is the simplest solution that I can think of.

piodul: This doesn't really fix it. OK, you added `value_type` field to `CassValue`, but you are still using `Arc::new()` which allocates a new object. You will still have memory leaks because users won't free this object.

Once again - the docstring explains that the function is supposed to return a `CassDataType` which lives as long as the `CassValue` it is associated with. Therefore, you just need to return a pointer to the `value_type`.

piodul:
Please avoid addressing review comments by adding commits that fix bugs introduced in previous commits. In particular, the first commit still has the issue with the lifetime of the value returned from `cass_value_data_type`. You should first change the `CassValue` to store `CassDataType` so that you can correctly implement `cass_value_data_type` in the next commit.

Moreover, please add commit descriptions. Each description should explain what is changed in the commit and why.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -276,13 +416,13 @@ pub unsafe extern "C" fn cass_value_get_string(
     output_size: *mut size_t,
 ) -> CassError {
     let val: &CassValue = ptr_to_ref(value);
-    match val {
+    match val.value.clone() {
```
piodul: Why `.clone()`? You aren't moving out anything of the value you are matching on, so probably `match &val.value` will suffice.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -100,12 +164,87 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_type(value: *const CassValue) -> CassValueType {
+    let value_from_raw = ptr_to_ref(value);
+
+    cass_data_type_type(&value_from_raw.value_type)
+}
+
+pub fn get_type_from_value(value: Option<CqlValue>) -> CassDataType {
```
piodul: CqlValue might potentially be a big object. There is no reason for this function to receive it by value, it only forces the caller to `.clone()` it. Please change the type of `value` to `&Option<CqlValue>`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -55,41 +44,116 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+        };
+
+        let value = match row_iterator.row.columns.get(iter_position) {
+            Some(col) => col.clone(),
+            None => return std::ptr::null(),
+        };
+        let value_type = get_type_from_value(value.clone());
+        let column = CassValue { value, value_type };
+
+        return Box::into_raw(Box::new(column)) as *const CassValue;
```
piodul: This function does not allocate a new value in the original implementation. It just returns a reference to an existing `CassValue`.

You need to make bigger changes in order to make it work. More specifically, You need to change `CassRow` so that it keeps a `Vec<CassValue>`. In turn, you probably need to modify the definition of `CassResult`, too. I suggest doing those changes first, in a separate commit, and then continuing with the rest.

piodul (changes requested):

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -79,7 +91,7 @@ pub unsafe extern "C" fn cass_iterator_get_row(iterator: *const CassIterator) ->
         None => return std::ptr::null(),
     };
 
-    let row: &Row = match iter
+    let row = match iter
```
piodul: Nit: please avoid doing irrelevant changes like this. They introduce unnecessary noise in the diff.

If you want to do some cleanups (and have a good reason to do so, e.g. it is related to the PR) then I recommend putting them in a separate commit. In this case it's better to drop it altogether IMO.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -3,21 +3,33 @@ use crate::cass_error::CassError;
...
 
 pub struct CassIterator {
     result: CassResult_,
     position: Option<usize>,
 }
 
-pub type CassRow = Row;
+pub struct CassRow {
+    pub columns: Vec<CassValue>,
+    pub result_: Arc<CassResultData>,
```
piodul: Why is there an underscore at the end of `result_`?

Besides, wouldn't a name like `metadata` be better, like in `CassResult::metadata`? I don't understand why it is inconsistent now.

scylla-rust-wrapper/src/session.rs
```diff
@@ -86,12 +88,44 @@ pub unsafe extern "C" fn cass_session_execute(
...
+                    columns: r.columns.clone(),
+                    result_: metadata.clone(),
+                })
+                .collect();
+
+            Some(cass_rows)
+        }
+        None => None,
+    }
+}
```
piodul: Tip: no need to write such a big `match` block. The `populate_cass_rows_from_rows` function returns an `Option`, so you can just use the `?` operator:

```rust
let rows = rows.as_ref()?;
let cass_rows = /* ... */
```

scylla-rust-wrapper/src/session.rs
```diff
@@ -86,12 +88,44 @@ pub unsafe extern "C" fn cass_session_execute(
...
+                });
+
+                Ok(CassResultValue::QueryResult(cass_result))
+            }
             Err(err) => Ok(CassResultValue::QueryError(Arc::new(err))),
         }
     })
 }
 
+fn populate_cass_rows_from_rows(
```
piodul: Nitpick: usually, `populate` means filling an already existing structure with data. This function creates a completely new `Option<Vec<CassRow>>`, so how about using `create` instead of `populate`?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -67,41 +57,114 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+    let iterator = CassRowIterator {
+        row: row_from_raw,
+        position: None,
     };
 
-    row
+    Box::into_raw(Box::new(CassIterator::CassRowIterator(iterator)))
+}
+
+// This was const for some reason, seems like a mistake in cpp driver
```
piodul: It looks like `free_arced` takes a const pointer as an argument. Shouldn't this function work with `*const CassResult`?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -112,12 +175,48 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+    name_length: *mut size_t,
+) -> CassError {
+    let result_from_raw = ptr_to_ref(result);
+    let index_usize: usize = index.try_into().unwrap();
+
+    if index_usize >= result_from_raw.metadata.col_specs.len() {
+        return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
+    }
+
+    let column_spec: &ColumnSpec = result_from_raw.metadata.col_specs.get(index_usize).unwrap();
```
piodul, line 199: How does the original function behave when `index_usize` is larger than the last column index? I doubt that we should crash here.

scylla-rust-wrapper/src/session.rs
```diff
@@ -86,12 +88,44 @@ pub unsafe extern "C" fn cass_session_execute(
...
+    rows: &Option<Vec<Row>>,
```
piodul: Takes `rows` by value, not by reference. This way you can later use `rs.into_iter()` instead of `rs.iter()` and avoid copying.

scylla-rust-wrapper/src/session.rs
```diff
@@ -126,6 +127,16 @@ fn populate_cass_rows_from_rows(
     }
 }
 
+fn populate_cass_row_columns(row: &Row) -> Vec<CassValue> {
```
piodul: If you address my comment about `populate_cass_rows_from_rows`, then you will be able to change this function so that it takes `row` by value.

Also, please change `populate` to `create` in the function's name.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -128,6 +129,40 @@ impl CassDataType {
     }
 }
 
+pub fn get_type_from_value(value: &Option<CqlValue>) -> CassDataType {
```
piodul: You cannot get a complete type just from the value alone. It doesn't contain full information about complex data types, i.e.: lists, sets, maps and UDTs. You must get this information from the metadata in `CassResultData`.

piodul (changes requested):

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -112,12 +175,48 @@ pub unsafe extern "C" fn cass_row_get_column(
...
+    name: *mut *const c_char,
+    name_length: *mut size_t,
+) -> CassError {
+    let result_from_raw = ptr_to_ref(result);
+    let index_usize: usize = index.try_into().unwrap();
+
+    if index_usize >= result_from_raw.metadata.col_specs.len() {
+        return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
+    }
+
```
Gor027, line 199: We check it here, and in case it's larger than the last column index, just returns a `CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS`. The implementation is the same as in the C++ driver.

scylla-rust-wrapper/src/query_result.rs
Gor027: We check it here, and in case it's larger than the last column index, just returns a `CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS`. The implementation is the same as in the C++ driver.

scylla-rust-wrapper/src/query_result.rs
Gor027: Here, the type of `row` was changed to `CassRow`, so I just removed the explicit type annotation. Maybe it is better to annotate it as `CassRow`.

scylla-rust-wrapper/src/cass_types.rs
Gor027: `CassResultData.col_specs` has enough information to retrieve a complete type, however, it's not accessible. The field `typ: ColumnType` of `ColumnSpec` is not public in the Rust driver. I think the only way of getting column-specific metadata is the `QueryResult.col_specs`, so we are obliged to make some changes in the Rust driver side: making the field `typ: ColumnType` to be public. Perhaps we should do that in another PR?

Gor027: EDIT: sorry, I looked at the forked driver version, in the original repo the mentioned `ColumnSepc.typ` is public. So, after merging #67, the function `get_type_from_value` can be fixed to return complete type.

Gor027: @piodul This is fixed now, could you please review it once again?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -128,6 +129,61 @@ impl CassDataType {
     }
 }
 
+pub fn get_column_type(column_type: Option<ColumnType>) -> CassDataType {
```
piodul: `ColumnType` can be potentially heavy when describing compound types such as collections or UDTs. This function doesn't have to take `column_type` by value, you can take it by reference.

scylla-rust-wrapper/src/session.rs
piodul: Changing the type of `rows` should be made in the first commit.

scylla-rust-wrapper/src/session.rs
```diff
@@ -107,21 +108,32 @@ pub unsafe extern "C" fn cass_session_execute(
...
         .collect();
 
     Some(cass_rows)
 }
 
+fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<CassValue> {
+    row.columns
+        .into_iter()
+        .enumerate()
+        .map(|(index, col)| CassValue {
```
piodul: Tip: the length of `col_specs` and `row.columns` should always be the same, so you should be able to just use `zip`:

```rust
row.columns
    .iter()
    .zip(metadata.col_specs.iter())
    .map(|(col, spec)| {
        // ...
    })
    .collect();
```

piodul (changes requested):

scylla-rust-wrapper/src/session.rs
```diff
@@ -86,12 +88,40 @@ pub unsafe extern "C" fn cass_session_execute(
...
             Err(err) => Ok(CassResultValue::QueryError(Arc::new(err))),
         }
     })
 }
 
+fn create_cass_rows_from_rows(
+    rows: Option<Vec<Row>>,
+    metadata: &Arc<CassResultData>,
+) -> Option<Vec<CassRow>> {
+    let rows = rows.as_ref()?;
```
piodul: Drop `as_ref()` here and you won't have to clone in line `117`. Rows contain user data so they can be potentially large and we would like to avoid copies.

scylla-rust-wrapper/src/query_result.rs
piodul: Right, I missed it, thanks.

piodul (approved):

constantfold (approved):

PR #65, opened 2022-07-20
Basic tests

Gor027:
This should be merged only after #66 is merged.
This PR is dependent upon #64

This PR adds implementation for some necessary functions to pass all `BasicTests`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,4 +33,4 @@ jobs:
       run: cmake -DCASS_BUILD_INTEGRATION_TESTS=ON . && make
 
     - name: Run integration tests on Scylla 5.0.0
-      run: valgrind --error-exitcode=123 ./cassandra-integration-tests --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="ClusterTests.*:BasicsTests.*RowsInRowsOut"
+      run: valgrind --error-exitcode=123 ./cassandra-integration-tests --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="ClusterTests.*:BasicsTests.*"
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -2,12 +2,14 @@ use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::cass_types::{cass_data_type_type, CassDataType, CassValueType};
 use crate::inet::CassInet;
+use crate::statement::CassStatement;
 use crate::types::*;
 use crate::uuid::CassUuid;
 use scylla::frame::response::result::{ColumnSpec, CqlValue};
-use scylla::Bytes;
+use scylla::{BufMut, Bytes, BytesMut};
 use std::convert::TryInto;
 use std::os::raw::c_char;
+use std::slice;
 use std::sync::Arc;
 
 pub struct CassResult {
@@ -402,6 +404,32 @@ pub unsafe extern "C" fn cass_value_get_string(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn cass_value_get_bytes(
+    value: *const CassValue,
+    output: *mut *const cass_byte_t,
+    output_size: *mut size_t,
+) -> CassError {
+    if value.is_null() {
+        return CassError::CASS_ERROR_LIB_NULL_VALUE;
+    }
+
+    let value_from_raw: &CassValue = ptr_to_ref(value);
+
+    // FIXME: This should be implemented for all CQL types
+    // Note: currently rust driver does not allow to get raw bytes of the CQL value.
+    match &value_from_raw.value {
+        Some(CqlValue::Blob(bytes)) => {
+            *output = bytes.as_ptr() as *const cass_byte_t;
+            *output_size = bytes.len() as u64;
+        }
+        Some(_) => return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE,
+        None => return CassError::CASS_ERROR_LIB_NULL_VALUE,
+    }
+
+    CassError::CASS_OK
+}
+
+#[no_mangle]
 pub unsafe extern "C" fn cass_value_is_null(value: *const CassValue) -> cass_bool_t {
     let val: &CassValue = ptr_to_ref(value);
     val.value.is_none() as cass_bool_t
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -33,4 +33,4 @@ jobs:
       run: cmake -DCASS_BUILD_INTEGRATION_TESTS=ON . && make
 
     - name: Run integration tests on Scylla 5.0.0
-      run: valgrind --error-exitcode=123 ./cassandra-integration-tests --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="ClusterTests.*:BasicsTests.*RowsInRowsOut"
+      run: valgrind --error-exitcode=123 ./cassandra-integration-tests --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="ClusterTests.*:BasicsTests.*"
```
piodul, line 36: This change seems to remove tests from the filter. Is that intended?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -436,6 +462,63 @@ pub unsafe extern "C" fn cass_result_first_row(result_raw: *const CassResult) ->
...
+) -> CassError {
+    let statement_from_raw = ptr_to_ref_mut(statement);
+
+    if paging_state.is_null() {
+        statement_from_raw.paging_state = None;
+        return CassError::CASS_ERROR_LIB_NULL_VALUE;
+    }
+
+    let paging_state_usize: usize = paging_state_size.try_into().unwrap();
+    let mut b = BytesMut::from(slice::from_raw_parts(
```
piodul: [`Bytes::copy_from_slice`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.copy_from_slice) will be more explicit here than `Bytes::from`. Its name more clearly conveys what exactly it is doing.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -436,6 +462,63 @@ pub unsafe extern "C" fn cass_result_first_row(result_raw: *const CassResult) ->
...
+            .paging_state
+            .as_ref()
+            .unwrap()
+            .len() as u64;
+        *paging_state = result_from_raw
+            .metadata
+            .paging_state
+            .clone()
+            .unwrap()
+            .as_ptr() as *const c_char;
```
piodul: You are (most probably) returning a pointer to freed memory here.

.clone() creates a temporary `Option<Bytes>` object, cloned from `paging_state`. I'm not entirely sure how it will work under the hood, but one of two things will happen:

- the new object will increase a reference count of the memory the old `Bytes` object pointed to (semantics similar to `Arc`),
- or the new object will allocate its own memory and will copy the contents from the old `Bytes` (semantics similar to `Box`).

If it's the first one, then you will just increase and decrease the reference count unnecessarily, but it will work.

If it's the second, then it's worse. You then call `as_ptr()` to get the pointer to the contents of the temporary `Bytes`, return it, but then the temporary object is deallocated, and the pointer you just returned points to freed memory.

----
I guess that the lifetime of the memory pointed by the result of `cass_result_paging_state_token` is bound to the `CassResult`, right? If so, then you should call `as_ref().unwrap().as_ptr()`, similarly to the previous statement.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -436,6 +462,63 @@ pub unsafe extern "C" fn cass_result_first_row(result_raw: *const CassResult) ->
...
+    paging_state: *mut *const c_char,
+    paging_state_size: *mut size_t,
+) -> CassError {
+    if cass_result_has_more_pages(result) == cass_false {
+        return CassError::CASS_ERROR_LIB_NO_PAGING_STATE;
+    }
+
+    let result_from_raw = ptr_to_ref(result);
+
+    if result_from_raw.metadata.paging_state.is_none() {
```
piodul: Instead of checking for `is_none()` and then `unwrap()`-ping, I'd recommend converting the `if {} else {}` block into `match` block.

tests/src/integration/tests/test_basics.cpp
```diff
@@ -254,7 +254,7 @@ CASSANDRA_INTEGRATION_TEST_F(BasicsTests, UnsetParameters) {
 
   // Execute the insert statement and validate the error code
   Result result = session_.execute(insert_statement, false);
-  if (server_version_ >= "2.2.0") {
+  if (server_version_ >= "release:2.2.0") {
```
piodul: I'm not sure it is the best way to do it. All other tests which perform such a comparison will probably have to be updated in such fashion. Perhaps we should strip the `release:` prefix elsewhere, before `server_version_` is initialized.

@jul-stas is this problem handled in the original driver?

piodul (changes requested):

.github/workflows/build.yml
Gor027: Actually, it does not remove a test instead, it enables all `BasicsTests` including the `RowsInRowsOut` test in it.

scylla-rust-wrapper/src/query_result.rs
Gor027: I think `BytesMut` does not have such a method as `copy_from_slice` or similar. However, I have slightly changed the implementation to first ensure the necessary capacity allocated and then to create `Bytes` from a slice, so that we can avoid unnecessary reallocation.

tests/src/integration/tests/test_basics.cpp
Gor027: I have changed the implementation to set `server_version_` to `3.0.8` in case provided version has the prefix `release:`. The `cpp-driver` tests are also run with the same version on Jenkins, and some of the tests which are not disabled may be skipped: e.g. `NoCompactEnabledConnection` test in `BasicsTests`.

Gor027:
V2:

* Changed `cass_statement_set_paging_state_token` implementation to first ensure necessary capacity and then create Bytes from a slice.
* `server_version_` is statically set to `3.0.8` if the provided version has the prefix `release:`.

.github/workflows/build.yml
piodul: Oh, I understand now. The filter used to include tests which match on `BasicsTests.*RowsInRowsOut`, now it is `BasicsTests.*` which matches on a larger number of tests.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -401,6 +403,30 @@ pub unsafe extern "C" fn cass_value_get_string(
...
+    output: *mut *const cass_byte_t,
+    output_size: *mut size_t,
+) -> CassError {
+    if value.is_null() {
+        return CassError::CASS_ERROR_LIB_NULL_VALUE;
+    }
+
+    let value_from_raw: &CassValue = ptr_to_ref(value);
+
+    match &value_from_raw.value {
```
piodul, line 420: AFAIK the original `cass_value_get_bytes` works on all CQL types (see [here](https://github.com/scylladb/cpp-driver/blob/739b87c15428d1d26ec0160444b7b693543772c2/src/value.cpp#L142-L149)). I understand that it might be difficult to implement it properly right now because the rust driver doesn't allow to return raw, serialized values but rather returns CqlValue - so please add a TODO or a FIXME comment about it.

scylla-rust-wrapper/src/query_result.rs
piodul: Now it's even better than what I requested, thanks.

piodul (commented):
LGTM after adding the TODO comment.

@jul-stas please take a look, especially at the part which sets the version to "3.0.8".

piodul (approved):

PR #66, opened 2022-07-21
Add running integration tests to GitHub actions

Gor027:
This PR adds running integration tests to GitHub actions. `scylla-ccm` is used for managing clusters and tests are run on `Scylla 5.0.0`. Tests are run with `Valgrind` which will detect memory corruption and will not show details of leaked memory. For now, only `ClusterTests` are enabled to be tested on GitHub actions.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+## Pre-review checklist
+
+<!--
+    Make sure you took care of the issues on the list.
+    Put 'x' into those boxes which apply.
+    You can also create the PR now and click on all relevant checkboxes.
+-->
+
+- [ ] I have split my patch into logically separate commits.
+- [ ] All commit messages clearly explain what they change and why.
+- [ ] PR description sums up the changes and reasons why they should be introduced.
+- [ ] I have enabled appropriate tests in `.github/workflows/build.yaml` in `gtest_filter`.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,5 +14,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+
+    - name: Setup Python 3
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Setup environment
+      run: |
+        sudo sh -c "echo 'deb http://security.ubuntu.com/ubuntu xenial-security main' >> /etc/apt/sources.list"
+        sudo apt-get update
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+## Pre-review checklist
+
+<!--
+    Make sure you took care of the issues on the list.
+    Put 'x' into those boxes which apply.
+    You can also create the PR now and click on all relevant checkboxes.
+    See CONTRIBUTING.md for more details.
```
piodul: This project does not have CONTRIBUTING.md, please remove this line.

.github/workflows/build.yml
```diff
@@ -14,5 +14,26 @@ jobs:
...
+
+    - name: Setup environment
+      run: |
+        sudo sh -c "echo 'deb http://security.ubuntu.com/ubuntu xenial-security main' >> /etc/apt/sources.list"
+        sudo apt-get update
+        sudo apt-get install libssl1.0.0
+        sudo apt-get install libuv1-dev
+        sudo apt-get install libkrb5-dev
+        sudo snap install valgrind --classic
+        sudo apt-get install libc6-dbg
```
piodul: You can do all of those in one line: `sudo apt-get install libssl1.0.0 libuv1-dev ...`. This will probably be slightly faster than doing it separately.

By the way, is there a good reason why valgrind is installed via snap and not apt-get?

Gor027: Yes, `apt-get` installs an old version (3.15.0) which caused some corrupted memory errors to appear while running `ClusterTests`, in newer versions of Valgrind that was fixed. Snap installs a newer version (3.18.1).

piodul: OK. In that case, please convert the `sudo apt-get install ...` lines into one, as per my request.

Gor027: Done.

.github/workflows/build.yml
```diff
@@ -14,5 +14,23 @@ jobs:
...
+    - name: Setup Python 3
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Setup environment
+      run: |
+        sudo sh -c "echo 'deb http://security.ubuntu.com/ubuntu xenial-security main' >> /etc/apt/sources.list"
+        sudo apt-get update
+        sudo apt-get install libssl1.0.0 libuv1-dev libkrb5-dev libc6-dbg
```
jul-stas, line 27: Are `libuv` and `krb5` really needed in `-dev` version?
EDIT: ok, apparently `libuv-dev` is necessary.

piodul (approved):

PR #67, opened 2022-07-26
Import fixed version of scylla-rust-driver

Gor027:
This is dependent upon [this](https://github.com/scylladb/scylla-rust-driver/pull/486) PR in `scylla-rust-driver`

This PR imports a fixed version of `scylla-rust-driver` (`0.4.7`) instead of [this](https://github.com/hackathon-rust-cpp/scylla-rust-driver.git) fork. Also, it adds necessary modifications in order to compile successfully.

### DIFF
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -11,10 +11,10 @@ license = "MIT OR Apache-2.0"
 
 
 [dependencies]
-scylla = { git = "https://github.com/hackathon-rust-cpp/scylla-rust-driver.git", branch = "main" }
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "90044469948bc31fcc7c794d88eb245e2607a68a"}
 tokio = { version = "1.1.0", features = ["full"] }
 lazy_static = "1.4.0"
-uuid = "0.8.2"
+uuid = "1.1.2"
 machine-uid = "0.2.0"
 rand = "0.8.4"
 num-traits = "0.2"
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -6,11 +6,14 @@ impl From<&QueryError> for CassError {
     fn from(error: &QueryError) -> Self {
         match error {
             QueryError::DbError(db_error, _string) => CassError::from(db_error),
+            // FIXME: CASS_ERROR_LAST_ENTRY as placeholder for a correct error
             QueryError::BadQuery(_bad_query) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::IoError(_io_error) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::ProtocolError(_str) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::InvalidMessage(_string) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::TimeoutError => CassError::CASS_ERROR_LAST_ENTRY,
+            QueryError::TooManyOrphanedStreamIds(_) => CassError::CASS_ERROR_LAST_ENTRY,
+            QueryError::UnableToAllocStreamId => CassError::CASS_ERROR_LAST_ENTRY,
         }
     }
 }
@@ -35,7 +38,7 @@ impl From<&DbError> for CassError {
             DbError::Invalid => CassError::CASS_ERROR_SERVER_INVALID_QUERY,
             DbError::ConfigError => CassError::CASS_ERROR_SERVER_CONFIG_ERROR,
             DbError::AlreadyExists { .. } => CassError::CASS_ERROR_SERVER_ALREADY_EXISTS,
-            DbError::Unprepared => CassError::CASS_ERROR_SERVER_UNPREPARED,
+            DbError::Unprepared { .. } => CassError::CASS_ERROR_SERVER_UNPREPARED,
             DbError::Other(num) => {
                 CassError((CassErrorSource::CASS_ERROR_SOURCE_SERVER.0 << 24) | *num as u32)
             }
@@ -71,6 +74,8 @@ impl From<&NewSessionError> for CassError {
             }
             NewSessionError::InvalidMessage(_string) => CassError::CASS_ERROR_LAST_ENTRY,
             NewSessionError::TimeoutError => CassError::CASS_ERROR_LAST_ENTRY,
+            NewSessionError::TooManyOrphanedStreamIds(_) => CassError::CASS_ERROR_LAST_ENTRY,
+            NewSessionError::UnableToAllocStreamId => CassError::CASS_ERROR_LAST_ENTRY,
         }
     }
 }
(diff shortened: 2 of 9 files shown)


scylla-rust-wrapper/src/query_error.rs
```diff
@@ -61,21 +61,41 @@ pub unsafe extern "C" fn cass_error_result_consistency(
...
-        QueryError::DbError(DbError::WriteFailure { consistency, .. }, _) => {
-            CassConsistency::from(*consistency)
-        }
+        QueryError::DbError(DbError::Unavailable { consistency, .. }, _) => match consistency {
+            LegacyConsistency::Regular(regular) => CassConsistency::from(*regular),
+            LegacyConsistency::Serial(serial) => match serial {
+                SerialConsistency::Serial => CassConsistency::CASS_CONSISTENCY_SERIAL,
+                SerialConsistency::LocalSerial => CassConsistency::CASS_CONSISTENCY_LOCAL_SERIAL,
+            },
+        },
```
piodul: Why not extract the `match consistency { ... }` part to a helper function/lambda? It's repeated several times.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -244,8 +244,20 @@ pub unsafe extern "C" fn cass_value_get_uuid(
     let val: &CassValue = ptr_to_ref(value);
     let out: &mut CassUuid = ptr_to_ref_mut(output);
     match val {
-        Some(CqlValue::Uuid(uuid)) => *out = (*uuid).into(),
-        Some(CqlValue::Timeuuid(uuid)) => *out = (*uuid).into(),
```
piodul: This should probably be not fixed like this. Look at `scylla-rust-wrapper/src/uuid.rs` - there is a `impl From<Uuid> for CassUuid` conversion defined by us. I'm not sure if it works the same as what you do here, and it would be nice to keep using it

I suspect that the `into()` conversion stopped working because you bumped the rust driver crate version before bumping the uuid crate version. The driver requires `uuid = 1.0`, but the wrapper requires `uuid = 0.8.2`; those versions are not compatible, so the Uuid struct used by the driver is __not__ the same thing as Uuid used in the wrapper.

I can see that you bumped the uuid requirement to `1.1.2`, which should be compatible with `1.0` required by the driver. IMO, the `.into()` conversion should work. Could you check?

Gor027: Yes, seems that it works with the uuid `1.1.2` version.

scylla-rust-wrapper/src/binding.rs
```diff
@@ -253,7 +253,7 @@ macro_rules! invoke_binder_maker_macro_with_type {
             $this,
             $consume_v,
             $fn,
-            |v: crate::uuid::CassUuid| Ok(Some(Uuid(v.into()))),
+            |v: crate::uuid::CassUuid| Ok(Some(Uuid(uuid::Uuid::from_u64_pair(v.time_and_version, v.clock_seq_and_node)))),
```
piodul: `.into()` should work now, so this is unnecessary (and probably incorrect).

scylla-rust-wrapper/src/query_error.rs
```diff
@@ -62,24 +62,34 @@ pub unsafe extern "C" fn cass_error_result_consistency(
...
         }
         QueryError::DbError(DbError::WriteFailure { consistency, .. }, _) => {
-            CassConsistency::from(*consistency)
+            get_cass_consistency_from_legacy(consistency)
         }
         _ => CassConsistency::CASS_CONSISTENCY_UNKNOWN,
     }
 }
 
+unsafe fn get_cass_consistency_from_legacy(consistency: &LegacyConsistency) -> CassConsistency {
```
piodul: Does this function need to be unsafe?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -13,8 +13,8 @@ use std::sync::Arc;
 
 #[derive(Clone)]
 enum CassClusterChildLoadBalancingPolicy {
-    RoundRobinPolicy(RoundRobinPolicy),
-    DcAwareRoundRobinPolicy(DcAwareRoundRobinPolicy),
+    RoundRobinPolicy,
+    DcAwareRoundRobinPolicy { local_dc: &'static str },
```
piodul: Using `&'static str` is a bad idea here. In the original implementation, when you create a DC-aware policy via `cass_cluster_set_load_balance_dc_aware_n`, that function makes a copy of the string. You should use `String` here instead so that the `CassClusterChildLoadBalancingPolicy` manages its own copy of the string.

piodul (changes requested):

scylla-rust-wrapper/Cargo.toml
```diff
@@ -11,10 +11,10 @@ license = "MIT OR Apache-2.0"
 
 
 [dependencies]
-scylla = { git = "https://github.com/hackathon-rust-cpp/scylla-rust-driver.git", branch = "main" }
+scylla = "0.4.7"
```
piodul: The fixes you sent for the driver are not present in 0.4.7 or any other version yet. You will have to point to a specific commit in the driver's repo.

scylla-rust-wrapper/Cargo.toml
```diff
@@ -11,10 +11,10 @@ license = "MIT OR Apache-2.0"
 
 
 [dependencies]
-scylla = { git = "https://github.com/hackathon-rust-cpp/scylla-rust-driver.git", branch = "main" }
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", branch = "main" }
```
piodul: Pointing to a branch is risky, because over time more commits are pushed and suddenly the project won't compile anymore. I'd recommend pointing to current latest commit of the main branch. See here: https://stackoverflow.com/a/54197014

piodul (changes requested):

scylla-rust-wrapper/Cargo.toml
Gor027: I pointed first to the exact commit [here](https://github.com/scylladb/cpp-rust-driver/commit/3ef3c790bc668f870ea80b33abf8504aa8064549), then changed to the `main` branch, thinking that maybe we should keep up with the new changes in the Rust driver.

piodul (approved):

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -11,6 +11,8 @@ impl From<&QueryError> for CassError {
             QueryError::ProtocolError(_str) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::InvalidMessage(_string) => CassError::CASS_ERROR_LAST_ENTRY,
             QueryError::TimeoutError => CassError::CASS_ERROR_LAST_ENTRY,
+            QueryError::TooManyOrphanedStreamIds(_) => CassError::CASS_ERROR_LAST_ENTRY,
```
constantfold, line 15: Although this problem was not introduced by you, `CASS_ERROR_LAST_ENTRY` seems like a placeholder for a correct error mapping. Please add "FIXME: CASS_ERROR_LAST_ENTRY as placeholder for a correct error" so that we don't forget about it.

constantfold (approved):
LGTM, apart from my one comment.

PR #68, opened 2022-07-28
Prepared tests

Gor027:
## Pre-review checklist

This PR is dependent upon #65

Adds necessary implementations to pass all `PreparedTests` besides `PreparedIDUnchangedDuringReprepare`, which is also ignored in the C++ driver.
Certain validations in tests are commented out for now, as their corresponding implementations in Rust are not yet present.
Also, `DowngradingConsistency` retry policies in those tests are changed to be `Default` retry consistency, as `DowngradingConsistency` is deprecated and not supported by the Rust driver.

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -15,6 +15,7 @@ use scylla::transport::errors::QueryError;
 use scylla::{QueryResult, Session};
 use std::os::raw::c_char;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::RwLock;
 
 pub type CassSession = RwLock<Option<Session>>;
@@ -64,10 +65,11 @@ pub unsafe extern "C" fn cass_session_execute(
     let statement_opt = ptr_to_ref(statement_raw);
     let paging_state = statement_opt.paging_state.clone();
     let bound_values = statement_opt.bound_values.clone();
+    let request_timeout_ms = statement_opt.request_timeout_ms;
 
     let statement = statement_opt.statement.clone();
 
-    CassFuture::make_raw(async move {
+    let future = async move {
         let session_guard = session_opt.read().await;
         if session_guard.is_none() {
             return Err((
@@ -104,7 +106,19 @@ pub unsafe extern "C" fn cass_session_execute(
             }
             Err(err) => Ok(CassResultValue::QueryError(Arc::new(err))),
         }
-    })
+    };
+
+    match request_timeout_ms {
+        Some(timeout_ms) => CassFuture::make_raw(async move {
+            match tokio::time::timeout(Duration::from_millis(timeout_ms), future).await {
+                Ok(result) => result,
+                Err(_timeout_err) => Ok(CassResultValue::QueryError(Arc::new(
+                    QueryError::TimeoutError,
+                ))),
+            }
+        }),
+        None => CassFuture::make_raw(future),
+    }
 }
 
 fn create_cass_rows_from_rows(
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -149,6 +149,40 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn cass_session_prepare_from_existing(
+    cass_session: *mut CassSession,
+    statement: *const CassStatement,
+) -> *const CassFuture {
+    let session = ptr_to_ref(cass_session);
+    let cass_statement = ptr_to_ref(statement);
+    let statement = cass_statement.statement.clone();
+
+    CassFuture::make_raw(async move {
+        let query = match &statement {
+            Statement::Simple(q) => q,
+            Statement::Prepared(ps) => {
+                return Ok(CassResultValue::Prepared(ps.clone()));
+            }
+        };
+
+        let session_guard = session.read().await;
+        if session_guard.is_none() {
+            return Err((
+                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
+                "Session is not connected".msg(),
+            ));
+        }
+        let session = session_guard.as_ref().unwrap();
+        let prepared = session
+            .prepare(query.clone())
+            .await
+            .map_err(|err| (CassError::from(&err), err.msg()))?;
+
+        Ok(CassResultValue::Prepared(Arc::new(prepared)))
+    })
+}
+
+#[no_mangle]
 pub unsafe extern "C" fn cass_session_prepare(
     session: *mut CassSession,
     query: *const c_char,
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -1,6 +1,7 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::query_result::CassResult;
+use crate::retry_policy::CassRetryPolicy;
 use crate::types::*;
 use scylla::frame::response::result::CqlValue;
 use scylla::frame::types::LegacyConsistency::{Regular, Serial};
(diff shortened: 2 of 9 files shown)


tests/src/integration/tests/test_prepared.cpp
```diff
@@ -138,10 +138,10 @@ CASSANDRA_INTEGRATION_TEST_F(PreparedTests, PrepareFromExistingSimpleStatement)
   Statement bound_statement = session_.prepare_from_existing(statement).bind();
 
   // Validate that the bound statement inherited the settings from the original statement
-  EXPECT_EQ(bound_statement.consistency(), CASS_CONSISTENCY_LOCAL_QUORUM);
-  EXPECT_EQ(bound_statement.serial_consistency(), CASS_CONSISTENCY_SERIAL);
-  EXPECT_EQ(bound_statement.request_timeout_ms(), 99999u);
-  EXPECT_EQ(bound_statement.retry_policy(), retry_policy.get());
+//  EXPECT_EQ(bound_statement.consistency(), CASS_CONSISTENCY_LOCAL_QUORUM);
```
constantfold, line 142: Missing `// TODO` comment

scylla-rust-wrapper/src/statement.rs
```diff
@@ -167,6 +178,82 @@ pub unsafe extern "C" fn cass_statement_set_tracing(
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_statement_set_request_timeout(
+    statement: *mut CassStatement,
+    timeout_ms: cass_uint64_t,
+) -> CassError {
+    match &mut ptr_to_ref_mut(statement).statement {
+        Statement::Simple(inner) => inner.set_timestamp(Some(timeout_ms as i64)),
+        Statement::Prepared(inner) => Arc::make_mut(inner).set_timestamp(Some(timeout_ms as i64)),
```
piodul: Timestamp is a completely different thing than timeout.

- A CQL timestamp defines the order in which writes are applied to the database.
- Timeout (here, client-side timeout) means the time the driver will wait until response arrives from the database. If no response arrives within the timeout period, the driver is supposed to return an error.

Here, I think you have accidentally implemented `cass_statement_set_timestamp`.

Currently, the rust driver doesn't support client timeouts directly, but you can implement them yourself in the wrapper code. You need to keep the timeout in the CassStatement somewhere, wrap `session.execute()` etc. calls with [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) and handle the timeout error returned by the `tokio::time::timeout` call appropriately.

scylla-rust-wrapper/src/session.rs
```diff
@@ -134,6 +134,83 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
         .collect()
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_session_prepare_from_existing(
+    cass_session: *mut CassSession,
+    statement: *const CassStatement,
+) -> *const CassFuture {
+    let session_opt = ptr_to_ref(cass_session);
+    let statement_opt = ptr_to_ref(statement);
```
piodul: Why do those names have an `_opt` suffix? Those aren't `Option`s?

scylla-rust-wrapper/src/session.rs
```diff
@@ -134,6 +134,83 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
...
+    let statement = statement_opt.statement.clone();
+
+    let query = match statement.clone() {
+        Statement::Simple(q) => q.contents,
+        Statement::Prepared(ps) => ps.get_statement().to_string(),
+    };
```
piodul: By doing `statement.clone()`, you are cloning the whole statement, but then you only need a small part of it and you have unnecessarily cloned the whole statement.

Perhaps you can match on a reference (`&statement`) and move `.clone()` to the first match, like this:

```rust
let query = match &statement {
    Statement::Simple(q) => q.contents.clone(),
    Statement::Prepared(ps) => ps.get_statement().to_string(),
};
```

scylla-rust-wrapper/src/session.rs
```diff
@@ -134,6 +134,83 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
...
+    cass_session: *mut CassSession,
+    statement: *const CassStatement,
+) -> *const CassFuture {
+    let session_opt = ptr_to_ref(cass_session);
+    let statement_opt = ptr_to_ref(statement);
+    let statement = statement_opt.statement.clone();
+
+    let query = match statement.clone() {
+        Statement::Simple(q) => q.contents,
+        Statement::Prepared(ps) => ps.get_statement().to_string(),
```
piodul: I don't think it makes sense at all to prepare a prepared statement again. I think in this case it makes sense to just clone a prepared statement, put it into a box and immediately return it.

scylla-rust-wrapper/src/session.rs
```diff
@@ -134,6 +134,83 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
...
+        let session_guard = session_opt.read().await;
+        if session_guard.is_none() {
+            return Err((
+                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
+                "Session is not connected".msg(),
+            ));
+        }
+        let session = session_guard.as_ref().unwrap();
+        let mut prepared = session
+            .prepare(query)
```
piodul: Instead of passing a string, you can put a `Query` here. AFAIK the resulting `PreparedStatement` should inherit all properties of the original `Query`, so setting them manually later shouldn't be necessary.

scylla-rust-wrapper/src/session.rs
```diff
@@ -134,6 +136,16 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
         .collect()
 }
 
+fn create_col_index_mapping(col_specs: &[ColumnSpec]) -> HashMap<String, usize> {
```
piodul: I think that constructing a hashmap to speed up lookup by name is an overkill. The number of columns is usually small, so linear lookup should be sufficient. Creating a hashmap allocates (at least one allocation for the hashmap itself + one allocation for each String), so it would be best to avoid it.

piodul (changes requested):

scylla-rust-wrapper/src/session.rs
```diff
@@ -64,45 +65,55 @@ pub unsafe extern "C" fn cass_session_execute(
...
-                let cass_rows = create_cass_rows_from_rows(result.rows, &metadata);
-                let cass_result: CassResult_ = Arc::new(CassResult {
-                    rows: cass_rows,
-                    metadata,
-                });
-
-                Ok(CassResultValue::QueryResult(cass_result))
-            }
-            Err(err) => Ok(CassResultValue::QueryError(Arc::new(err))),
+        match tokio::time::timeout(Duration::from_millis(request_timeout_ms), future).await {
```
constantfold: In the future it should be replaced with driver's own client-side timeout implementation, which is a planned feature for the driver itself (https://github.com/scylladb/scylla-rust-driver/issues/304). 

cc: @wprzytula

scylla-rust-wrapper/src/statement.rs
```diff
@@ -25,6 +27,7 @@ pub struct CassStatement {
     pub statement: Statement,
     pub bound_values: Vec<MaybeUnset<Option<CqlValue>>>,
     pub paging_state: Option<Bytes>,
+    pub request_timeout_ms: cass_uint64_t,
```
constantfold: Maybe replace it with `Option<cass_uint64_t>` - if option is `None` then apply no timeout, instead of hardcoding 2 year timeout.

scylla-rust-wrapper/src/session.rs
```diff
@@ -145,6 +145,40 @@ fn create_cass_row_columns(row: Row, metadata: &Arc<CassResultData>) -> Vec<Cass
...
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_session_prepare_from_existing(
+    cass_session: *mut CassSession,
+    statement: *const CassStatement,
+) -> *const CassFuture {
+    let session_from_raw = ptr_to_ref(cass_session);
+    let statement_from_raw = ptr_to_ref(statement);
+    let statement = statement_from_raw.statement.clone();
```
constantfold: Small naming nit, maybe replace it with:

```
    let session = ptr_to_ref(cass_session);
    let statement = ptr_to_ref(statement);
    let statement = statement_from_raw.statement.clone();
```

scylla-rust-wrapper/src/session.rs
wprzytula: I confirm, I'm going to work on such implementation soon.

constantfold: And to be clear, we won't be waiting for this driver support before merging this PR - it's good as-is.

scylla-rust-wrapper/src/session.rs
Gor027: Do you mean to rename `session_from_raw` and `statement_from_raw` to `session` and `statement`?

constantfold: Yes, but it looks OK after the latest commits.

PR #69, opened 2022-08-09
Cassandra types tests

Gor027:
## Pre-review checklist

This PR is dependent on #68.

Adds necessary implementation to pass all `CassandraTypesTests` excluding `UDT` tests as it requires an implementation to get cluster schema metadata, so it will be done in another PR.
Tests are modified to properly iterate over collections to check their values and size.
As `CassandraTypesTests` tests collections with all possible types, overall `208` tests are enabled. Only tests for `Decimal` and `Varint` types are disabled, as `invoke_binder_maker_macro_with_type` is not yet implemented for `Decimal` type and `cass_value_get_bytes` is not implemented for `Varint`. The latter should be implemented for all types, however, it requires the rust driver to return raw bytes of values instead of `CqlValue`.

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -74,10 +74,17 @@ pub struct CassCollectionIterator {
     position: Option<usize>,
 }
 
+pub struct CassMapIterator {
+    value: CassValue_,
+    count: u64,
+    position: Option<usize>,
+}
+
 pub enum CassIterator {
     CassResultIterator(CassResultIterator),
     CassRowIterator(CassRowIterator),
     CassCollectionIterator(CassCollectionIterator),
+    CassMapIterator(CassMapIterator),
 }
 
 #[no_mangle]
@@ -116,6 +123,13 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
             collection_iterator.position = Some(new_pos);
 
             (new_pos < collection_iterator.count.try_into().unwrap()) as cass_bool_t
+        }
+        CassIterator::CassMapIterator(map_iterator) => {
+            let new_pos: usize = map_iterator.position.map_or(0, |prev_pos| prev_pos + 1);
+
+            map_iterator.position = Some(new_pos);
+
+            (new_pos < map_iterator.count.try_into().unwrap()) as cass_bool_t
         }
     }
 }
@@ -204,6 +218,60 @@ pub unsafe extern "C" fn cass_iterator_get_value(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn cass_iterator_get_map_key(
+    iterator: *const CassIterator,
+) -> *const CassValue {
+    let iter = ptr_to_ref(iterator);
+
+    if let CassIterator::CassMapIterator(map_iterator) = iter {
+        let iter_position = match map_iterator.position {
+            Some(pos) => pos,
+            None => return std::ptr::null(),
+        };
+
+        let entry = match &map_iterator.value.value {
+            Some(Value::CollectionValue(Collection::Map(map))) => map.get(iter_position),
+            _ => return std::ptr::null(),
+        };
+
+        if entry.is_none() {
+            return std::ptr::null();
+        }
+
+        return &entry.unwrap().0 as *const CassValue;
+    }
+
+    std::ptr::null()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_iterator_get_map_value(
+    iterator: *const CassIterator,
+) -> *const CassValue {
+    let iter = ptr_to_ref(iterator);
+
+    if let CassIterator::CassMapIterator(map_iterator) = iter {
+        let iter_position = match map_iterator.position {
+            Some(pos) => pos,
+            None => return std::ptr::null(),
+        };
+
+        let entry = match &map_iterator.value.value {
+            Some(Value::CollectionValue(Collection::Map(map))) => map.get(iter_position),
+            _ => return std::ptr::null(),
+        };
+
+        if entry.is_none() {
+            return std::ptr::null();
+        }
+
+        return &entry.unwrap().1 as *const CassValue;
+    }
+
+    std::ptr::null()
+}
+
+#[no_mangle]
 pub unsafe extern "C" fn cass_iterator_from_result(result: *const CassResult) -> *mut CassIterator {
     let result_from_raw: CassResult_ = clone_arced(result);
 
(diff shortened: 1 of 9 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1,6 +1,6 @@
 use crate::argconv::*;
```
constantfold, line 1: Question about the commit message:
"Change CassValue to contain full information about collections
Previously CassValue contained value of type Option\<CassValue> which did
not allow to retrieve type information of collection elements."

What did you mean by "Previously CassValue contained value of type Option\<CassValue>"? Did you mean `Option<CqlValue>`?

I'd also appreciate a stronger explanation of why `CqlValue` was not enough - an example of type information missing ("which did not allow to retrieve type information of collection elements").

Gor027: Yes, it should have been `Option<CqlValue>`. The reason to store the `CassDataType` in `CassValue` is that the lifetime of the type is bound to the lifetime of the value, and also, `CqlValue` is not enough to retrieve complete type information of a value, particularly for collections. So, when processing the values of columns in a query result, we should properly allocate `CassDataType` and then bind it to the `CassValue`.
Edit: Thanks for noticing it, already updated the commit message.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -12,18 +12,25 @@ use scylla::query::Query;
...
-    Simple(Query),
+    Simple(SimpleQuery),
     // Arc is needed, because PreparedStatement is passed by reference to session.execute
     Prepared(Arc<PreparedStatement>),
 }
 
+#[derive(Clone)]
+pub struct SimpleQuery {
+    pub query: Query,
+    pub name_to_bound_index: HashMap<String, usize>,
```
constantfold, line 31: Was this map added only for performance reasons? If yes, maybe this optimization (and complexity it adds) isn't needed - the user is using SimpleQuery instead of PreparedStatement, so they don't care about the performance.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -26,6 +26,8 @@ pub type CassResult_ = Arc<CassResult>;
 
 pub type CassRow_ = &'static CassRow;
 
+pub type CassValue_ = &'static CassValue;
```
constantfold, line 32: Please add a comment about the lifetimes: that `CassRow_` and `CassValue_` live as long as `CassResult_` lives (am I correct?)

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -394,6 +484,13 @@ pub unsafe extern "C" fn cass_value_get_int64(
     match val.value {
         Some(Value::RegularValue(CqlValue::BigInt(i))) => *out = i,
         Some(Value::RegularValue(CqlValue::Counter(i))) => *out = i.0 as cass_int64_t,
+        Some(Value::RegularValue(CqlValue::Time(d))) => match d.num_nanoseconds() {
+            Some(nanos) => *out = nanos as cass_int64_t,
+            None => return CassError::CASS_ERROR_LIB_NULL_VALUE,
+        },
+        Some(Value::RegularValue(CqlValue::Timestamp(d))) => {
+            *out = d.num_milliseconds() as cass_int64_t
+        }
```
constantfold, line 587: Why is this code in "Add collection iterator implementation" commit?

tests/src/integration/tests/test_cassandra_types.cpp
```diff
@@ -542,8 +561,27 @@ CASSANDRA_INTEGRATION_TYPED_TEST_P(CassandraTypesTests, Set) {
     }
     Result result = this->session_.execute(select_statement);
     ASSERT_EQ(1u, result.row_count());
-    Set<TypeParam> result_set = result.first_row().next().as<Set<TypeParam> >();
-    ASSERT_EQ(set.value(), result_set.value());
+    const CassValue *result_set(result.first_row().next().get_value());
```
constantfold, line 568: Small nit about commit message:
"Modified tests to properly iterate over collections and check values
and collection size."

I'd describe in more detail that previous to this patch, ~~only the first element was checked~~ what was the previous check? "properly iterate" suggest that code used to "improperly iterate".

tests/src/integration/tests/test_cassandra_types.cpp
```diff
@@ -632,9 +632,28 @@ CASSANDRA_INTEGRATION_TYPED_TEST_P(CassandraTypesTests, Map) {
       select_statement.bind<Map<TypeParam, TypeParam> >(0, map);
       Result result = this->session_.execute(select_statement);
       ASSERT_EQ(1u, result.row_count());
-      Column column = result.first_row().next();
-      Map<TypeParam, TypeParam> result_map(column.as<Map<TypeParam, TypeParam> >());
-      ASSERT_EQ(map_values, result_map.value());
```
constantfold, line 599: Maybe let's leave the old test code (alongside the new code). That way, it'll be clear and beyond any doubt that this change is a superset of previous change, not some modification that removes some checks.

scylla-rust-wrapper/src/statement.rs
Gor027: Actually, the implementation with a map is not for optimization purposes, I could have used `Vec<(String, usize)>` instead of it, but I think `HashMap` is more straightforward. Simple queries do not have any metadata about the columns, like `col_specs` in `PreparedMetadata`, so we need to somehow remember the position of the bound value in `bound_values` in `CassStatement`.

scylla-rust-wrapper/src/query_result.rs
Gor027: It seems to be a change for `CassValue`, however, the `CollectionIterator` should work properly for `Time` and `Timestamp` types, too. So, for example, calling `cass_value_get_int64(cass_iterator_get_value(collection_iterator), output)` should set the `Time` or `Timestamp` value in output. This is a change mainly for testing `CollectionIterator` against all possible types.

scylla-rust-wrapper/src/query_result.rs
Gor027: Yes. `CassRow` lives as long as `CassResult`, and `CassValue`'s lifetime is bound to `CassRow`.

constantfold:
Interestingly, BasicsTests.Integration_Cassandra_Timestamps failed in the latest run (https://github.com/scylladb/cpp-rust-driver/actions/runs/2890498078/attempts/1). It looks like it's a bit time-sensitive and might be flaky on slower machines (such as GitHub Actions workers). 

@Gor027 Have you seen this before?

Gor027:
> 

Yes, I was aware of it. Maybe we can manually increase the acceptable difference in the test, to avoid failures because of slow machines?

constantfold:
The `Integration_Cassandra_Timestamps` will be fixed in a follow-up PR as agreed.

PR #70, opened 2022-08-16
Add SSL/TLS support

Gor027:
## Pre-review checklist

This PR is dependent upon #69

This PR adds SSL/TLS support to the driver.
Integration tests that are enabled in the C++ driver and are actually run during `Jenkins` builds, are also enabled for the `cpp-rust-driver`. One of the reasons that the SSL tests are disabled for `cpp-driver` is that during the setup `scylla-ccm` fails while trying to start a Scylla cluster. So, to make sure the SSL properly works, I have also run the integration tests on a Cassandra cluster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
 :CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_*\
 :BatchSingleNodeClusterTests*:BatchCounterSingleNodeClusterTests*:BatchCounterThreeNodeClusterTests*\
 :ErrorTests.*\
+:SslNoClusterTests*:SslNoSslOnClusterTests*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :*5.Integration_Cassandra_*\
 :*19.Integration_Cassandra_*\
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -3,8 +3,11 @@ use crate::cass_error::CassError;
 use crate::future::CassFuture;
 use crate::retry_policy::CassRetryPolicy;
 use crate::retry_policy::RetryPolicy::*;
+use crate::ssl::CassSsl;
 use crate::types::*;
 use core::time::Duration;
+use openssl::ssl::SslContextBuilder;
+use openssl_sys::SSL_CTX_up_ref;
 use scylla::load_balancing::{
     DcAwareRoundRobinPolicy, LoadBalancingPolicy, RoundRobinPolicy, TokenAwarePolicy,
 };
@@ -423,4 +426,16 @@ pub unsafe extern "C" fn cass_cluster_set_retry_policy(
     let boxed_retry_policy = retry_policy.clone_boxed();
 
     cluster.session_builder.config.retry_policy = boxed_retry_policy;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_ssl(cluster: *mut CassCluster, ssl: *mut CassSsl) {
+    let cluster_from_raw = ptr_to_ref_mut(cluster);
+    let cass_ssl = clone_arced(ssl);
+
+    let ssl_context_builder = SslContextBuilder::from_ptr(cass_ssl.ssl_context);
+    // Reference count is increased as tokio_openssl will try to free `ssl_context` when calling `SSL_free`.
+    SSL_CTX_up_ref(cass_ssl.ssl_context);
+
+    cluster_from_raw.session_builder.config.ssl_context = Some(ssl_context_builder.build());
 }
--- a/scylla-rust-wrapper/src/ssl.rs
+++ b/scylla-rust-wrapper/src/ssl.rs
@@ -1,0 +1,289 @@
+use crate::argconv::{clone_arced, free_arced};
+use crate::cass_error::CassError;
+use crate::types::size_t;
+use libc::{c_int, strlen};
+use openssl::ssl::SslVerifyMode;
+use openssl_sys::{
+    BIO_free_all, BIO_new_mem_buf, EVP_PKEY_free, PEM_read_bio_PrivateKey, PEM_read_bio_X509,
+    SSL_CTX_add_extra_chain_cert, SSL_CTX_free, SSL_CTX_new, SSL_CTX_set_cert_store,
+    SSL_CTX_set_verify, SSL_CTX_use_PrivateKey, SSL_CTX_use_certificate, TLS_method,
+    X509_STORE_add_cert, X509_STORE_new, X509_free, BIO, SSL_CTX, X509_STORE,
+};
+use std::convert::TryInto;
+use std::os::raw::c_char;
+use std::os::raw::c_void;
+use std::sync::Arc;
+
+pub struct CassSsl {
+    pub(crate) ssl_context: *mut SSL_CTX,
+    pub(crate) trusted_store: *mut X509_STORE,
+}
+
+pub const CASS_SSL_VERIFY_NONE: i32 = 0x00;
+pub const CASS_SSL_VERIFY_PEER_CERT: i32 = 0x01;
+pub const CASS_SSL_VERIFY_PEER_IDENTITY: i32 = 0x02;
+pub const CASS_SSL_VERIFY_PEER_IDENTITY_DNS: i32 = 0x04;
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_new() -> *const CassSsl {
+    openssl_sys::init();
+    cass_ssl_new_no_lib_init()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_new_no_lib_init() -> *const CassSsl {
+    let ssl_context: *mut SSL_CTX = SSL_CTX_new(TLS_method());
+    let trusted_store: *mut X509_STORE = X509_STORE_new();
+
+    SSL_CTX_set_cert_store(ssl_context, trusted_store);
+    SSL_CTX_set_verify(ssl_context, CASS_SSL_VERIFY_NONE, None);
+
+    let ssl = CassSsl {
+        ssl_context,
+        trusted_store,
+    };
+
+    Arc::into_raw(Arc::new(ssl)) as *const CassSsl
+}
+
+impl Drop for CassSsl {
+    fn drop(&mut self) {
+        unsafe {
+            SSL_CTX_free(self.ssl_context);
+        }
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_free(ssl: *mut CassSsl) {
+    free_arced(ssl);
+}
+
+unsafe extern "C" fn pem_password_callback(
+    buf: *mut c_char,
+    size: c_int,
+    _rwflag: c_int,
+    u: *mut c_void,
+) -> c_int {
+    if u.is_null() {
+        return 0;
+    }
+
+    let len = strlen(u as *const c_char);
+    if len == 0 {
+        return 0;
+    }
+
+    let mut to_copy = size;
+    if len < to_copy.try_into().unwrap() {
+        to_copy = len as c_int;
+    }
+
+    // Same as: memcpy(buf, u, to_copy);
+    std::ptr::copy_nonoverlapping(u as *const c_char, buf, to_copy as usize);
+
+    len as c_int
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_add_trusted_cert(
+    ssl: *mut CassSsl,
+    cert: *const c_char,
+) -> CassError {
+    if cert.is_null() {
+        return CassError::CASS_ERROR_SSL_INVALID_CERT;
+    }
+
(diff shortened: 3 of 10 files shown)


scylla-rust-wrapper/src/ssl.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub trusted_cert: Option<X509>,
+    pub verify_flags: Option<SslVerifyMode>,
+    pub cert_chain: Option<Vec<X509>>,
+    pub private_key: Option<PKey<Private>>,
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_new() -> *const CassSsl {
+    cass_ssl_new_no_lib_init()
+}
```
piodul: The difference between `cass_ssl_new` and `cass_ssl_new_no_lib_init` is that the first is supposed to perform some additional initialization. What does this initialization do in the original implementation?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -403,3 +406,44 @@ pub unsafe extern "C" fn cass_cluster_set_token_aware_routing(
...
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_ssl(cluster: *mut CassCluster, ssl: *mut CassSsl) {
+    let cluster_from_raw = ptr_to_ref_mut(cluster);
+    let cass_ssl = ptr_to_ref(ssl);
+
+    let mut ssl_context_builder = SslContextBuilder::new(SslMethod::tls_client()).unwrap();
+
+    if cass_ssl.trusted_cert.is_some() {
```
piodul: Use `if let Some(trusted_cert) = &cass_ssl.trusted_cert` please, instead of `is_some()` + `unwrap()`. The less places which can potentially panic in the code, the better.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -403,3 +406,44 @@ pub unsafe extern "C" fn cass_cluster_set_token_aware_routing(
...
+
+    if cass_ssl.verify_flags.is_some() {
+        ssl_context_builder.set_verify(cass_ssl.verify_flags.unwrap());
+    }
+
+    if cass_ssl.cert_chain.is_some() {
+        // This should contain the entire certificate chain starting with the certificate itself.
+        let cert_chain = cass_ssl.cert_chain.as_ref().unwrap();
+        let cert = cert_chain.get(0).unwrap();
+        ssl_context_builder.set_certificate(cert).unwrap();
```
piodul: There are some `ssl_context_builder.set_xyz` operations here which can fail, and you `unwrap` them. Are they supposed to always succeed? If they do, please add comments which explain that. If not, then I think that the implementation need to be rethought so that those operations are performed in functions that can return errors.

It is worth noting that the original implementation's `cass_cluster_set_ssl` is very simple and doesn't call any functions from the `openssl` library. If we are using openssl directly like the original impl, then maybe our implementation of the SSL-related functions can mirror the original ones more closely?

piodul (changes requested):

scylla-rust-wrapper/src/ssl.rs
Gor027: Yes, in `cass_ssl_new`, `openssl::init()` should be explicitly called. `SslContextBuilder::new()` does that anyway, so I thought that should be enough.

scylla-rust-wrapper/src/cluster.rs
Gor027: The C++ driver implementation creates an `OpenSslContext` and sets the necessary configuration explicitly, however, `rust-openssl` obliges to use the `SslContextBuilder` to do so, and then consume the builder to create a `SslContext`. Another limitation is that `SslContextBuilder` cannot be cloned, so it makes us create the builder once and immediately consume it, as it is done in the `cass_cluster_set_ssl`. The `SslContext` setters, e.g. `cass_ssl_add_trusted_cert`, `cass_ssl_set_verify_flags`, validate the user-provided configuration, and `cass_cluster_set_ssl` only sets them in the `SslContextBuilder`, so I think those `unwrap` calls should always succeed.

Gor027:
V2:

The SSL support is reimplemented, this time based on the `openssl-sys` crate which provides rust bindings for the original `OpenSSL` library. With this approach, it is possible to mirror the behavior of the SSL-related functions more closely. As before, the functionality was tested manually on ScyllaDB (some integration tests are enabled), and it was also tested on a Cassandra cluster.

.github/workflows/build.yml
```diff
@@ -39,6 +39,7 @@ jobs:
 :BasicsTests.*\
 :PreparedTests.*\
 :CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_*\
+:SslNoClusterTests*:SslNoSslOnClusterTests*\
```
constantfold, line 44: @Gor027 Also run tests with Cassandra (not only Scylla 5.0.0) so that we can test (almost?) all SSL tests.

Gor027: After merging [this](https://github.com/scylladb/cpp-rust-driver/pull/74) PR, I will add running SSL-related tests on Cassandra.

scylla-rust-wrapper/src/ssl.rs
```diff
@@ -0,0 +1,264 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_new_no_lib_init() -> *const CassSsl {
+    let ssl_context: *mut SSL_CTX = SSL_CTX_new(TLS_method());
+    let trusted_store: *mut X509_STORE = X509_STORE_new();
+
+    SSL_CTX_set_cert_store(ssl_context, trusted_store);
+
```
piodul, line 40: There are some calls missing from the original implementation here:

```c++
SSL_CTX_set_verify(ssl_ctx_, SSL_VERIFY_NONE, ssl_no_verify_callback);
// Limit to TLS 1.2 for now. TLS 1.3 has broken the handshake code.
SSL_CTX_set_max_proto_version(ssl_ctx_, TLS1_2_VERSION);
```

scylla-rust-wrapper/src/ssl.rs
```diff
@@ -0,0 +1,264 @@
...
+    X509_free(x509);
+
+    CassError::CASS_OK
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_set_verify_flags(ssl: *mut CassSsl, flags: i32) {
+    let ssl = clone_arced(ssl);
+
+    SSL_CTX_set_verify(ssl.ssl_context, flags, None);
```
piodul: I don't think this is correct. The flags should be of type `CassSslVerifyFlags`, which is a custom enum defined by the driver, which I'm not sure if are a valid argument to `SSL_CTX_set_verify`. The original implementation seems to do something different - it stores the flags in a field which is later used in a custom verification logic.

piodul (changes requested):
There seems to be some error checking/reporting logic missing from the ported function (`ssl_log_errors`, `ERR_peek_error`).

scylla-rust-wrapper/src/ssl.rs
Gor027: I guess `SSL_VERIFY_NONE` is the default mode, but I agree, that it is better to set it manually. I am not sure though, whether should we limit the maximum protocol version, as it works fine currently, perhaps with the latest protocol version available in `openssl-sys`.

scylla-rust-wrapper/src/ssl.rs
Gor027: The `CassSslVerifyFlags` partially matches with the expected argument for the `SSL_CTX_set_verify`. The C++ driver overrides the OpenSSL handshake implementation to verify the client's hostname and IP address, however, OpenSSL 1.0.2 and newer are able to verify hostname and IP addresses themselves. So, the flags `CASS_SSL_VERIFY_PEER_IDENTITY` and `CASS_SSL_VERIFY_PEER_IDENTITY_DNS` can be either removed or considered to be `CASS_SSL_VERIFY_PEER_CERT`. Although, I have tested setting those 2 additional flags while establishing a connection with a Scylla and Cassandra cluster, and it seemed to work.

scylla-rust-wrapper/src/ssl.rs
piodul: OK, I agree about the TLS version. The original comment is probably quite old.

scylla-rust-wrapper/src/ssl.rs
piodul: Here is how CassSslVerifyFlags is defined:

```c
typedef enum CassSslVerifyFlags_ {
  CASS_SSL_VERIFY_NONE              = 0x00,
  CASS_SSL_VERIFY_PEER_CERT         = 0x01,
  CASS_SSL_VERIFY_PEER_IDENTITY     = 0x02,
  CASS_SSL_VERIFY_PEER_IDENTITY_DNS = 0x04
} CassSslVerifyFlags;
```

And here is how the flags for the `mode` argument of `SSL_CTX_set_verify` are defined:

```c
# define SSL_VERIFY_NONE                 0x00
# define SSL_VERIFY_PEER                 0x01
# define SSL_VERIFY_FAIL_IF_NO_PEER_CERT 0x02
# define SSL_VERIFY_CLIENT_ONCE          0x04
# define SSL_VERIFY_POST_HANDSHAKE       0x08
```

According to the man page here: https://linux.die.net/man/3/ssl_ctx_set_verify, `SSL_VERIFY_CLIENT_ONCE` and `SSL_VERIFY_POST_HANDSHAKE` are ignored in client mode (the driver is a client). `CASS_SSL_VERIFY_PEER_IDENTITY` and `CASS_SSL_VERIFY_PEER_IDENTITY_DNS` "seem to work" because they map to the values I mentioned in the previous sentence.

Ignoring the flags is not the proper behavior. The driver flags and the SSL flags do _not_ mean the same.

Gor027: I guess `SSL_VERIFY_PEER` also verifies the peer's hostname and IP address. I think we barely can simulate the behavior of the C++ driver in this case, because the handshake(where the certificate verification happens) should occur while trying to establish a session between the client and the database, and this is being done on the Rust driver side.

piodul: OK. In that case, let's print a warning to stderr in case one of the non-supported flags is provided and leave a TODO comment in the code. Furthermore, please explicitly map `CASS_XYZ` values to `SSL_XYZ` values in the code - instead of passing `CassSslVerifyFlags` value to the SSL function.

scylla-rust-wrapper/src/ssl.rs
```diff
@@ -0,0 +1,283 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_ssl_set_verify_flags(ssl: *mut CassSsl, flags: i32) {
+    let ssl = clone_arced(ssl);
+
+    match flags {
+        CASS_SSL_VERIFY_NONE => {
+            SSL_CTX_set_verify(ssl.ssl_context, SslVerifyMode::NONE.bits(), None)
+        }
+        CASS_SSL_VERIFY_PEER_CERT => {
```
piodul, line 140: The `CASS_SSL_VERIFY_` constants are flags, they are not mutually exclusive variants of an enum. The user can specify more than one of them by OR-ing them together.

Notice that their values are powers of two:

```c
typedef enum CassSslVerifyFlags_ {
  CASS_SSL_VERIFY_NONE              = 0x00,
  CASS_SSL_VERIFY_PEER_CERT         = 0x01,
  CASS_SSL_VERIFY_PEER_IDENTITY     = 0x02,
  CASS_SSL_VERIFY_PEER_IDENTITY_DNS = 0x04
} CassSslVerifyFlags;
```

Gor027: Currently, there is no support for flags other than `CASS_SSL_VERIFY_NONE` and `CASS_SSL_VERIFY_PEER_CERT`, so in my opinion OR-ing with `CASS_SSL_VERIFY_PEER_IDENTITY` or `CASS_SSL_VERIFY_PEER_IDENTITY_DNS` should not be accepted.

scylla-rust-wrapper/src/ssl.rs
```diff
@@ -0,0 +1,289 @@
...
+            SSL_CTX_set_verify(ssl.ssl_context, SslVerifyMode::PEER.bits(), None)
+        }
+        _ => {
+            // Check at position 1 the bit is set to 1
+            if (flags >> 1) & 1 == 1 {
```
piodul: Why use number literals when you can use constants?

```rust
if flags & CASS_SSL_VERIFY_PEER_IDENTITY_DNS != 0 {
    // ...
}
```

scylla-rust-wrapper/src/ssl.rs
piodul: Why not? If the user provides `CASS_SSL_VERIFY_PEER_CERT`, then we should enable the functionality provided by this flag. If they additionally provide `CASS_SSL_VERIFY_PEER_IDENTITY`, we should print a warning and ignore it, but I don't see why this should disable the first flag.

I agree that it would be the best to return an error when an unsupported flag is provided, but the function doesn't really allow us to do that.

piodul (commented):
A few small comments remaining, otherwise LGTM.

Gor027:
After [logging support](https://github.com/scylladb/cpp-rust-driver/pull/78) is merged into the master, the error checking/reporting functionality can be added to the SSL-related functions.

PR #71, opened 2022-08-16
Add batch support

Gor027:
## Pre-review checklist

This PR is dependent upon #69

Adds batch support to the driver.
As `BatchCounterThreeNodeClusterTests.Integration_Cassandra_MixedPreparedAndSimple` integration test creates a 3-node cluster which takes a lot of time, a special flag `--skip-wait-for-gossip-to-settle=0` is added to `ccm` to avoid unnecessary gossip delays.

### DIFF
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -1,0 +1,146 @@
+use crate::argconv::{free_boxed, ptr_to_ref, ptr_to_ref_mut};
+use crate::cass_error::CassError;
+use crate::cass_types::CassConsistency;
+use crate::cass_types::{make_batch_type, CassBatchType};
+use crate::statement::{CassStatement, Statement};
+use crate::types::*;
+use scylla::batch::Batch;
+use scylla::frame::response::result::CqlValue;
+use scylla::frame::value::MaybeUnset;
+use std::convert::TryInto;
+use std::sync::Arc;
+
+pub struct CassBatch {
+    pub state: Arc<CassBatchState>,
+    pub batch_request_timeout_ms: Option<cass_uint64_t>,
+}
+
+#[derive(Clone)]
+pub struct CassBatchState {
+    pub batch: Batch,
+    pub bound_values: Vec<Vec<MaybeUnset<Option<CqlValue>>>>,
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_batch_new(type_: CassBatchType) -> *mut CassBatch {
+    if let Some(batch_type) = make_batch_type(type_) {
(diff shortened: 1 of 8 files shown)


scylla-rust-wrapper/src/batch.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::types::*;
+use scylla::batch::Batch;
+use scylla::frame::response::result::CqlValue;
+use scylla::frame::value::MaybeUnset;
+use std::convert::TryInto;
+use std::sync::Arc;
+
+pub struct CassBatch {
+    pub state: Arc<CassBatchState>,
+    pub batch_request_timeout_ms: cass_uint64_t,
```
constantfold: This should be done the same way as in: https://github.com/scylladb/cpp-rust-driver/commit/fe2dadbd6cc2edbe6a8ce232c0a2ffa1943ea4ec#diff-20666f56cf01ee22aba41d78226242e6e1689fbe1ecaf7e82b8c3e3f532d8775R30

scylla-rust-wrapper/src/session.rs
```diff
@@ -59,6 +61,59 @@ pub unsafe extern "C" fn cass_session_connect(
...
+
+    match request_timeout_ms {
+        Some(timeout_ms) => {
+            CassFuture::make_raw(async move { request_with_timout(timeout_ms, future).await })
+        }
+        None => CassFuture::make_raw(future),
+    }
+}
+
+async fn request_with_timout(
```
piodul: Typo: `timout` -> `timeout`

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -531,3 +534,49 @@ pub unsafe extern "C" fn cass_data_type_add_sub_value_type_by_name_n(
...
+            _ => Err(()),
+        }
+    }
+}
+
+pub type CassBatchType = std::os::raw::c_uint;
+
+pub const CASS_BATCH_TYPE_LOGGED: CassBatchType = 0;
+pub const CASS_BATCH_TYPE_UNLOGGED: CassBatchType = 1;
+pub const CASS_BATCH_TYPE_COUNTER: CassBatchType = 2;
```
piodul: We have some magic in the `build.rs` file which parses `cassandra.h` and generates some of the enums for us, e.g. `CassConsistency`. Please turn it on for `CassBatchType` instead of defining the enum manually here.

piodul (commented):
LGTM after fixing the issue with CassBatchType being defined manually instead of generated from cassandra.h

piodul (approved):

constantfold (approved):
LGTM, merging

PR #72, opened 2022-08-17
Add cluster schema metadata and UDT iterator

Gor027:
## Pre-review checklist

This PR is dependent upon #69

This PR adds a possibility to fetch cluster schema metadata and adds an implementation of `UDT` iterator and functions to get `UDT` field names and values.
The cluster metadata is refreshed automatically in the Rust driver immediately after successfully reaching schema agreement.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,6 +44,5 @@ jobs:
 :SslNoClusterTests*:SslNoSslOnClusterTests*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :*5.Integration_Cassandra_*\
-:*19.Integration_Cassandra_*\
-:CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_UDT"
+:*19.Integration_Cassandra_*"
       run: valgrind --error-exitcode=123 ./cassandra-integration-tests --scylla --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,7 @@ jobs:
 :BatchSingleNodeClusterTests*:BatchCounterSingleNodeClusterTests*:BatchCounterThreeNodeClusterTests*\
 :ErrorTests.*\
 :SslNoClusterTests*:SslNoSslOnClusterTests*\
+:SchemaMetadataTest.*KeyspaceMetadata:SchemaMetadataTest.*MetadataIterator\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :*5.Integration_Cassandra_*\
 :*19.Integration_Cassandra_*"
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -3,6 +3,8 @@ use crate::cass_error::CassError;
 use crate::types::*;
 use scylla::batch::{BatchType, Consistency, SerialConsistency};
 use scylla::frame::response::result::ColumnType;
+use scylla::transport::topology::{CollectionType, CqlType, NativeType};
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::os::raw::c_char;
 use std::ptr;
@@ -12,7 +14,7 @@ include!(concat!(env!("OUT_DIR"), "/cppdriver_data_types.rs"));
 include!(concat!(env!("OUT_DIR"), "/cppdriver_data_query_error.rs"));
 include!(concat!(env!("OUT_DIR"), "/cppdriver_batch_types.rs"));
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct UDTDataType {
     // Vec to preserve the order of types
     pub field_types: Vec<(String, CassDataTypeArc)>,
@@ -27,6 +29,32 @@ impl UDTDataType {
             field_types: Vec::new(),
             keyspace: "".to_string(),
             name: "".to_string(),
+        }
+    }
+
+    pub fn create_with_params(
+        user_defined_types: &HashMap<String, Vec<(String, CqlType)>>,
+        keyspace_name: &str,
+        name: &str,
+    ) -> UDTDataType {
+        UDTDataType {
+            field_types: user_defined_types
+                .get(name)
+                .unwrap_or(&vec![])
+                .iter()
+                .map(|(udt_field_name, udt_field_type)| {
+                    (
+                        udt_field_name.clone(),
+                        Arc::new(get_column_type_from_cql_type(
+                            udt_field_type,
+                            user_defined_types,
+                            keyspace_name,
+                        )),
+                    )
+                })
+                .collect(),
+            keyspace: keyspace_name.to_string(),
+            name: name.to_owned(),
         }
     }
 
@@ -60,7 +88,7 @@ impl Default for UDTDataType {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub enum CassDataType {
     Value(CassValueType),
     UDT(UDTDataType),
@@ -72,6 +100,78 @@ pub enum CassDataType {
 }
 
 pub type CassDataTypeArc = Arc<CassDataType>;
+
+impl From<NativeType> for CassValueType {
+    fn from(native_type: NativeType) -> CassValueType {
+        match native_type {
+            NativeType::Ascii => CassValueType::CASS_VALUE_TYPE_ASCII,
+            NativeType::Boolean => CassValueType::CASS_VALUE_TYPE_BOOLEAN,
+            NativeType::Blob => CassValueType::CASS_VALUE_TYPE_BLOB,
+            NativeType::Counter => CassValueType::CASS_VALUE_TYPE_COUNTER,
+            NativeType::Date => CassValueType::CASS_VALUE_TYPE_DATE,
+            NativeType::Decimal => CassValueType::CASS_VALUE_TYPE_DECIMAL,
+            NativeType::Double => CassValueType::CASS_VALUE_TYPE_DOUBLE,
+            NativeType::Duration => CassValueType::CASS_VALUE_TYPE_DURATION,
+            NativeType::Float => CassValueType::CASS_VALUE_TYPE_FLOAT,
+            NativeType::Int => CassValueType::CASS_VALUE_TYPE_INT,
+            NativeType::BigInt => CassValueType::CASS_VALUE_TYPE_BIGINT,
+            NativeType::Text => CassValueType::CASS_VALUE_TYPE_TEXT,
+            NativeType::Timestamp => CassValueType::CASS_VALUE_TYPE_TIMESTAMP,
+            NativeType::Inet => CassValueType::CASS_VALUE_TYPE_INET,
+            NativeType::SmallInt => CassValueType::CASS_VALUE_TYPE_SMALL_INT,
+            NativeType::TinyInt => CassValueType::CASS_VALUE_TYPE_TINY_INT,
+            NativeType::Time => CassValueType::CASS_VALUE_TYPE_TIME,
+            NativeType::Timeuuid => CassValueType::CASS_VALUE_TYPE_TIMEUUID,
+            NativeType::Uuid => CassValueType::CASS_VALUE_TYPE_UUID,
+            NativeType::Varint => CassValueType::CASS_VALUE_TYPE_VARINT,
+        }
+    }
+}
+
+pub fn get_column_type_from_cql_type(
+    cql_type: &CqlType,
+    user_defined_types: &HashMap<String, Vec<(String, CqlType)>>,
+    keyspace_name: &str,
+) -> CassDataType {
+    match cql_type {
+        CqlType::Native(native) => CassDataType::Value(native.clone().into()),
+        CqlType::Collection { type_, .. } => match type_ {
+            CollectionType::List(list) => CassDataType::List(Some(Arc::new(
+                get_column_type_from_cql_type(list, user_defined_types, keyspace_name),
+            ))),
+            CollectionType::Map(key, value) => CassDataType::Map(
+                Some(Arc::new(get_column_type_from_cql_type(
+                    key,
+                    user_defined_types,
+                    keyspace_name,
+                ))),
+                Some(Arc::new(get_column_type_from_cql_type(
+                    value,
+                    user_defined_types,
+                    keyspace_name,
+                ))),
+            ),
+            CollectionType::Set(set) => CassDataType::Set(Some(Arc::new(
+                get_column_type_from_cql_type(set, user_defined_types, keyspace_name),
+            ))),
+        },
+        CqlType::Tuple(tuple) => CassDataType::Tuple(
+            tuple
+                .iter()
+                .map(|field_type| {
+                    Arc::new(get_column_type_from_cql_type(
+                        field_type,
+                        user_defined_types,
+                        keyspace_name,
+                    ))
+                })
+                .collect(),
+        ),
+        CqlType::UserDefinedType { name, .. } => CassDataType::UDT(
+            UDTDataType::create_with_params(user_defined_types, keyspace_name, name),
+        ),
+    }
+}
 
 impl CassDataType {
     fn get_sub_data_type(&self, index: usize) -> Option<&CassDataTypeArc> {
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -80,11 +80,18 @@ pub struct CassMapIterator {
     position: Option<usize>,
 }
 
+pub struct CassUdtIterator {
+    value: CassValue_,
+    count: u64,
+    position: Option<usize>,
+}
+
 pub enum CassIterator {
     CassResultIterator(CassResultIterator),
     CassRowIterator(CassRowIterator),
     CassCollectionIterator(CassCollectionIterator),
     CassMapIterator(CassMapIterator),
+    CassUdtIterator(CassUdtIterator),
 }
 
 #[no_mangle]
@@ -130,6 +137,13 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
             map_iterator.position = Some(new_pos);
 
             (new_pos < map_iterator.count.try_into().unwrap()) as cass_bool_t
+        }
+        CassIterator::CassUdtIterator(udt_iterator) => {
+            let new_pos: usize = udt_iterator.position.map_or(0, |prev_pos| prev_pos + 1);
+
+            udt_iterator.position = Some(new_pos);
+
+            (new_pos < udt_iterator.count.try_into().unwrap()) as cass_bool_t
         }
     }
 }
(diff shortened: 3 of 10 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -26,6 +27,30 @@ impl UDTDataType {
         }
     }
 
+    pub fn create_with_params(
+        udt_fields: &Vec<(String, CqlType)>,
```
piodul: Nit: unless you have a good reason to do so, always prefer `&[T]` over `&Vec<T>` in arguments. The former is able to capture a wider range of values, e.g. references to arrays, slices and references to vectors, while the last one is only able to accept a reference to vector.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -26,6 +27,30 @@ impl UDTDataType {
...
+        keyspace_name: &String,
```
piodul: Similarly, here. `&str` is better than `&String`. You can convert `&String` to `&str`, but not the other way around.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -69,6 +94,86 @@ pub enum CassDataType {
...
+            NativeType::Timeuuid => CassValueType::CASS_VALUE_TYPE_TIMEUUID,
+            NativeType::Uuid => CassValueType::CASS_VALUE_TYPE_UUID,
+            NativeType::Varint => CassValueType::CASS_VALUE_TYPE_VARINT,
+        }
+    }
+}
+
+pub fn get_column_type_from_cql_type(
+    cql_type: &CqlType,
+    udt_fields: &Vec<(String, CqlType)>,
```
piodul: Nit: `&[(String, CqlType)]`

scylla-rust-wrapper/src/session.rs
```diff
@@ -345,3 +359,144 @@ pub unsafe extern "C" fn cass_session_close(session: *mut CassSession) -> *const
...
+        }
+
+        let refresh_output = session_guard.as_ref().unwrap().refresh_metadata().await;
+
+        match refresh_output {
+            Ok(_) => Ok(CassResultValue::Empty),
+            Err(error) => Ok(CassResultValue::QueryError(Arc::new(error))),
+        }
+    });
+    ptr_to_ref(refresh_future).with_waited_result(|_| ());
```
piodul: `cass_session_get_schema_meta` shouldn't block. Long-running async operations should be returned in futures. The original implementation doesn't seem to block, either.

Please investigate why refreshing metadata in original implementation's `cass_session_get_schema_meta` isn't necessary. I suspect that the cpp driver refreshes metadata (either partially or fully) after receiving a Schema_change response to a query and then performs the refresh. If so, then it should be implemented in the rust driver, too - please create an issue there and work on it.

piodul (changes requested):

scylla-rust-wrapper/src/session.rs
Gor027: The C++ driver partially updates schema metadata on each schema change response. Currently, the Rust driver does not update the metadata on schema change responses, so I added support in [this PR](https://github.com/scylladb/scylla-rust-driver/pull/545) to fully refresh metadata if there is a schema change.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -26,6 +27,30 @@ impl UDTDataType {
...
+    ) -> UDTDataType {
+        UDTDataType {
+            field_types: udt_fields
+                .iter()
+                .map(|(udt_field_name, udt_field_type)| {
+                    (
+                        udt_field_name.clone(),
+                        Arc::new(get_column_type_from_cql_type(
+                            udt_field_type,
+                            udt_fields,
```
piodul: Why are you passing `udt_fields` here? Shouldn't the `udt_fields` be only used to create the main `UDTDataType`? I don't see why they are needed to create the field types.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -69,6 +94,86 @@ pub enum CassDataType {
...
+                        keyspace_name,
+                    ))
+                })
+                .collect(),
+        ),
+        CqlType::UserDefinedType {
+            frozen: _frozen,
+            name,
+        } => CassDataType::UDT(UDTDataType::create_with_params(
+            udt_fields,
```
piodul: This is wrong. You can have nested UDTs (a UDT can have a field which is also a UDT but of different type). Currently, in `get_column_type_from_cql_type`, you are always passing the same `udt_fields` around, so if you have a nested UDT then it will have wrong fields. Moreover, you will probably cause infinite recursion (`UDTDataType::create_with_params` calls `get_column_type_from_cql_type` and vice versa).

.github/workflows/build.yml
```diff
@@ -41,6 +41,5 @@ jobs:
 :CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :*5.Integration_Cassandra_*\
-:*19.Integration_Cassandra_*\
-:CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_UDT"
+:*19.Integration_Cassandra_*"
```
piodul, line 48: Typo in the "Add user defined type iterator" commit: `Enableed` -> `Enabled`

scylla-rust-wrapper/src/session.rs
```diff
@@ -24,6 +27,18 @@ use tokio::sync::RwLock;
 pub type CassSession = RwLock<Option<Session>>;
 type CassSession_ = Arc<CassSession>;
 
+pub struct CassKeyspaceMeta {
+    name: String,
+    // User defined type name to fields(field_name, field_type)
+    pub user_defined_types: HashMap<String, Vec<(String, CassDataType)>>,
+    // User defined type name to type
+    pub user_defined_type_data_type: HashMap<String, CassDataType>,
```
piodul: Why do we need two hashmaps? Isn't information about fields kept in `user_defined_type_data_type`?

scylla-rust-wrapper/src/session.rs
```diff
@@ -405,12 +425,53 @@ pub unsafe extern "C" fn cass_session_get_schema_meta(
                         )),
                     );
                 });
+            keyspace
```
piodul: Please change `for_each` into regular `for` loops. The usage of `for_each` introduces more indentation than necessary (2 tabs instead of 1)

piodul (changes requested):

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -69,6 +97,86 @@ pub enum CassDataType {
...
+
+pub fn get_column_type_from_cql_type(
+    cql_type: &CqlType,
+    user_defined_types: &HashMap<String, Vec<(String, CqlType)>>,
+    keyspace_name: &str,
+) -> CassDataType {
+    match cql_type {
+        CqlType::Native(native) => CassDataType::Value(native.clone().into()),
+        CqlType::Collection {
+            frozen: _frozen,
```
piodul: Nit: if you are matching on a variant with named fields, you can skip them with dots:

```rust
match cql_type {
    CqlType::Collection { type_, .. } => {
        // ...
    }
}

File: scylla-rust-wrapper/src/session.rs (hunk 4 above)
Comment (piodul) [reply scylla-rust-wrapper/src/session.rs]: Is this still necessary? I think that your PR to the rust driver was merged. The refresh here probably can be removed.

File: scylla-rust-wrapper/src/session.rs [hunk 11]
```diff
@@ -345,3 +357,136 @@ pub unsafe extern "C" fn cass_session_close(session: *mut CassSession) -> *const
...
+    type_length: size_t,
+) -> *const CassDataType {
+    if type_.is_null() {
+        return std::ptr::null();
+    }
+
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    let user_type_name = ptr_to_cstr_n(type_, type_length).unwrap();
+
+    return match keyspace_meta
```
piodul: `return` at the end of the function - not necessary.

tests/src/integration/tests/test_cassandra_types.cpp
```diff
@@ -824,6 +824,7 @@ CASSANDRA_INTEGRATION_TYPED_TEST_P(CassandraTypesTests, UDT) {
     field << "field" << i;
     udt_values[field.str()] = values[i];
   }
+  this->msleep(1000);
```
piodul: What is this sleep for? We shouldn't depend on timing in tests.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -145,6 +153,15 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
```
piodul, line 171: There is a lot of repeated code in the `cass_iterator_next` function. It looks like it could use some refactoring (in a different PR, though).

tests/src/integration/tests/test_cassandra_types.cpp
```diff
@@ -857,8 +858,27 @@ CASSANDRA_INTEGRATION_TYPED_TEST_P(CassandraTypesTests, UDT) {
...
+        cass_iterator_get_user_type_field_name(collection_iterator, &field_name, &field_name_length);
+        TypeParam field_value = TypeParam(cass_iterator_get_user_type_field_value(collection_iterator));
+        std::string key(field_name, field_name_length);
+        ASSERT_EQ(field_value, udt_values[key]);
+      } else {
+        throw Exception("No more values available");
+      }
+    }
+
+    cass_iterator_free(collection_iterator);
```
piodul, line 880: What is the reason for this change? Is it just to test the UDT iterator by reusing an old test? This removes some previous assertions, I'd rather not have them removed without a good reason.

If this is only to test the UDT iterator, then aren't there existing tests for it?

scylla-rust-wrapper/src/session.rs
```diff
@@ -434,11 +456,51 @@ pub unsafe extern "C" fn cass_session_get_schema_meta(
...
+                                CassColumnType::CASS_COLUMN_TYPE_CLUSTERING_KEY
+                            }
+                            ColumnKind::PartitionKey => {
+                                CassColumnType::CASS_COLUMN_TYPE_PARTITION_KEY
+                            }
+                        },
+                    };
+
+                    columns_metadata.insert(column_name.clone(), cass_column_meta);
+                });
```
piodul: Tip: you can collect into a hashmap from an iterator of pairs:

```rust
let columns_metadata: HashMap<_, _> = table_metadata.columns
    .iter()
    .map(|(column_name, column_metadata)| {
        // ...
        (column_name.clone(), cass_column_meta)
    })
    .collect();
```

tests/src/integration/tests/test_cassandra_types.cpp
Gor027: Unfortunately, this is the only test for UDT iterator, earlier we also modified tests [here](https://github.com/scylladb/cpp-rust-driver/pull/69) for other collection types(map, set, list, tuple) as those tests did not use API functions to check the size or the elements of the collections, instead, the `CassValue` was cast to a C++ object.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -395,6 +428,87 @@ pub unsafe extern "C" fn cass_iterator_get_keyspace_meta(
     std::ptr::null()
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_iterator_get_table_meta(
+    iterator: *const CassIterator,
+) -> *const CassTableMeta {
+    let iter = ptr_to_ref(iterator);
+
+    if let CassIterator::CassKeyspaceMetaIterator(keyspace_meta_iterator) = iter {
```
piodul: It looks like `CassKeyspaceMetaIterator` is used for two purposes: to iterate over tables and user types. I think those should be two separate variants so that it's harder to misuse the iterator.

piodul (commented):
I can't see any existing tests for the iterators added in this PR. Please make sure that all newly added functionality is covered either by the existing, or by new tests.

tests/src/integration/tests/test_schema_metadata.cpp
```diff
@@ -194,3 +194,157 @@ CASSANDRA_INTEGRATION_TEST_F(SchemaMetadataTest, VirtualMetadata) {
...
+  EXPECT_EQ(cass_data_type_type(cass_column_meta_data_type(column_meta)), CASS_VALUE_TYPE_TEXT);
+
+  column_meta = cass_table_meta_column_by_name(table_meta, "value");
+  ASSERT_TRUE(column_meta);
+  EXPECT_EQ(cass_data_type_type(cass_column_meta_data_type(column_meta)), CASS_VALUE_TYPE_BIGINT);
+
+  cass_schema_meta_free(schema_meta);
+}
+
+CASSANDRA_INTEGRATION_TEST_F(SchemaMetadataTest, MetadataIterator) {
```
piodul, line 257: It would be nice if the test also checked that the iterators do not return more items than expected.

tests/src/integration/tests/test_schema_metadata.cpp
```diff
@@ -27,8 +27,8 @@ class SchemaMetadataTest : public Integration {
   void SetUp() {
     CHECK_VERSION(2.2.0);
     Integration::SetUp();
-    populateSchema();
-    schema_meta_ = session_.schema();
+//    populateSchema();
+//    schema_meta_ = session_.schema();
```
piodul: Why doesn't it break the existing test cases in the file? It looks like other tests here require `schema_meta_` to be initialized after they start, they don't use the `schema_meta()` method.

tests/src/integration/tests/test_schema_metadata.cpp
```diff
@@ -194,3 +194,157 @@ CASSANDRA_INTEGRATION_TEST_F(SchemaMetadataTest, VirtualMetadata) {
...
+  std::string table_name = "table_meta";
+  std::string user_type = "address";
+
+  session_.execute(format_string("CREATE KEYSPACE IF NOT EXISTS %s WITH "
+                                 "replication = { 'class': 'SimpleStrategy', 'replication_factor': '1'}", ks_name.c_str()));
+
+  session_.execute(format_string("CREATE TABLE IF NOT EXISTS %s.%s (key text, value bigint, "
+                                 "PRIMARY KEY (key))", ks_name.c_str(), table_name.c_str()));
+
+  session_.execute(format_string("CREATE TYPE IF NOT EXISTS %s.%s (street text, city text)", ks_name.c_str(), user_type.c_str()));
```
piodul: How about using the keyspace/table/UDT already defined in `populateSchema()`? You won't have to introduce the `schema_metadata()` method.

tests/src/integration/tests/test_schema_metadata.cpp
Gor027: The existing tests are disabled for the C++ driver, and the implementation of the tests is not based on the driver's API functions, so I thought it is better to write new tests with separate keyspace/table, in case the existing tests may be fixed in the future.

tests/src/integration/tests/test_schema_metadata.cpp
Gor027: The `schema_meta()` method is introduced to get a raw pointer `const CassSchemaMeta*` and pass it as an argument to functions that return fields of schema metadata.

tests/src/integration/tests/test_schema_metadata.cpp
piodul: I don't understand. The other tests in the `test_schema_metadata.cpp` file do seem to use the driver API - mostly related to materialized views. Why would the other tests need "fixing"? I know that we don't support materialized views yet (you have sent another PR to fix it), but I don't see why they wouldn't work after the appropriate functions are implemented.

tests/src/integration/tests/test_schema_metadata.cpp
piodul: The `(SchemaMetadataTest, DropView)` test drops a materialized view and then calls `session_.schema()` to verify that the view disappeared from the metadata. Can we do the same here?

Gor027:
The build failed because of `cargo clippy` warnings, perhaps because of some updates in cargo clippy. [This](https://github.com/scylladb/cpp-rust-driver/pull/82) PR should fix the issue.

tests/src/integration/tests/test_schema_metadata.cpp
Gor027: The problem was that the `populateSchema()` creates UDF, aggregate and index. Support for UDF should be manually enabled and `scylla-ccm` does not provide a way to do it. So, for now, I just commented out the creation of the user-defined function, aggregate and index and used the existing keyspace name, table name and user defined type in newly added tests for metadata.

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -73,13 +73,13 @@ impl From<&CassUserType> for CqlValue {
...
-    let data_type = clone_arced(data_type_raw);
+    let data_type = ptr_to_ref(data_type_raw);
 
-    match &*data_type {
+    match data_type {
         CassDataType::UDT(udt_data_type) => {
             let field_values = vec![None; udt_data_type.field_types.len()];
             Box::into_raw(Box::new(CassUserType {
-                data_type,
+                data_type: Arc::new(CassDataType::UDT(udt_data_type.clone())),
```
constantfold: This change isn't described in the commit message.

Gor027: This was not an intentional change, thanks for noticing it.

tests/src/integration/tests/test_schema_metadata.cpp
```diff
@@ -36,6 +36,11 @@ class SchemaMetadataTest : public Integration {
                                    "PRIMARY KEY (key))",
                                    table_name_.c_str()));
 
+/*
+ * Support for UDF should be manually enabled and currently, ccm does not provide any way to do it.
```
constantfold: I don't think that's true - you can configure ccm to start Scylla/Cassandra with UDFs enabled. With Cassandra you can do it with `ccm updateconf enable_user_defined_functions:true` and with Scylla I belive you can do it with `ccm updateconf experimental_features:udf`.

scylla-rust-wrapper/src/user_type.rs
Gor027: Actually, this was a necessary change, however, it was in a wrong commit. The function should create and return a new `CassUserType` object which must be freed separately without freeing/modifying the passed `data_type` argument of type `CassDataType`. So, instead of just passing `data_type`, we should create a new `CassDataTypeArc` object with cloning the value of `data_type`.

Lorak-mmk: I don't follow your reasoning, could you explain a bit more? I think the whole point of reference counting here was to achieve what you said.
Also, the original driver doesn't perform a copy - it seems to only increase refcount, as were we before this change.

tests/src/integration/tests/test_schema_metadata.cpp
```diff
@@ -69,8 +75,8 @@ class SchemaMetadataTest : public Integration {
     }
     session_.execute("CREATE TYPE address (street text, city text)");
 
-    session_.execute(
-        format_string("CREATE INDEX schema_meta_index ON %s (value)", table_name_.c_str()));
+//    session_.execute(
+//        format_string("CREATE INDEX schema_meta_index ON %s (value)", table_name_.c_str()));
```
Lorak-mmk: Why is this commented out? If there is a reason, it needs a comment in the code.

Gor027: It is commented out as the created index is not being checked in these tests.

scylla-rust-wrapper/src/user_type.rs
Gor027: I guess the C++ driver creates another `SharedRefPtr` with the `CassDataType` pointer, too, instead of just incrementing the ref count.

tests/src/integration/tests/test_schema_metadata.cpp
Lorak-mmk: Please comment it in the code

scylla-rust-wrapper/src/user_type.rs
Lorak-mmk: Can we do the same?

Gor027: I think the current implementation does that. It clones the `udt_data_type` and creates a new Arc'ed `CassDataType` object.

Lorak-mmk (approved):

PR #73, opened 2022-08-19
Allow bigger difference between timestamps in BasicsTests.Timestamps

Gor027:
Increased the acceptable time difference between sequential inserts in
`BasicsTests.Timestamps` test, cause it happens to fail on very slow
machines like Github actions.

## Pre-review checklist

Increased the acceptable time difference between sequential inserts in
`BasicsTests.Timestamps` test, cause it happens to fail on very slow
machines like Github actions.


constantfold:
I will re-run it a couple of times before merging.

constantfold:
As evidenced by https://github.com/scylladb/cpp-rust-driver/actions/runs/3129242901/jobs/5078086621, `150000` was still not enough. For the time being, I'll restart the integration test - this is not a priority thing to fix right now.

PR #74, opened 2022-09-02
Add workflow to Github actions to run tests on Cassandra

Gor027:
## Pre-review checklist

This PR adds a workflow to Github actions to run integration tests on a Cassandra cluster. There are some tests, e.g. SSL-related tests, that cannot be run on a Scylla cluster because of `scylla-ccm` but they can be tested on a Cassandra cluster. So, new tests that will be enabled in the Github actions, should also be added in `cassandra.yml`, unless the newly added features are not compatible with Cassandra.


piodul (approved):
LGTM. Please remember to update test filters in your other PRs after this is merged.

PR #75, opened 2022-09-07
Add getting description for CassError

Gor027:
Added implementation for `cass_error_desc` function to return description for a certain error.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/external.rs
+++ b/scylla-rust-wrapper/src/external.rs
@@ -1,9 +1,79 @@
+use crate::argconv::ptr_to_ref;
 use crate::cass_error::*;
 use std::os::raw::c_char;
 
 #[no_mangle]
-pub unsafe extern "C" fn cass_error_desc(_error: *const CassError) -> *const c_char {
-    // FIXME: add proper implementation
-    let error = "my_custom_error\0";
-    error.as_ptr() as *const c_char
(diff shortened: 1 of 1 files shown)


scylla-rust-wrapper/src/external.rs
```diff
@@ -1,9 +1,129 @@
...
 use std::os::raw::c_char;
 
 #[no_mangle]
-pub unsafe extern "C" fn cass_error_desc(_error: *const CassError) -> *const c_char {
-    // FIXME: add proper implementation
-    let error = "my_custom_error\0";
-    error.as_ptr() as *const c_char
+pub unsafe extern "C" fn cass_error_desc(error: *const CassError) -> *const c_char {
+    let error: &CassError = ptr_to_ref(error);
+    match *error {
```
piodul: You can try moving out the `.as_ptr() as *const c_char` part of the code outside the match. The lines will become shorter and `cargo fmt` will perhaps not require you to do as many line breaks as it currently does.

I have something like this in mind:

```rust
let s: &'static str = match *error {
    CassError::CASS_ERROR_LIB_BAD_PARAMS => "Bad parameters",
    CassError::CASS_ERROR_LIB_NO_STREAMS => "No streams available",
    // ...
};
s.as_ptr() as *const c_char
```

piodul (approved):

PR #76, opened 2022-09-09
Materialized view

Gor027:
## Pre-review checklist

This PR is dependent upon #72 

Adds support to fetch materialized views from cluster metadata and implementation for functions to return fields of a certain view in the metadata. Additionally, it adds support to iterate over materialized views.

### DIFF
--- a/scylla-rust-wrapper/src/metadata.rs
+++ b/scylla-rust-wrapper/src/metadata.rs
@@ -1,0 +1,310 @@
+use crate::argconv::*;
+use crate::cass_types::get_column_type_from_cql_type;
+use crate::cass_types::CassDataType;
+use crate::types::*;
+use scylla::transport::topology::{ColumnKind, CqlType, Table};
+use std::collections::HashMap;
+use std::os::raw::c_char;
+use std::sync::Arc;
+use std::sync::Weak;
+
+include!(concat!(env!("OUT_DIR"), "/cppdriver_column_type.rs"));
+
+pub type CassSchemaMeta_ = &'static CassSchemaMeta;
+
+pub struct CassSchemaMeta {
+    pub keyspaces: HashMap<String, CassKeyspaceMeta>,
+}
+
+pub type CassKeyspaceMeta_ = &'static CassKeyspaceMeta;
+
+pub struct CassKeyspaceMeta {
+    pub name: String,
+
+    // User defined type name to type
+    pub user_defined_type_data_type: HashMap<String, Arc<CassDataType>>,
+    pub tables: HashMap<String, Arc<CassTableMeta>>,
+    pub views: HashMap<String, Arc<CassMaterializedViewMeta>>,
+}
+
+pub type CassTableMeta_ = &'static CassTableMeta;
+
+pub struct CassTableMeta {
+    pub name: String,
+    pub columns_metadata: HashMap<String, CassColumnMeta>,
+    pub partition_keys: Vec<String>,
+    pub clustering_keys: Vec<String>,
+    pub views: HashMap<String, Arc<CassMaterializedViewMeta>>,
+}
+
+pub type CassMaterializedViewMeta_ = &'static CassMaterializedViewMeta;
+
+pub struct CassMaterializedViewMeta {
+    pub name: String,
+    pub view_metadata: CassTableMeta,
+    pub base_table: Weak<CassTableMeta>,
+}
+
+pub struct CassColumnMeta {
+    pub name: String,
+    pub column_type: CassDataType,
+    pub column_kind: CassColumnType,
+}
+
+pub unsafe fn create_table_metadata(
+    keyspace_name: &str,
+    table_name: &str,
+    table_metadata: &Table,
+    user_defined_types: &HashMap<String, Vec<(String, CqlType)>>,
+) -> CassTableMeta {
+    let mut columns_metadata = HashMap::new();
+    table_metadata
+        .columns
+        .iter()
+        .for_each(|(column_name, column_metadata)| {
+            let cass_column_meta = CassColumnMeta {
+                name: column_name.clone(),
+                column_type: get_column_type_from_cql_type(
+                    &column_metadata.type_,
+                    user_defined_types,
+                    keyspace_name,
+                ),
+                column_kind: match column_metadata.kind {
+                    ColumnKind::Regular => CassColumnType::CASS_COLUMN_TYPE_REGULAR,
+                    ColumnKind::Static => CassColumnType::CASS_COLUMN_TYPE_STATIC,
+                    ColumnKind::Clustering => CassColumnType::CASS_COLUMN_TYPE_CLUSTERING_KEY,
+                    ColumnKind::PartitionKey => CassColumnType::CASS_COLUMN_TYPE_PARTITION_KEY,
+                },
+            };
+
+            columns_metadata.insert(column_name.clone(), cass_column_meta);
+        });
+
+    CassTableMeta {
+        name: table_name.to_owned(),
+        columns_metadata,
+        partition_keys: table_metadata.partition_key.clone(),
+        clustering_keys: table_metadata.clustering_key.clone(),
+        views: HashMap::new(),
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_schema_meta_free(schema_meta: *mut CassSchemaMeta) {
+    free_boxed(schema_meta)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_schema_meta_keyspace_by_name(
+    schema_meta: *const CassSchemaMeta,
+    keyspace_name: *const c_char,
+) -> *const CassKeyspaceMeta {
+    cass_schema_meta_keyspace_by_name_n(schema_meta, keyspace_name, strlen(keyspace_name))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_schema_meta_keyspace_by_name_n(
+    schema_meta: *const CassSchemaMeta,
+    keyspace_name: *const c_char,
+    keyspace_name_length: size_t,
+) -> *const CassKeyspaceMeta {
+    if keyspace_name.is_null() {
+        return std::ptr::null();
+    }
+
+    let metadata = ptr_to_ref(schema_meta);
+    let keyspace = ptr_to_cstr_n(keyspace_name, keyspace_name_length).unwrap();
+
+    let keyspace_meta = metadata.keyspaces.get(keyspace);
+
+    match keyspace_meta {
+        Some(meta) => meta,
+        None => std::ptr::null(),
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_keyspace_meta_name(
+    keyspace_meta: *const CassKeyspaceMeta,
+    name: *mut *const c_char,
+    name_length: *mut size_t,
+) {
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    write_str_to_c(keyspace_meta.name.as_str(), name, name_length)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_keyspace_meta_user_type_by_name(
+    keyspace_meta: *const CassKeyspaceMeta,
+    type_: *const c_char,
+) -> *const CassDataType {
+    cass_keyspace_meta_user_type_by_name_n(keyspace_meta, type_, strlen(type_))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_keyspace_meta_user_type_by_name_n(
+    keyspace_meta: *const CassKeyspaceMeta,
+    type_: *const c_char,
+    type_length: size_t,
+) -> *const CassDataType {
+    if type_.is_null() {
+        return std::ptr::null();
+    }
+
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    let user_type_name = ptr_to_cstr_n(type_, type_length).unwrap();
+
+    match keyspace_meta
+        .user_defined_type_data_type
+        .get(user_type_name)
+    {
+        Some(udt) => Arc::as_ptr(udt),
+        None => std::ptr::null(),
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_keyspace_meta_table_by_name(
+    keyspace_meta: *const CassKeyspaceMeta,
+    table: *const c_char,
+) -> *const CassTableMeta {
+    cass_keyspace_meta_table_by_name_n(keyspace_meta, table, strlen(table))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_keyspace_meta_table_by_name_n(
+    keyspace_meta: *const CassKeyspaceMeta,
+    table: *const c_char,
+    table_length: size_t,
+) -> *const CassTableMeta {
+    if table.is_null() {
+        return std::ptr::null();
+    }
+
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    let table_name = ptr_to_cstr_n(table, table_length).unwrap();
+
+    let table_meta = keyspace_meta.tables.get(table_name);
+
+    match table_meta {
+        Some(meta) => Arc::as_ptr(meta) as *const CassTableMeta,
+        None => std::ptr::null(),
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_table_meta_name(
+    table_meta: *const CassTableMeta,
+    name: *mut *const c_char,
+    name_length: *mut size_t,
+) {
+    let table_meta = ptr_to_ref(table_meta);
+    write_str_to_c(table_meta.name.as_str(), name, name_length)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_table_meta_column_count(table_meta: *const CassTableMeta) -> size_t {
+    let table_meta = ptr_to_ref(table_meta);
+    table_meta.columns_metadata.len() as size_t
+}
+
+#[no_mangle]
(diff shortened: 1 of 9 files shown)


Lorak-mmk:
Please rebase on current version of #72

Lorak-mmk:
Please rebase on current master - it's a bit difficult to review with those unrelated changes from #72

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -73,13 +73,13 @@ impl From<&CassUserType> for CqlValue {
...
-    let data_type = clone_arced(data_type_raw);
+    let data_type = ptr_to_ref(data_type_raw);
 
-    match &*data_type {
+    match data_type {
         CassDataType::UDT(udt_data_type) => {
             let field_values = vec![None; udt_data_type.field_types.len()];
             Box::into_raw(Box::new(CassUserType {
-                data_type,
+                data_type: Arc::new(CassDataType::UDT(udt_data_type.clone())),
```
Lorak-mmk: I think in UDT PR this change was reverted after we debugged the problem?

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -73,13 +73,13 @@ impl From<&CassUserType> for CqlValue {
...
                 field_values,
             }))
         }
```
Lorak-mmk: This commit does not do what the message says.

scylla-rust-wrapper/src/session.rs
```diff
@@ -49,6 +51,13 @@ pub struct CassTableMeta {
     pub columns_metadata: HashMap<String, CassColumnMeta>,
     pub partition_keys: Vec<String>,
     pub clustering_keys: Vec<String>,
+    pub views: Option<HashMap<String, Arc<CassMaterializedViewMeta>>>,
+}
+
+pub struct CassMaterializedViewMeta {
+    pub name: String,
+    pub view_metadata: CassTableMeta,
+    pub base_table: Arc<RefCell<CassTableMeta>>,
```
Lorak-mmk: CassTableMeta and CassMatarializedViewMeta have Arc's of each other  - isn't there going to be a memory leak? Should one of those Arcs be a Weak?

You probably want to use https://doc.rust-lang.org/std/sync/struct.Arc.html#method.new_cyclic to initialize those structs (after breaking ownership cycle using Weak) - that way the RefCell may become redundant.

scylla-rust-wrapper/src/session.rs
```diff
@@ -774,3 +774,204 @@ pub unsafe extern "C" fn cass_column_meta_type(
...
+) -> *const CassMaterializedViewMeta {
+    if view.is_null() {
+        return std::ptr::null();
+    }
+
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    let view_name = ptr_to_cstr_n(view, view_length).unwrap();
+
+    match keyspace_meta.views.get(view_name) {
+        Some(view_meta) => Arc::as_ptr(view_meta) as *const CassMaterializedViewMeta,
```
Lorak-mmk: Why is this pointer cast necessary? What's the type returned by `Arc::as_ptr` here?

scylla-rust-wrapper/src/session.rs
```diff
@@ -774,3 +774,204 @@ pub unsafe extern "C" fn cass_column_meta_type(
...
+    if view.is_null() {
+        return std::ptr::null();
+    }
+
+    let table_meta = ptr_to_ref(table_meta);
+    let view_name = ptr_to_cstr_n(view, view_length).unwrap();
+
+    match &table_meta.views {
+        Some(table_views) => match table_views.get(view_name) {
+            Some(view_meta) => Arc::as_ptr(view_meta) as *const CassMaterializedViewMeta,
```
Lorak-mmk: Again, why is this cast needed?

Lorak-mmk (changes requested):
I don't like that all the metadata methods are in session.rs - they should have their own file, as they are not methods on the CassSesson object.

scylla-rust-wrapper/src/user_type.rs
Gor027: Yes, this is an unnecessary change, the commit should be dropped.

scylla-rust-wrapper/src/user_type.rs
Gor027: These changes survived after a rebase as there were memory corruption issues with this part of the code. As we have debugged and solved the problem, the changes will be reverted.

scylla-rust-wrapper/src/session.rs
Gor027: The type returned by `Arc::as_ptr` should be `*const CassMaterializedVIewMeta`, I added the cast here to explicitly mark in the code what is returned from the function. I guess anyway the compiler does some optimization and removes the cast here.

scylla-rust-wrapper/src/session.rs
Gor027: It is not necessary and can be safely removed.

scylla-rust-wrapper/src/session.rs
Gor027: Reference to the base table of a materialized view should be Weak, you are right. I will add some tests to ensure we avoid memory leaks.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: I think the cast makes it less readable, can you remove it?

scylla-rust-wrapper/src/session.rs
```diff
@@ -464,40 +475,34 @@ pub unsafe extern "C" fn cass_session_get_schema_meta(
...
+                    view_name,
+                    &view_metadata.view_metadata,
+                    &keyspace.user_defined_types,
+                );
+                let mut cass_view_meta = CassMaterializedViewMeta {
+                    name: view_name.clone(),
+                    view_metadata: cass_view_table_meta,
+                    base_table: Weak::new(),
+                };
+                cass_view_meta.base_table = Arc::downgrade(&cass_table_meta);
```
Lorak-mmk: Why not init to `Arc::downgrade(&cass_table_meta);` in the first place? Then you could get rid of `mut`.

scylla-rust-wrapper/src/session.rs
```diff
@@ -49,6 +50,15 @@ pub struct CassTableMeta {
     pub columns_metadata: HashMap<String, CassColumnMeta>,
     pub partition_keys: Vec<String>,
     pub clustering_keys: Vec<String>,
+    pub views: std::sync::RwLock<Option<HashMap<String, Arc<CassMaterializedViewMeta>>>>,
```
Lorak-mmk: Why `Option`? I don't see any place where `None` is treated differently than `Some` with hashmap not containing relevant key.

Lorak-mmk: Is this RwLock only used as `write()` to initialize the structure? If so, the I think this is the wrong tool here. RefCell could be used without other changes. But this also should not be necessary - can we initialize the structure using https://doc.rust-lang.org/std/sync/struct.Arc.html#method.new_cyclic ? That would allow getting rid of a lot of `.read().unwrap()` from other places in this PR, making the code more readable. While applied along with previous code, the type of this field would be just `HashMap<String, Arc<CassMaterializedViewMeta>>`.

scylla-rust-wrapper/src/session.rs
```diff
@@ -39,7 +39,8 @@ pub struct CassKeyspaceMeta {
 
     // User defined type name to type
     pub user_defined_type_data_type: HashMap<String, Arc<CassDataType>>,
-    pub tables: HashMap<String, CassTableMeta>,
+    pub tables: HashMap<String, Arc<CassTableMeta>>,
+    pub views: HashMap<String, Arc<CassMaterializedViewMeta>>,
```
Lorak-mmk: This is not that important, but why metadata-related structures use `Arc` and not just `Rc`? cppdriver only requires `CassSession` and `CassFuture` to be thread safe. Do we do something similar (stronger thread-safety than cppdriver requires) in other places?
Here, using `Arc` instead of `Rc` allows to call `cass_schema_meta_free` on a different thread than `cass_session_get_schema_meta` was called from (those are the only places where the counters are modified) - that doesn't seem like a big benefit.

Lorak-mmk (changes requested):
I don't quite like how are the structs / functions places between `session.rs` and `metadata.rs`. Before this PR we had no `metadata.rs` file, so everything was put into `session.rs.` Now that we have it, we can fix this situation.

I'd propose the following:
- Leave `query_results.rs` as it is - meaning that functions accepting or returning `CassIterator` go there.
- Move metadata structs to `metadata.rs` - `CassKeyspaceMeta_`, `CassKeyspaceMeta`, `CassTableMeta_`, `CassTableMeta`, `CassMaterializedViewMeta_`, `CassMaterializedViewMeta`, `CassColumnMeta`, `CassSchemaMeta_`, `CassSchemaMeta`.
- Move metadata functions that don't touch session into `metadata.rs` - everything from `cass_schema_meta_free` downward.
- `create_table_metadata` should probably also go into `metadata.rs`, but I'm not 100% sure about it.

What do you think?

scylla-rust-wrapper/src/session.rs
Gor027: The initialization of `cass_table_meta` with `Arc::new_cyclic` allowed to avoid `Arc::downgrade` and to make `cass_view_meta` immutable.

scylla-rust-wrapper/src/session.rs
Gor027: The purpose of `Option` was to differentiate uninitialized `views` from empty `views` for better error handling, however, it seemed to be useless, so I removed `Option`.

Gor027: Initialization with `Arc::new_cyclic` allowed to get rid of `RwLock` and keep the `views` type as simple as `HashMap<String, Arc<CassMaterializedViewMeta>>`.

Gor027:
> I don't quite like how are the structs / functions places between `session.rs` and `metadata.rs`. Before this PR we had no `metadata.rs` file, so everything was put into `session.rs.` Now that we have it, we can fix this situation.
> 
> I'd propose the following:
> 
> * Leave `query_results.rs` as it is - meaning that functions accepting or returning `CassIterator` go there.
> * Move metadata structs to `metadata.rs` - `CassKeyspaceMeta_`, `CassKeyspaceMeta`, `CassTableMeta_`, `CassTableMeta`, `CassMaterializedViewMeta_`, `CassMaterializedViewMeta`, `CassColumnMeta`, `CassSchemaMeta_`, `CassSchemaMeta`.
> * Move metadata functions that don't touch session into `metadata.rs` - everything from `cass_schema_meta_free` downward.
> * `create_table_metadata` should probably also go into `metadata.rs`, but I'm not 100% sure about it.
> 
> What do you think?

As you suggested above, I moved all metadata-related structures and functions to `metadata.rs`. Importing the `CassColumnType` enum is also moved to `metadata.rs`, so I think it is better to have `create_table_metadata` also in that file rather than in `session.rs`.

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+    pub name: String,
+    pub column_type: CassDataType,
+    pub column_kind: CassColumnType,
+}
+
+pub type CassSchemaMeta_ = &'static CassSchemaMeta;
+
+pub struct CassSchemaMeta {
+    pub keyspaces: HashMap<String, CassKeyspaceMeta>,
+}
```
Lorak-mmk: Nit / subjective: Move them above `CassKeyspaceMeta_`

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    let keyspace_meta = ptr_to_ref(keyspace_meta);
+    let user_type_name = ptr_to_cstr_n(type_, type_length).unwrap();
+
+    match keyspace_meta
+        .user_defined_type_data_type
+        .get(user_type_name)
+    {
+        Some(udt) => Arc::into_raw(udt.clone()) as *const CassDataType,
```
Lorak-mmk: Why is there a clone? I see that cpp-driver doesn't clone this, so we'd be incompatible (in this version user would have to explicitly free this udt, in contrary to cpp-driver).

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+        return std::ptr::null();
+    }
+
+    let metadata = ptr_to_ref(schema_meta);
+    let keyspace = ptr_to_cstr_n(keyspace_name, keyspace_name_length).unwrap();
+
+    let keyspace_meta = metadata.keyspaces.get(keyspace);
+
+    match keyspace_meta {
+        Some(meta) => meta,
```
Lorak-mmk: `meta` is a reference, so maybe add an explicit cast here?

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+) -> *const CassMaterializedViewMeta {
+    if view.is_null() {
+        return std::ptr::null();
+    }
+
+    let table_meta = ptr_to_ref(table_meta);
+    let view_name = ptr_to_cstr_n(view, view_length).unwrap();
+
+    match table_meta.views.get(view_name) {
+        Some(view_meta) => Arc::as_ptr(view_meta) as *const CassMaterializedViewMeta,
```
Lorak-mmk: Remove the cast (as_ptr already returns correct type)

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_table_meta_materialized_view(
+    table_meta: *const CassTableMeta,
+    index: size_t,
+) -> *const CassMaterializedViewMeta {
+    let table_meta = ptr_to_ref(table_meta);
+
+    match table_meta.views.iter().nth(index as usize) {
+        Some(view_meta) => Arc::as_ptr(view_meta.1) as *const CassMaterializedViewMeta,
```
Lorak-mmk: Remove the cast (as_ptr already returns correct type)

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -0,0 +1,502 @@
...
+    let view_meta = ptr_to_ref(view_meta);
+    write_str_to_c(view_meta.name.as_str(), name, name_length)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_materialized_view_meta_base_table(
+    view_meta: *const CassMaterializedViewMeta,
+) -> *const CassTableMeta {
+    let view_meta = ptr_to_ref(view_meta);
+    view_meta.base_table.as_ptr() as *const CassTableMeta
```
Lorak-mmk: Is the type returned by `as_ptr()` different than `*const CassTableMeta`? If not, remove the cast

Lorak-mmk (changes requested):
I like the overall changes now (apart from few comments I left).
Few issues I have with the commits:
- "Add materialized view metadata ": message is incorrect, there is not RefCell anymore
- "Add impl for functions returning view metadata fields " - this commit should not include moving existing functions between files, that should be a seperate one.

scylla-rust-wrapper/src/metadata.rs
Gor027: Done.

scylla-rust-wrapper/src/metadata.rs
Gor027: The `udt` object is of type `&Arc<CassDataType>` and `Arc::into_raw` expects argument by value rather than by reference, so I think a clone is necessary here.

scylla-rust-wrapper/src/metadata.rs
Gor027: Done.

scylla-rust-wrapper/src/metadata.rs
Gor027: Done.

scylla-rust-wrapper/src/metadata.rs
Gor027: Done.

scylla-rust-wrapper/src/metadata.rs
Gor027: The returned type is `*const CassTableMeta`, so the cast is removed.

scylla-rust-wrapper/src/metadata.rs
Gor027: Ok, there is the `Arc::as_ptr` function which helps to avoid the clone. Will fix it.

Gor027:
Update:

Rebased on #95 and enabled materialized view metadata tests on GitHub Actions (`SchemaMetadataTest.*View*`).
This PR should be merged only after #95 is merged.

scylla-rust-wrapper/src/metadata.rs
Lorak-mmk: No it's not

Gor027: The cast perhaps appeared after the rebase on #95. Hope it is fixed now.

Lorak-mmk (approved):

Lorak-mmk:
Actually, one more changed is needed - adjustment of the limitations table in README to take into account changes made by this PR

Gor027:
@Lorak-mmk Guess it is good now to be merged.

Gor027:
Also, as you mentioned earlier, a nice optimization maybe not copy the metadata information, but have a pointer to a `ClusterData` object and on each metadata-related function call calculate and return the necessary information. I will open an issue to be discussed in the future.

PR #77, opened 2022-09-09
Add data type frozen

Gor027:
## Pre-review checklist

This PR is dependent upon #72 

Adds `frozen` field to `CassDataType` for collections including UDT. By default, it sets `frozen` to be false if there is not enough information provided.


Lorak-mmk:
This probably needs rebasing on current version of #72?
Are there any tests for this? It would be good to have some tests showing that this field can be true,
Apart from this, looks good.

roydahan:
This is P0 according to the plan.
Let's rebase and see if it's can easily get in.

muzarski:
I'm closing this in favor of https://github.com/scylladb/cpp-rust-driver/pull/141

PR #78, opened 2022-09-16
Add tracing subscriber to enable logging

Gor027:
Motivation:

Added `tracing-subscriber` crate to have one common way of logging with the Rust driver, as the Rust driver uses `tracing` crate for logging and cpp-rust driver should allow clients to configure logging by specifying the max log level and providing callback that should be called on each event.

Solution:

As logging should be configured globally, the current implementation creates a custom tracing subscriber with default configuration, which later can be changed by the user. The global and mutable state of the logging configuration is initialized in the `lazy_static!` block, however, it will be evaluated only after the first access in the code. As, the `cass_session_new()` function most probably is the first function that the user will call (almost all tests do that, too), the initialization of logging is chosen to be done in this function.

The CassLogMessage has a field `function` that should always be empty, as it cannot be properly set, because event metadata does not include information about it.

Enabled `LoggingTests` tests in github actions.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -1,4 +1,5 @@
 use crate::types::size_t;
+use std::cmp::min;
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::sync::Arc;
@@ -36,6 +37,29 @@ pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
 
 pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static str> {
     std::str::from_utf8(std::slice::from_raw_parts(ptr as *const u8, size as usize)).ok()
+}
+
+pub unsafe fn arr_to_cstr<const N: usize>(arr: &[c_char]) -> Option<&'static str> {
+    let null_char = '\0' as c_char;
+    let end_index = arr[..N].iter().position(|c| c == &null_char).unwrap_or(N);
+    ptr_to_cstr_n(arr.as_ptr(), end_index as size_t)
+}
+
+pub fn str_to_arr<const N: usize>(s: &str) -> [c_char; N] {
+    let mut result = ['\0' as c_char; N];
+
+    // Max length must be null-terminated
+    let mut max_len = min(N - 1, s.as_bytes().len());
+
+    while !s.is_char_boundary(max_len) {
+        max_len -= 1;
+    }
+
+    for (i, c) in s.as_bytes().iter().enumerate().take(max_len) {
+        result[i] = *c as c_char;
+    }
+
+    result
 }
 
 pub unsafe fn write_str_to_c(s: &str, c_str: *mut *const c_char, c_strlen: *mut size_t) {
--- a/scylla-rust-wrapper/src/lib.rs
+++ b/scylla-rust-wrapper/src/lib.rs
@@ -1,6 +1,12 @@
 #![allow(clippy::missing_safety_doc)]
+
+use crate::logging::set_tracing_subscriber_with_level;
+use crate::logging::stderr_log_callback;
+use crate::logging::Logger;
 use lazy_static::lazy_static;
+use std::sync::RwLock;
 use tokio::runtime::Runtime;
+use tracing::dispatcher::DefaultGuard;
 
 #[macro_use]
 mod binding;
@@ -13,6 +19,7 @@ pub mod collection;
 mod external;
 pub mod future;
 pub mod inet;
+mod logging;
 pub mod prepared;
 pub mod query_error;
 pub mod query_result;
@@ -28,6 +35,13 @@ pub mod uuid;
 
 lazy_static! {
     pub static ref RUNTIME: Runtime = Runtime::new().unwrap();
+    pub static ref LOGGER: RwLock<Logger> = RwLock::new(Logger {
+        cb: Some(stderr_log_callback),
+        data: std::ptr::null_mut(),
+    });
+    pub static ref LOG: RwLock<Option<DefaultGuard>> = RwLock::new(Some(
+        set_tracing_subscriber_with_level(tracing::Level::WARN)
+    ));
 }
 
 // To send a Rust object to C:
--- a/scylla-rust-wrapper/src/logging.rs
+++ b/scylla-rust-wrapper/src/logging.rs
@@ -1,0 +1,228 @@
+mod cass_log {
+    #![allow(non_camel_case_types)]
+    include!(concat!(env!("OUT_DIR"), "/cppdriver_log.rs"));
+}
+use crate::argconv::{arr_to_cstr, ptr_to_cstr, ptr_to_ref, str_to_arr};
+use crate::types::size_t;
+use crate::LOG;
+use crate::LOGGER;
+use cass_log::*;
+use std::convert::TryFrom;
+use std::fmt::Debug;
+use std::fmt::Write;
+use std::os::raw::{c_char, c_void};
+use std::time::{SystemTime, UNIX_EPOCH};
+use tracing::debug;
+use tracing::dispatcher::DefaultGuard;
+use tracing::field::Field;
+use tracing::Level;
+use tracing_subscriber::layer::Context;
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::Layer;
+
+pub type CassLogCallback =
+    Option<unsafe extern "C" fn(message: *const CassLogMessage, data: *mut c_void)>;
+
+unsafe extern "C" fn noop_log_callback(_message: *const CassLogMessage, _data: *mut c_void) {}
+
+pub struct Logger {
+    pub cb: CassLogCallback,
+    pub data: *mut c_void,
+}
+
+// The field `data` in the struct `Logger` is neither Send nor Sync.
+// It can be mutated only in the user-provided `CassLogCallback`, so it is safe
+// to implement Sync and Send manually for the `Logger` struct.
+unsafe impl Sync for Logger {}
+unsafe impl Send for Logger {}
+
+impl From<Level> for CassLogLevel {
+    fn from(level: Level) -> Self {
+        match level {
+            Level::TRACE => CassLogLevel::CASS_LOG_TRACE,
+            Level::DEBUG => CassLogLevel::CASS_LOG_DEBUG,
+            Level::INFO => CassLogLevel::CASS_LOG_INFO,
+            Level::WARN => CassLogLevel::CASS_LOG_WARN,
+            Level::ERROR => CassLogLevel::CASS_LOG_ERROR,
+        }
+    }
+}
+
+impl TryFrom<CassLogLevel> for Level {
+    type Error = String;
+
+    fn try_from(log_level: CassLogLevel) -> Result<Self, Self::Error> {
+        let level = match log_level {
+            CassLogLevel::CASS_LOG_TRACE => Level::TRACE,
+            CassLogLevel::CASS_LOG_DEBUG => Level::DEBUG,
+            CassLogLevel::CASS_LOG_INFO => Level::INFO,
+            CassLogLevel::CASS_LOG_WARN => Level::WARN,
+            CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
+            CassLogLevel::CASS_LOG_CRITICAL => Level::ERROR,
+            _ => return Err("Error while converting CassLogLevel to Level".to_string()),
+        };
+
+        Ok(level)
+    }
+}
+
+pub const CASS_LOG_MAX_MESSAGE_SIZE: usize = 1024;
+
+pub unsafe extern "C" fn stderr_log_callback(message: *const CassLogMessage, _data: *mut c_void) {
+    let message = ptr_to_ref(message);
+
+    eprintln!(
+        "{} [{}] ({}:{}) {}",
+        message.time_ms as u64,
+        ptr_to_cstr(cass_log_level_string(message.severity)).unwrap(),
+        ptr_to_cstr(message.file).unwrap(),
+        message.line,
+        arr_to_cstr::<{ CASS_LOG_MAX_MESSAGE_SIZE }>(&message.message).unwrap(),
+    )
+}
+
+pub struct CustomLayer;
+
+pub struct PrintlnVisitor {
+    log_message: String,
+}
+
+// Collects all fields and values in a single log event into a single String
+// to set into CassLogMessage::message.
+impl tracing::field::Visit for PrintlnVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn Debug) {
+        if self.log_message.is_empty() {
+            write!(self.log_message, "{}: {:?}", field, value).unwrap();
+        } else {
+            write!(self.log_message, ", {}: {:?}", field, value).unwrap();
+        }
+    }
+}
+
+impl<S> Layer<S> for CustomLayer
+where
+    S: tracing::Subscriber,
+{
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
+        let start = SystemTime::now();
+        let since_the_epoch = start
+            .duration_since(UNIX_EPOCH)
+            .expect("Current time is before UNIX_EPOCH");
+        let log_time_ms = since_the_epoch.as_millis() as u64;
+
+        let message = "";
+        let mut target = event.metadata().target().to_string();
+        target.push('\0');
+
+        let mut log_message = CassLogMessage {
+            time_ms: log_time_ms,
+            severity: (*event.metadata().level()).into(),
+            file: target.as_ptr() as *const c_char,
+            line: event.metadata().line().unwrap_or(0) as i32,
+            function: "\0".as_ptr() as *const c_char, // ignored, as cannot be fetched from event metadata
+            message: str_to_arr(message),
+        };
+
+        let mut visitor = PrintlnVisitor {
+            log_message: message.to_string(),
+        };
+        event.record(&mut visitor);
+
+        visitor.log_message.push('\0');
+        log_message.message =
+            str_to_arr::<{ CASS_LOG_MAX_MESSAGE_SIZE }>(visitor.log_message.as_str());
+
+        let logger = LOGGER.read().unwrap();
+
+        if let Some(log_cb) = logger.cb {
+            unsafe {
+                log_cb(&log_message as *const CassLogMessage, logger.data);
+            }
+        }
+    }
+}
+
+// Sets tracing subscriber with specified `level` and returns `DefaultGuard`.
+// The subscriber is valid for the duration of the lifetime of the returned `DefaultGuard`.
+pub fn set_tracing_subscriber_with_level(level: Level) -> DefaultGuard {
+    tracing::subscriber::set_default(
+        tracing_subscriber::registry()
+            .with(
+                tracing_subscriber::EnvFilter::from_default_env()
+                    .add_directive(level.to_owned().into()),
+            )
+            .with(CustomLayer),
+    )
+}
+
+pub fn init_logging() {
+    let _log = (*LOG.read().unwrap()).as_ref().unwrap();
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_level(log_level: CassLogLevel) {
+    if log_level == CassLogLevel::CASS_LOG_DISABLED {
+        debug!("Logging is disabled!");
+        *LOG.write().unwrap() = None;
+        return;
+    }
+
+    let level = Level::try_from(log_level).unwrap_or(Level::WARN);
+
+    // Drops the `DefaultGuard` of the current tracing subscriber making it invalid.
+    // Setting LOG to None is vital, otherwise, the new tracing subscriber created in
+    // `set_tracing_subscriber_with_level` will be ignored as the previous `DefaultGuard` is
+    // still alive.
+    *LOG.write().unwrap() = None;
+    // Sets the tracing subscriber with new log level.
+    *LOG.write().unwrap() = Some(set_tracing_subscriber_with_level(level));
+
+    debug!("Log level is set to {}", level);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_level_string(log_level: CassLogLevel) -> *const c_char {
+    let log_level_str = match log_level {
+        CassLogLevel::CASS_LOG_TRACE => "TRACE\0",
+        CassLogLevel::CASS_LOG_DEBUG => "DEBUG\0",
+        CassLogLevel::CASS_LOG_INFO => "INFO\0",
+        CassLogLevel::CASS_LOG_WARN => "WARN\0",
+        CassLogLevel::CASS_LOG_ERROR => "ERROR\0",
+        CassLogLevel::CASS_LOG_CRITICAL => "CRITICAL\0",
+        CassLogLevel::CASS_LOG_DISABLED => "DISABLED\0",
+        _ => "\0",
+    };
+
+    log_level_str.as_ptr() as *const c_char
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_callback(callback: CassLogCallback, data: *mut c_void) {
+    let logger = Logger {
+        cb: Some(callback.unwrap_or(noop_log_callback)),
+        data,
+    };
+
+    *LOGGER.write().unwrap() = logger;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_get_callback_and_data(
+    callback_out: *mut CassLogCallback,
+    data_out: *mut *const c_void,
+) {
+    let logger = LOGGER.read().unwrap();
+
+    *callback_out = logger.cb;
+    *data_out = logger.data;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_cleanup() {
+    // Deprecated
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_queue_size(_queue_size: size_t) {
+    // Deprecated
+}
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -7,6 +7,7 @@ use crate::cass_types::{
 use crate::cluster::build_session_builder;
 use crate::cluster::CassCluster;
 use crate::future::{CassFuture, CassResultValue};
+use crate::logging::init_logging;
 use crate::query_result::Value::{CollectionValue, RegularValue};
 use crate::query_result::{
     CassResult, CassResultData, CassResult_, CassRow, CassValue, Collection, Value,
@@ -65,6 +66,8 @@ pub struct CassSchemaMeta {
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_session_new() -> *const CassSession {
+    init_logging();
+
     let session: CassSession_ = Arc::new(RwLock::new(None));
     Arc::into_raw(session)
 }
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -18,28 +18,6 @@
 
 using namespace datastax::internal;
 
-extern "C" {
-
-void cass_log_cleanup() {
-  // Deprecated
-}
-
-void cass_log_set_level(CassLogLevel log_level) { Logger::set_log_level(log_level); }
-
-void cass_log_set_callback(CassLogCallback callback, void* data) {
-  Logger::set_callback(callback, data);
-}
-
-void cass_log_get_callback_and_data(CassLogCallback* callback_out, void** data_out) {
-  Logger::get_callback_and_data(callback_out, data_out);
-}
-
-void cass_log_set_queue_size(size_t queue_size) {
-  // Deprecated
-}
-
-} // extern "C"
-
 namespace datastax { namespace internal { namespace core {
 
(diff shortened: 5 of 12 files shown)


scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,32 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+
+    while i < N {
+        if arr[i] != '\0' as c_char {
+            vec[i] = arr[i] as u8 as char;
+        }
+        i += 1;
+    }
+
+    vec.iter().collect::<String>()
+}
```
piodul: There are a couple of issues with this code:

- It takes an array of `N` characters by value as an argument, where `N` can be quite large (1024 in our case). You can change the type of the array either to a reference to an array, or just a slice (the function doesn't have to be generic IMO).
- It creates a 1024 byte buffer pre-populated with zero bytes, copies bytes from `arr` until it encounters `'\0` and then collects the whole 1024 byte buffer into a `String`. Please remember that in Rust, unlike in C, strings are _not_ null terminated, so this always creates a 1024 byte String. I can see that you are later trimming the zero bytes unnecessarily.
- It always allocates a string, however it could just borrow from the `arr` argument if it were passed by reference.

I think the implementation should be similar to:
- Find the first null terminator byte
- Use `ptr_to_cstr_n` to create the string, if there was no null terminator then use `N` as length.

We can consider using `String::from_utf8_lossy` instead of `std::str::from_utf8` so that if the string contains invalid characters, then the resulting string will still be printable.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,32 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+
+pub fn str_to_arr<const N: usize>(s: &str) -> [c_char; N] {
```
piodul, line 48: It's better to take the output argument as `&mut [c_char]` rather than returning a big slice so that you don't have to initialize the whole 1024 bytes every single time.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,32 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+    let mut result = ['\0' as c_char; N];
+
+    for (i, c) in s.chars().enumerate() {
```
piodul: `s.chars().enumerate().take(N)`

If you change it like this, then you won't have to put the `i < N` check inside the loop.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +42,96 @@ pub mod types;
...
+
+pub struct PrintlnVisitor {
+    log_message: String,
+}
+
+// Collects all fields and values in a single log event into a single String
+// to set into CassLogMessage::message.
+impl tracing::field::Visit for PrintlnVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn Debug) {
+        self.log_message = format!("{}{}: {:?}, ", self.log_message, field, value);
```
piodul: This is not a very efficient way to append strings. It forces you to allocate every time you add a field. Consider using the `+=` operator instead.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +42,96 @@ pub mod types;
...
+where
+    S: tracing::Subscriber,
+{
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
+        let start = SystemTime::now();
+        let since_the_epoch = start
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards");
+        let log_time_ms =
+            since_the_epoch.as_secs() * 1000 + since_the_epoch.subsec_nanos() as u64 / 1_000_000;
```
piodul: How about [`Duration::as_millis`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_millis)?

Gor027: `Duration::as_millis` returns `u128`, however, `CassLogMessage::time_ms` is of type `u64`.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_level(log_level: CassLogLevel) {
+    let log_level = match log_level {
+        CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
+        CassLogLevel::CASS_LOG_WARN => Level::WARN,
+        CassLogLevel::CASS_LOG_INFO => Level::INFO,
+        CassLogLevel::CASS_LOG_DEBUG => Level::DEBUG,
+        CassLogLevel::CASS_LOG_TRACE => Level::TRACE,
+        _ => Level::INFO, // TODO gor: print warning that flag is not supported
```
constantfold: Should this `TODO` be done before merging?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,99 @@
...
+    let log_level_str = match log_level {
+        CassLogLevel::CASS_LOG_CRITICAL => "CRITICAL\0",
+        CassLogLevel::CASS_LOG_ERROR => "ERROR\0",
+        CassLogLevel::CASS_LOG_WARN => "WARN\0",
+        CassLogLevel::CASS_LOG_INFO => "INFO\0",
+        CassLogLevel::CASS_LOG_DEBUG => "DEBUG\0",
+        CassLogLevel::CASS_LOG_TRACE => "TRACE\0",
+        CassLogLevel::CASS_LOG_DISABLED => "DISABLED\0",
+        CassLogLevel::CASS_LOG_LAST_ENTRY => "\0",
+        _ => "\0",
```
constantfold:
```suggestion
        _ => "\0",
```

constantfold: `CASS_LOG_LAST_ENTRY` is a dummy enum entry, so it shouldn't be mentioned anywhere.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +42,94 @@ pub mod types;
...
+
+pub struct PrintlnVisitor {
+    log_message: String,
+}
+
+// Collects all fields and values in a single log event into a single String
+// to set into CassLogMessage::message.
+impl tracing::field::Visit for PrintlnVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn Debug) {
+        self.log_message += format!("{}: {:?}, ", field, value).as_str();
```
piodul: Hmm, it's not exactly what I meant. This still requires you to allocate a string - `format!` performs an allocation. On the other hand you have to convert the `value` to string, so I don't think that it's actually possible to avoid an allocation with `+=`

However, it looks like you can use the `write!` macro to do what I originally meant, see here: https://stackoverflow.com/questions/28333612/how-can-i-append-a-formatted-string-to-an-existing-string

scylla-rust-wrapper/src/argconv.rs
piodul: What about this?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,31 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+
+    ptr_to_cstr_n(arr.as_ptr(), N as size_t)
+}
+
+pub fn str_to_arr<const N: usize>(s: &str) -> [c_char; N] {
+    let mut result = ['\0' as c_char; N];
+
+    for (i, c) in s.chars().enumerate() {
+        if i < N {
+            result[i] = c as c_char;
```
piodul: This won't work if you have a non-ascii string. Perhaps it would be good to just take the bytes of the string and copy it to the `result`. In the meantime, please be careful not to accidentally slice a character - UTF-8 can use multiple bytes to encode a single character, and we either want a whole character or none of them. See here for an implementation of a function which returns a slice that has max N bytes: https://github.com/rust-lang/rfcs/issues/2566

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +43,94 @@ pub mod types;
...
+        cb: Some(stderr_log_callback),
+        data: std::ptr::null_mut(),
+    });
+    pub static ref LOG: () = tracing::subscriber::set_global_default(
+        tracing_subscriber::registry()
+            .with(
+                tracing_subscriber::EnvFilter::from_default_env()
+                    .add_directive(LOG_LEVEL.read().unwrap().to_owned().into())
+            )
+            .with(CustomLayer)
```
piodul: Why are we only allowed to do this once per program? Does the `tracing` crate forbid to change the subscriber later?

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +43,94 @@ pub mod types;
...
+
+pub struct PrintlnVisitor {
+    log_message: String,
+}
+
+// Collects all fields and values in a single log event into a single String
+// to set into CassLogMessage::message.
+impl tracing::field::Visit for PrintlnVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn Debug) {
+        write!(self.log_message, "{}: {:?}, ", field, value).unwrap();
```
piodul: It's a bit weird because it leaves a comma at the end. How about checking if the string is empty - if it is, then don't add a comma, if it isn't then put a comma before the appended fragment?

`, {}: {:?}`

scylla-rust-wrapper/src/lib.rs
Gor027: As the documentation of the C++ driver suggests, the logging properties such as the level or the callback should be set before calling any function that might log. I am not sure if the crate forbids changing the subscriber runtime. But I also do not see how it may be useful for the users, and from which API function the subscriber should be changed.

piodul: OK.

scylla-rust-wrapper/src/session.rs
```diff
@@ -22,12 +23,21 @@ use std::os::raw::c_char;
...
 use tokio::sync::RwLock;
+use tracing::warn;
 
 pub type CassSession = RwLock<Option<Session>>;
 type CassSession_ = Arc<CassSession>;
 
+#[allow(clippy::let_unit_value)]
+fn init_logging() {
+    let _log = LOG.to_owned();
+    warn!(message = "Logging initialized, do not call log-related functions 'cass_log_*'!");
```
piodul: I users will find it annoying that a warning is printed when the driver is used as intended. How about printing a warning only when the user actually uses some `cass_log_` function after the log is initialized?

Gor027: Ok, will fix it.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,34 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+
+    while i < N {
+        i += 1;
+        if arr[i] == '\0' as c_char {
+            return ptr_to_cstr_n(arr.as_ptr(), i as size_t);
+        }
+    }
+
+    ptr_to_cstr_n(arr.as_ptr(), N as size_t)
+}
```
Lorak-mmk: This code doesn't check for null at the beggining of the message - is this intended?
If not, then something like this should work:
```
let end_index = arr[..N].iter().position(|c| c == '\0').unwrap_or(N);
ptr_to_cstr_n(arr.as_ptr(), end_index as size_t)
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -38,6 +38,34 @@ pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static
...
+
+pub fn str_to_arr<const N: usize>(s: &str) -> [c_char; N] {
+    let mut result = ['\0' as c_char; N];
+    let mut max_len = N - 1;
```
Lorak-mmk: A commend saying this it must be `N-1` instead of `N` to have null-terminated string would be appreciated

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+
+unsafe extern "C" fn noop_log_callback(_message: *const CassLogMessage, _data: *mut c_void) {}
+
+pub struct Logger {
+    pub cb: CassLogCallback,
+    pub data: *mut c_void,
+}
+
+unsafe impl Sync for Logger {}
+unsafe impl Send for Logger {}
```
Lorak-mmk, line 37: I think this warrants a comment.
- Why are those not auto implemented (which members of Logger are not Send/Sync)
- Why is this safe?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_level(log_level: CassLogLevel) {
+    let level = match log_level {
+        CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
+        CassLogLevel::CASS_LOG_WARN => Level::WARN,
+        CassLogLevel::CASS_LOG_INFO => Level::INFO,
+        CassLogLevel::CASS_LOG_DEBUG => Level::DEBUG,
+        CassLogLevel::CASS_LOG_TRACE => Level::TRACE,
+        _ => Level::INFO,
+    };
```
Lorak-mmk: Can you implement `TryFrom<CassLogLevel> for Level` and use it?
If it's not possible because of orphan rule, maybe implement `TryInto<Level> for CassLogLevel`?
What do you think @piodul ?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    // Drops the `DefaultGuard` of the tracing subscriber making it invalid.
+    *LOG.write().unwrap() = None;
```
Lorak-mmk, line 176: Why is this necessary? Wouldn't the next line drop the subscriber anyway?

src/logger.cpp
```diff
@@ -18,28 +18,6 @@
...
-void cass_log_get_callback_and_data(CassLogCallback* callback_out, void** data_out) {
-  Logger::get_callback_and_data(callback_out, data_out);
-}
-
-void cass_log_set_queue_size(size_t queue_size) {
-  // Deprecated
-}
-
-} // extern "C"
-
```
Lorak-mmk, line 42: Is the rest of this file used by anything, or can the whole file be removed?

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Also, N will usually be 1024 - so this most likely won't be a performance problem, but maybe we could do something like `let mut max_len = std::min(N - 1, s.as_bytes().len());` just to be safe?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/logging.rs
Gor027: A `Logger` object contains `data: *mut libc::c_void` for which neither `Send` nor `Sync` is implemented. It should be safe to manually implement `Send` and `Sync` for `Logger` as `data` can be mutated only in `CassLogCallback` which is defined by the user.

scylla-rust-wrapper/src/logging.rs
Gor027: Unfortunately, `lazy_static` is not freed that simply and I needed to force a free here. I did some research and the simplest way to do so was using `Option`.

src/logger.cpp
Gor027: `logger.hpp` is used in the integration tests, so perharps, it cannot be removed safely.

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: This seems done - marking as resolved

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Looks done, marking as resolved

scylla-rust-wrapper/src/lib.rs
Lorak-mmk: Looks done, marking as resolved

scylla-rust-wrapper/src/lib.rs
Lorak-mmk: Looks done, marking as resolved

scylla-rust-wrapper/src/lib.rs
Lorak-mmk: Looks done, marking as resolved

scylla-rust-wrapper/src/session.rs
Lorak-mmk: Looks done, marking as resolved

scylla-rust-wrapper/src/logging.rs
Lorak-mmk: Please put this explanation in the comment in code

scylla-rust-wrapper/src/logging.rs
Lorak-mmk: So what would happen if you simply did
`*LOG.write().unwrap() = Some(set_tracing_subscriber_with_level(level));`
without doing
`*LOG.write().unwrap() = None;`
first?
Would the memory be leaked? This seems a bit counterintuitive

Gor027: The `set_tracing_subscriber_with_level` sets the tracing subscriber as default for the duration of the lifetime of the returned `DefaultGuard`. If the `LOG` is not set to `None` then, I guess, the `DefaultGuard` still lives and the new tracing subscriber is not set as default. You can test the behavior with this test case `LoggingTests.Integration_Cassandra_Callback`.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn from(level: &Level) -> Self {
+        match *level {
+            Level::INFO => CassLogLevel::CASS_LOG_INFO,
+            Level::DEBUG => CassLogLevel::CASS_LOG_DEBUG,
+            Level::WARN => CassLogLevel::CASS_LOG_WARN,
+            Level::ERROR => CassLogLevel::CASS_LOG_ERROR,
+            Level::TRACE => CassLogLevel::CASS_LOG_TRACE,
+        }
+    }
+}
```
Lorak-mmk: `Level` is `Copy`, so this can most likely be changed to `From<Level>` without any problems

scylla-rust-wrapper/src/logging.rs
Lorak-mmk: Ok, I get it. If you just did `*LOG.write().unwrap() = Some(set_tracing_subscriber_with_level(level));`, then the RHS would be evaluated first (and `set_tracing_subscriber_with_level` would do nothing, because previous guard still lives), and only then the previous guard would be dropped. This would result in no messages being printed afterwards according to my test. Is that correct?

I don't like this, but I don't see a better solution off the top of my head. Please comment it a bit more because it's hard to understand.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/logging.rs
Gor027: > Ok, I get it. If you just did `*LOG.write().unwrap() = Some(set_tracing_subscriber_with_level(level));`, then the RHS would be evaluated first (and `set_tracing_subscriber_with_level` would do nothing, because previous guard still lives), and only then the previous guard would be dropped. This would result in no messages being printed afterwards according to my test. Is that correct?

I think that is the proper explanation. I will leave more descriptions of the solution.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,125 @@
...
+    type Error = ();
+
+    fn try_from(log_level: CassLogLevel) -> Result<Self, Self::Error> {
+        let level = match log_level {
+            CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
+            CassLogLevel::CASS_LOG_WARN => Level::WARN,
+            CassLogLevel::CASS_LOG_INFO => Level::INFO,
+            CassLogLevel::CASS_LOG_DEBUG => Level::DEBUG,
+            CassLogLevel::CASS_LOG_TRACE => Level::TRACE,
+            _ => Level::INFO,
```
Lorak-mmk: Why the default option is `Level::Info` instead of returning an error?
If there is a reason, then why is this `TryFrom` and not simply `From`? There are currently no paths returning and error

Gor027: There are unmatched `CassLogLevel` types such as `CASS_LOG_CRITICAL` which I thought is better to match as `CASS_LOG_INFO` or `CASS_LOG_WARN`.

scylla-rust-wrapper/src/cass_log.rs
```diff
@@ -0,0 +1,3 @@
+#![allow(non_camel_case_types)]
+
+include!(concat!(env!("OUT_DIR"), "/cppdriver_log.rs"));
```
Lorak-mmk: Why is there a separate file for this instead of putting it in logging.rs?

scylla-rust-wrapper/src/logging.rs
Lorak-mmk: Such levels should be listed explicitly - CassLogLevel is not a Rust enum, it's an int and can have invalid value, so right now there is no difference between `CASS_LOG_CRITICAL` and arbitrary int value.
I think this should by TryFrom with unknown values returning error - you can then handle the error in `cass_log_set_level` using e.g. `unwrap_or` if we don't want to validate the value.

Also, `CASS_LOG_CRITICAL` should probably map to `Level::ERROR`, not `Level::INFO`

scylla-rust-wrapper/src/lib.rs
```diff
@@ -26,8 +44,104 @@ pub mod types;
...
+
 lazy_static! {
     pub static ref RUNTIME: Runtime = Runtime::new().unwrap();
+    pub static ref LOGGER: RwLock<Logger> = RwLock::new(Logger {
+        cb: Some(stderr_log_callback),
+        data: std::ptr::null_mut(),
+    });
+    pub static ref LOG: RwLock<Option<DefaultGuard>> = RwLock::new(Some(
+        set_tracing_subscriber_with_level(tracing::Level::WARN)
+    ));
```
Lorak-mmk: All of this new code is related to logging - wouldn't it fit better in logging.rs?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+pub fn init_logging() {
+    let _log = (*LOG.read().unwrap()).as_ref().unwrap();
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_level(log_level: CassLogLevel) {
+    let level = Level::try_from(log_level).ok().unwrap();
```
Lorak-mmk: Why .ok().unwrap() instead of only .unwrap()?
Also, after changing `From<CassLogLevel> for Level` to `TryFrom<CassLogLevel> for Level` this will need to be changed to `.unwrap_or`

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,121 @@
...
+    let logger = match callback {
+        Some(log_cb) => Logger {
+            cb: Some(log_cb),
+            data,
+        },
+        None => Logger {
+            cb: Some(noop_log_callback),
+            data,
+        },
+    };
```
Lorak-mmk: Maybe something like this?
```suggestion
    let logger = Logger {
            cb: Some(callback.unwrap_or(noop_log_callback)),
            data,
        };
```

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cass_log.rs
Gor027: There was a need to manually disable `non_camel_case_types` warnings to pass cargo clippy checks, so this was a NIT suggestion by @piodul to separate the logic into a new file.

scylla-rust-wrapper/src/lib.rs
Gor027: Yes, for some reason I left it here, but it is possible to move it to `logging.rs`. Will do it, thanks.

scylla-rust-wrapper/src/cass_log.rs
Lorak-mmk: That's a good point. I have 2 other ideas on how to solve it - not sure any of them will work, but could you try?
1. Put the include in module and disable warnings for module.
```
#[allow(non_camel_case_types)]
mod cass_log {
    include!(concat!(env!("OUT_DIR"), "/cppdriver_log.rs"));
}
use cass_log::*;
```
2. Use https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.module_raw_line to add this `allow` to the module itself - but that will probably affect also the file it is `!include`'d in, so the option 1 seems better.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,234 @@
...
+    let _log = (*LOG.read().unwrap()).as_ref().unwrap();
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_log_set_level(log_level: CassLogLevel) {
+    if log_level == CassLogLevel::CASS_LOG_CRITICAL {
+        debug!("Logging is disabled!");
+        *LOG.write().unwrap() = None;
+        return;
+    }
```
Lorak-mmk: Why is logging disabled when set to CRITICAL?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,234 @@
...
+unsafe impl Send for Logger {}
+
+impl From<Level> for CassLogLevel {
+    fn from(level: Level) -> Self {
+        match level {
+            Level::INFO => CassLogLevel::CASS_LOG_INFO,
+            Level::DEBUG => CassLogLevel::CASS_LOG_DEBUG,
+            Level::WARN => CassLogLevel::CASS_LOG_WARN,
+            Level::ERROR => CassLogLevel::CASS_LOG_ERROR,
+            Level::TRACE => CassLogLevel::CASS_LOG_TRACE,
```
Lorak-mmk: Can we put that in order (trace, debug, info, warn, error)?

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,234 @@
...
+    type Error = String;
+
+    fn try_from(log_level: CassLogLevel) -> Result<Self, Self::Error> {
+        let level = match log_level {
+            CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
+            CassLogLevel::CASS_LOG_WARN => Level::WARN,
+            CassLogLevel::CASS_LOG_INFO => Level::INFO,
+            CassLogLevel::CASS_LOG_DEBUG => Level::DEBUG,
+            CassLogLevel::CASS_LOG_TRACE => Level::TRACE,
+            CassLogLevel::CASS_LOG_CRITICAL => Level::ERROR,
```
Lorak-mmk: Same here - either put critical on top or reverse the order of other fields.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl<S> Layer<S> for CustomLayer
+where
+    S: tracing::Subscriber,
+{
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
+        let start = SystemTime::now();
+        let since_the_epoch = start
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards");
```
Lorak-mmk: This error message can be improved - `Time wen backwards` can mean that the system clock was not monotonic - which can often happen. I'd go with `Current time is before UNIX_EPOCH`.

scylla-rust-wrapper/src/logging.rs
Lorak-mmk: This comment still stands.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/logging.rs
Gor027: It was a copy-paste issue, thanks for noticing.

scylla-rust-wrapper/src/logging.rs
Gor027: I have applied the suggestion in the same commit.

scylla-rust-wrapper/src/logging.rs
Gor027: Done.

scylla-rust-wrapper/src/logging.rs
Gor027: Done.

scylla-rust-wrapper/src/logging.rs
Gor027: Done.

scylla-rust-wrapper/src/logging.rs
Gor027: Done.

Lorak-mmk (approved):

PR #79, opened 2022-09-19
Fix cass_error_desc argument type

Gor027:
The argument of enum type CassError is passed by value, earlier the function got the argument by pointer reference and tried to dereference it, which led to some undefined behavior.

## Pre-review checklist


scylla-rust-wrapper/src/external.rs
```diff
@@ -1,11 +1,9 @@
-use crate::argconv::ptr_to_ref;
 use crate::cass_error::*;
 use std::os::raw::c_char;
 
 #[no_mangle]
-pub unsafe extern "C" fn cass_error_desc(error: *const CassError) -> *const c_char {
-    let error: &CassError = ptr_to_ref(error);
-    let desc = match *error {
+pub unsafe extern "C" fn cass_error_desc(error: CassError) -> *const c_char {
+    let desc = match error {
```
piodul, line 6: Could you add a simple test which verifies that the function works now? No need to check all error messages, one or two should suffice - just to make sure that the function works at all when called from cpp.

Gor027: Yes, it's worth adding tests for this. By default, Rust's string types are not null terminated, however, the API of the driver required that the `const char*` being passed to them is null-terminated.

piodul: Oh, so there was another bug. It would be nice then if we made sure that all strings are null-terminated - I think you can just extend the cpp test to test all error descriptions. You can probably use the `CASS_ERROR_MAPPING(XX)` macro to keep the test short and simple. It would also be good to test the unknown error case so that it returns a valid, empty string.

Gor027: I have added a test to check the returned error description for all `CassError` types including `CASS_OK` and `CASS_ERROR_LAST_ENTRY`. Fortunately, this was the only place where a Rust string is returned from an API function.

piodul (approved):

PR #81, opened 2022-09-21
Add extensive information about the driver in README

Gor027:
## Pre-review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,313 @@
-# cpp-rust-driver
-Wrapper around ScyllaDB's rust-driver compatible with Datastax cpp-driver (WIP)
+# ScyllaDB Cpp-Rust Driver
+___
+Wrapper around ScyllaDB's rust-driver compatible with Datastax cpp-driver.
+
+#### Note: It is work in progress, bug reports and pull requests are welcome!
+
+# Examples
+___
+
+There are some examples in the [examples](https://github.com/scylladb/cpp-rust-driver/tree/master/examples) directory.
+
+```c++
+#include <cassandra.h>
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+  /* Setup and connect to cluster */
+  CassFuture* connect_future = NULL;
+  CassCluster* cluster = cass_cluster_new();
+  CassSession* session = cass_session_new();
+  char* hosts = "127.0.0.1";
+  if (argc > 1) {
+    hosts = argv[1];
+  }
+
+  /* Add contact points */
+  cass_cluster_set_contact_points(cluster, hosts);
+
+  /* Provide the cluster object as configuration to connect the session */
+  connect_future = cass_session_connect(session, cluster);
+
+  if (cass_future_error_code(connect_future) == CASS_OK) {
+    CassFuture* close_future = NULL;
+
+    /* Build statement and execute query */
+    const char* query = "SELECT release_version FROM system.local";
+    CassStatement* statement = cass_statement_new(query, 0);
+
+    CassFuture* result_future = cass_session_execute(session, statement);
+
+    if (cass_future_error_code(result_future) == CASS_OK) {
+      /* Retrieve result set and get the first row */
+      const CassResult* result = cass_future_get_result(result_future);
+      const CassRow* row = cass_result_first_row(result);
+
+      if (row) {
+        const CassValue* value = cass_row_get_column_by_name(row, "release_version");
+
+        const char* release_version;
+        size_t release_version_length;
+        cass_value_get_string(value, &release_version, &release_version_length);
+        printf("release_version: '%.*s'\n", (int)release_version_length, release_version);
+      }
+
+      cass_result_free(result);
+    } else {
+      /* Handle error */
+      const char* message;
+      size_t message_length;
+      cass_future_error_message(result_future, &message, &message_length);
+      fprintf(stderr, "Unable to run query: '%.*s'\n", (int)message_length, message);
+    }
+
+    cass_statement_free(statement);
+    cass_future_free(result_future);
+
+    /* Close the session */
+    close_future = cass_session_close(session);
+    cass_future_wait(close_future);
+    cass_future_free(close_future);
+  } else {
+    /* Handle error */
+    const char* message;
+    size_t message_length;
+    cass_future_error_message(connect_future, &message, &message_length);
+    fprintf(stderr, "Unable to connect: '%.*s'\n", (int)message_length, message);
+  }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,169 @@
-# cpp-rust-driver
-Wrapper around ScyllaDB's rust-driver compatible with Datastax cpp-driver (WIP)
+# ScyllaDB Cpp-Rust Driver
+___
+Wrapper around ScyllaDB's rust-driver compatible with Datastax cpp-driver.
+As opposed to C/C++ driver, it works with the Cassandra Query Language v4 (CQL4).
```
constantfold: What do you mean by Cassandra Query Language v4? It looks like C/C++ driver works with `CASS_PROTOCOL_VERSION_V4`.

README.md
```diff
@@ -1,12 +1,169 @@
...
+Now, use `--gtest_filter` to run certain integration tests:
+
+```shell
+./cassandra-integration-tests --scylla --install-dir=[PATH_TO_SCYLLA] --version=3.0.8 --category=CASSANDRA --verbose=ccm --gtest_filter="ClusterTests.*"
+```
+
+##### Note: Tests that pass with ScyllaDB and Cassandra clusters can be found in GitHub Actions [`build.yml`](https://github.com/scylladb/cpp-rust-driver/blob/master/.github/workflows/build.yml) and [`cassandra.yml`](https://github.com/scylladb/cpp-rust-driver/blob/master/.github/workflows/cassandra.yml) workflows.
+
+
+# Features
```
constantfold: This should be moved before "Testing" section.

constantfold:
Please also add a section about limitations. Additionally, we should clean up the GitHub Issues that were opened - closing the ones that are now implemented.

README.md
Gor027: This was mentioned in the [C++ driver description](https://github.com/scylladb/cpp-driver#compatibility):
#### This driver works exclusively with the Cassandra Query Language v3 (CQL3) and Cassandra's native protocol.

Perhaps, it is outdated information.

README.md
Gor027: Do you mean at the beginning of the `Testing` section, or in the description at the very beginning of the document?

constantfold: I mean that the entire "Features" section should be swapped with the entire "Testing" section, changing the order.

README.md
```diff
@@ -1,12 +1,167 @@
...
+# ScyllaDB Cpp-Rust Driver
+___
+Wrapper around ScyllaDB's rust-driver compatible with Datastax cpp-driver.
+
+#### Note: It is work in progress, bug reports and pull requests are welcome!
+
+# Examples
+___
+
+There are some examples in the [examples](https://github.com/scylladb/cpp-driver/tree/master/examples) directory.
```
Lorak-mmk: We have those examples in cpp-rust-driver repo, why point to the old repository?

README.md
```diff
@@ -1,12 +1,167 @@
...
+  * Assuming `scylla` is installed and built in the release mode <br> ``` ln -s [PATH_TO_TOOLS]/scylla-tools-java [PATH_TO_SCYLLA]/resources/cassandra```
+
+Finally, to build the integration tests:
 
 ```
 mkdir build && cd build
-cmake -DCASS_BUILD_INTEGRATION_TESTS=ON  ..
-make
-./cassandra-integration-tests
+cmake -DCASS_BUILD_INTEGRATION_TESTS=ON  .. & make
```
Lorak-mmk: `&` -> `&&`

Lorak-mmk (changes requested):

Lorak-mmk (approved):

Lorak-mmk (commented):
Actually, the limitations section mentiond by @avelanarius is important - please write it before merging.

Gor027:
Sure, I am working on it. I have updated the issues marking functions that are implemented and functions that are still in progress. I will base the limitations section on those issues unless you have another opinion.

Gor027:
V2:

Added a separate section about the limitations of some features.
More information will be added to this section. For now, it covers the main limitations that the current implementation of the driver's features has.

Lorak-mmk:
I'd like for the limitations section to include a note that the section itself may be incomplete - unless you are sure that the list is fairly complete?

One thing I'm not sure about:
In bindings.rs, the function that checks types is unimplemented: https://github.com/scylladb/cpp-rust-driver/blob/5c009e88c6e3d0021d4318855053b0ef514da42c/scylla-rust-wrapper/src/binding.rs#L53-L56
This I think is used for all binding - and you only noted that we are not checking types for `cass_collection_append_*` - is that correct?

Gor027:
> I'd like for the limitations section to include a note that the section itself may be incomplete - unless you are sure that the list is fairly complete?

Sure, I will add a note that the list is incomplete. For now, I have included the main functions which are not implemented or are implemented with some limitations. Functions that are relatively easy to implement (that do not require some implementation in the Rust driver, or maybe even in ScyllaDB) are just noted as `unimplemented`.

> This I think is used for all binding - and you only noted that we are not checking types for `cass_collection_append_*` - is that correct?

What do you mean by `all binding`? The compatibility of types is being checked while appending or setting values in collections, tuples and UDTs. For binding a value to a prepared statement, the type is not being checked and in case of binding with incompatible type, the execution of the prepared statement should not be successful.

Lorak-mmk:
Are you sure that prepared statements are not checked? I did a quick experiment: https://gist.github.com/Lorak-mmk/dbce7e0484b69000f8523df4ad963531

And for me it prints: 
```
===== Using optimized driver!!! =====
bind 1 id: 0
bind 1 somestring: 0
execute 1: 0
bind 2 id: 0
bind 2 somestring: 16777229
execute 2: 0
DONE!
```
Which would mean that they are checked. After reading abstract_data.hh it also seems that the types should be checked for prepared statement.

Gor027:
> Which would mean that they are checked. After reading abstract_data.hh it also seems that the types should be checked for prepared statement.

Yeah, I guess you are right. The `UserTypeValue` and `Statement` classes extend from `AbstractData` class, so setting value by index or name should also check the type compatibility. Will fix the description.

Edit: The Rust driver, too, does not validate the types passed to queries [here](https://github.com/scylladb/scylla-rust-driver/issues/463), however, we can add the validation of types independently from the Rust driver.

Lorak-mmk:
There is already a placeholder function for that, it's just not implemented yet - always returns true.
See bindings.rs:53-56

Lorak-mmk (approved):

PR #84, opened 2022-10-12
Fix binding by name

Gor027:
Fixes case sensitive binding by name:
 * binding by name inside double quotes is case sensitive
 * binding by name without quotes should bind the value to multiple columns

Enabled `ByNameTests` tests for both ScyllaDB and Cassandra

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -739,14 +739,24 @@ pub unsafe extern "C" fn cass_row_get_column_by_name_n(
     name_length: size_t,
 ) -> *const CassValue {
     let row_from_raw = ptr_to_ref(row);
-    let name_str = ptr_to_cstr_n(name, name_length).unwrap().to_lowercase();
+    let mut name_str = ptr_to_cstr_n(name, name_length).unwrap();
+    let mut is_case_sensitive = false;
+
+    if name_str.starts_with('\"') && name_str.ends_with('\"') {
+        name_str = name_str.strip_prefix('\"').unwrap();
+        name_str = name_str.strip_suffix('\"').unwrap();
+        is_case_sensitive = true;
+    }
 
     return row_from_raw
         .result_metadata
         .col_specs
         .iter()
         .enumerate()
-        .find(|(_, spec)| spec.name.to_lowercase() == name_str)
+        .find(|(_, spec)| {
+            is_case_sensitive && spec.name == name_str
+                || !is_case_sensitive && spec.name.eq_ignore_ascii_case(name_str)
+        })
         .map(|(index, _)| {
             return match row_from_raw.columns.get(index) {
                 Some(value) => value as *const CassValue,
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -48,22 +48,52 @@ impl CassStatement {
         }
     }
 
+    fn bind_multiple_values_by_name(
+        &mut self,
+        indices: &[usize],
+        value: Option<CqlValue>,
+    ) -> CassError {
+        for i in indices {
+            let bind_status = self.bind_cql_value(*i, value.clone());
+
+            if bind_status != CassError::CASS_OK {
+                return bind_status;
+            }
+        }
+
+        CassError::CASS_OK
+    }
+
     fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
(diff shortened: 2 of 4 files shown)


scylla-rust-wrapper/src/statement.rs
```diff
@@ -48,22 +48,53 @@ impl CassStatement {
...
                     .iter()
                     .enumerate()
-                {
-                    if col.name == name {
-                        return self.bind_cql_value(i, value);
-                    }
+                    .filter(|(_, col)| {
+                        is_case_sensitive && col.name == name_str
+                            || !is_case_sensitive
+                                && col.name.to_lowercase() == name_str.to_lowercase()
```
Lorak-mmk: Instead of using `to_lowercase()` which needs to allocate new string, please write a new function that compares strings case-insensitive without allocating

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -403,14 +403,22 @@ pub unsafe extern "C" fn cass_row_get_column_by_name_n(
...
-    let name_str = ptr_to_cstr_n(name, name_length).unwrap().to_lowercase();
+    let mut name_str = ptr_to_cstr_n(name, name_length).unwrap().to_string();
+
+    if name_str.starts_with('\"') {
+        name_str = name_str.strip_prefix('\"').unwrap().to_string();
+    }
+
+    if name_str.ends_with('\"') {
+        name_str = name_str.strip_suffix('\"').unwrap().to_string();
+    }
```
Lorak-mmk: Why is this logic different than the one in `bind_cql_value_by_name`? The other one looks correct (from the cppdriver compatibility POV), this one doesn't.

Lorak-mmk: Also, can you prepare some abstraction for this (finding and stripping quotes, comparing case sensitive / case-insensitive) to avoid code repetition?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -48,22 +48,53 @@ impl CassStatement {
...
+    }
+
     fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
         let mut set_bound_val_index: Option<usize> = None;
+        let mut name_str = name.to_string();
+        let mut is_case_sensitive = false;
+
+        if name_str.starts_with('\"') && name_str.ends_with('\"') {
+            name_str = name_str.strip_prefix('\"').unwrap().to_string();
+            name_str = name_str.strip_suffix('\"').unwrap().to_string();
```
Lorak-mmk: Why `to_string()`? It would be better to only use `&str` to avoid allocations

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: [eq_ignore_ascii_case](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) could maybe be enough - original cpp driver doesn't handle utf8 at all.

scylla-rust-wrapper/src/query_result.rs
Gor027: In `bind_cql_value_by_name`, if the `name` is inside quotes then it looks for a case-sensitive match, otherwise, if quotes are not present, it looks for case-insensitive matches and binds the value to all of the matched columns. In `cass_row_get_column_by_name_n`, it always returns the first match.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -737,14 +737,19 @@ pub unsafe extern "C" fn cass_row_get_column_by_name_n(
     name_length: size_t,
 ) -> *const CassValue {
     let row_from_raw = ptr_to_ref(row);
-    let name_str = ptr_to_cstr_n(name, name_length).unwrap().to_lowercase();
+    let mut name_str = ptr_to_cstr_n(name, name_length).unwrap();
+
+    if name_str.starts_with('\"') && name_str.ends_with('\"') {
+        name_str = name_str.strip_prefix('\"').unwrap();
+        name_str = name_str.strip_suffix('\"').unwrap();
+    }
```
Lorak-mmk: I'm looking at the cppdriver code, and it looks like here the comparision should also be case-insensitive if the name is quoted - just like in `bind_cql_value_by_name`

Lorak-mmk (approved):

PR #85, opened 2022-10-13
Fix `ConfigTests` tests

Gor027:
Added implementation for functions in `testing.cpp` used in `ConfigTests` tests:
 * get_connect_timeout_from_cluster
 * get_port_from_cluster
 * get_contact_points_from_cluster

To avoid generating bindings for functions that are directly under namespaces and may return objects of custom types, the `testing_utils` header is added as a median between the Rust implementation of those functions and the `testing.hpp` header.

Added `ConfigTests` tests to GitHub Actions for Cassandra and ScyllaDB.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/testing.rs
+++ b/scylla-rust-wrapper/src/testing.rs
@@ -1,1 +1,39 @@
+use crate::argconv::{free_boxed, ptr_to_ref, write_str_to_c};
+use crate::cluster::CassCluster;
+use crate::types::*;
+use std::os::raw::c_char;
 
+#[no_mangle]
+pub unsafe extern "C" fn testing_get_connect_timeout_from_cluster(
+    cluster: *mut CassCluster,
+) -> cass_uint16_t {
+    let cluster = ptr_to_ref(cluster);
+    cluster.session_builder.config.connect_timeout.as_millis() as cass_uint16_t
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn testing_get_contact_points_from_cluster(
+    cluster: *mut CassCluster,
+    contact_points: *mut *const c_char,
+    contact_points_length: *mut size_t,
+    contact_points_boxed: *mut *const String,
+) {
+    let cluster = ptr_to_ref(cluster);
+    let str = Box::new(cluster.contact_points.join(","));
+    let result = str.as_str();
+
+    write_str_to_c(result, contact_points, contact_points_length);
+    *contact_points_boxed = Box::into_raw(str);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn testing_free_contact_points_string(contact_points_boxed: *mut String) {
+    free_boxed(contact_points_boxed)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn testing_get_port_from_cluster(cluster: *mut CassCluster) -> cass_int32_t {
+    let cluster = ptr_to_ref(cluster);
+
+    cluster.port as cass_int32_t
+}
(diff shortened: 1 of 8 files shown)


scylla-rust-wrapper/src/testing.rs
```diff
@@ -1 +1,39 @@
...
+    let cluster = ptr_to_ref(cluster);
+    cluster.session_builder.config.connect_timeout.as_millis() as cass_uint16_t
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn testing_get_contact_points_from_cluster(
+    cluster: *mut CassCluster,
+    contact_points: *mut *const c_char,
+    contact_points_length: *mut size_t,
+    contact_points_boxed: *mut *const String,
```
Lorak-mmk, line 19: It probably doesn't matter, but  maybe this should be `*mut *mut`? We use it as `*mut String` when freeing.

scylla-rust-wrapper/build.rs
```diff
@@ -33,6 +33,20 @@ fn prepare_basic_types(out_path: &Path) {
...
+        .layout_tests(true)
+        .generate_comments(false)
+        .generate()
+        .expect("Unable to generate bindings");
+
+    testing_bindings
+        .write_to_file(out_path.join("testing_bindings.rs"))
+        .expect("Couldn't write bindings!");
+}
+
```
Lorak-mmk: Why are you generating those bindings? It doesn't look like your'e using them anywhere

src/testing_utils.cpp
```diff
@@ -0,0 +1,26 @@
...
+  throw std::runtime_error("UNIMPLEMENTED testing_get_retry_policy\n");
+}
+
+CASS_EXPORT const char* testing_get_server_name(CassFuture* future) {
+  throw std::runtime_error("UNIMPLEMENTED testing_get_server_name\n");
+}
+
+CASS_EXPORT void testing_set_record_attempted_hosts(CassStatement* statement, cass_bool_t enable) {
+  throw std::runtime_error("UNIMPLEMENTED testing_set_record_attempted_hosts\n");
+}
```
Lorak-mmk, line 26: Why are those implementations required? Can't you just remove them from testing_utils.h?

src/testing_utils.h
```diff
@@ -0,0 +1,34 @@
...
+  CASS_EXPORT cass_uint64_t testing_get_request_timeout_ms(const CassStatement* statement);
+
+  CASS_EXPORT const CassRetryPolicy* testing_get_retry_policy(const CassStatement* statement);
+
+  CASS_EXPORT const char* testing_get_server_name(CassFuture* future);
+
+  CASS_EXPORT void testing_set_record_attempted_hosts(CassStatement* statement, cass_bool_t enable);
+}
+
+#endif //CPPDRIVERV2_TESTING_UTILS_H
```
Lorak-mmk, line 34: This file is present twice for some reason, with a slight modification. Do we need it twice in our repository?

Also, if I'm understanding this correctly, those are signatures for Rust functions that are used in implementation of `testing.hpp`. If that's the case then:
1. Name should be different than `testing_utils.h` - something like `testing_rust_impls.h` maybe?
2. This file should not be written by hand, but rather generated - cbindgen is a tool to do that

Lorak-mmk (changes requested):

src/testing_utils.cpp
Gor027: We can safely remove them, but I guess this way it will be much easier to find which testing functions are yet to be implemented.

scylla-rust-wrapper/src/testing.rs
Gor027: I guess that does not have an effect on the generated code and rather is a question of documenting intent.

muzarski:
Closing in favor of https://github.com/scylladb/cpp-rust-driver/pull/163

PR #87, opened 2022-10-26
Resolve minor fixme and todo comments

Gor027:
Replaced `CASS_ERROR_LAST_ENTRY` with more descriptive error codes while converting `QueryError` to `CassError`.

Set fetching schema metadata flag in `SessionBuilder` config.

Added port validation in cluster configuration.

Imported Rust driver by commit hash 358eabbbb71c.
This enabled to use the `allow_remote_dcs_for_local_cl` flag while creating a `DcAwareRoundRobinPolicy`.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -6,14 +6,13 @@ impl From<&QueryError> for CassError {
     fn from(error: &QueryError) -> Self {
         match error {
             QueryError::DbError(db_error, _string) => CassError::from(db_error),
-            // FIXME: CASS_ERROR_LAST_ENTRY as placeholder for a correct error
-            QueryError::BadQuery(_bad_query) => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::IoError(_io_error) => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::ProtocolError(_str) => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::InvalidMessage(_string) => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::TimeoutError => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::TooManyOrphanedStreamIds(_) => CassError::CASS_ERROR_LAST_ENTRY,
-            QueryError::UnableToAllocStreamId => CassError::CASS_ERROR_LAST_ENTRY,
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -43,6 +42,7 @@ impl From<&DbError> for CassError {
             DbError::Other(num) => {
                 CassError((CassErrorSource::CASS_ERROR_SOURCE_SERVER.0 << 24) | *num as u32)
             }
+            DbError::RateLimitReached { .. } => CassError::CASS_ERROR_SERVER_UNAVAILABLE,
```
Lorak-mmk, line 46: Add a TODO to change this to appropriate error in the future

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -247,23 +260,28 @@ pub unsafe extern "C" fn cass_cluster_set_load_balance_dc_aware_n(
...
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
-    // FIXME: used_hosts_per_remote_dc, allow_remote_dcs_for_local_cl ignored
-    // as there is no equivalent configuration in Rust Driver.
-    // TODO: string error handling
+    if used_hosts_per_remote_dc != 0 {
+        // TODO: Add warning that the parameter is deprecated and not supported in the driver.
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    }
```
Lorak-mmk, line 273: Is this parameter deprecated because Rust driver doesn't support it? It looks like it's still working in cpp driver.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cluster.rs
Gor027: Yep, the Rust driver does not support it, but the parameter Is also deprecated in the C++ driver.

```Rust
/*
* @deprecated The remote DC settings for DC-aware are not suitable for most
* scenarios that require DC failover. There is also unhandled gap between
* replication factor number of nodes failing and the full cluster failing. Only
* the remote DC settings are being deprecated.
* ...
* @param[in] used_hosts_per_remote_dc The number of hosts used in each remote
* DC if no hosts are available in the local dc (<b>deprecated</b>)
* ...
*/
```

Lorak-mmk (approved):

PR #89, opened 2022-11-23
Fix cargo clippy warnings after clippy update

Gor027:
## Pre-review checklist


constantfold:
Tests failed for some reason: https://github.com/scylladb/cpp-rust-driver/actions/runs/3535995948/jobs/5934618626

I restarted them (to confirm they are flaky), but we should analyze it further.

Lorak-mmk (approved):

PR #90, opened 2022-12-07
Fix setup step in Build and Cassandra workflows

Gor027:
The workflows fail for current pull requests:

```
Err:6 http://security.ubuntu.com/ubuntu xenial-security InRelease
The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 40976EAF437D05B5 NO_PUBKEY 3B4FE6ACC0B21F32
Reading package lists...
W: GPG error: http://security.ubuntu.com/ubuntu xenial-security InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 40976EAF437D05B5 NO_PUBKEY 3B4FE6ACC0B21F32
E: The repository 'http://security.ubuntu.com/ubuntu xenial-security InRelease' is not signed.
```

It seems that xenial-security does not maintain keys for the Ubuntu keyserver.
This PR will manually add the keys to the system.

## Pre-review checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,9 +22,7 @@ jobs:
 
     - name: Setup environment
       run: |
-        sudo sh -c "echo 'deb http://security.ubuntu.com/ubuntu xenial-security main' >> /etc/apt/sources.list"
-        sudo apt-get update
-        sudo apt-get install libssl1.0.0 libuv1-dev libkrb5-dev libc6-dbg
+        sudo apt-get install libssl1.1 libuv1-dev libkrb5-dev libc6-dbg
         sudo snap install valgrind --classic
         pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
         sudo sh -c "echo 2097152 >> /proc/sys/fs/aio-max-nr"
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -28,9 +28,7 @@ jobs:
 
       - name: Setup environment
         run: |
-          sudo sh -c "echo 'deb http://security.ubuntu.com/ubuntu xenial-security main' >> /etc/apt/sources.list"
-          sudo apt-get update
-          sudo apt-get install libssl1.0.0 libuv1-dev libkrb5-dev libc6-dbg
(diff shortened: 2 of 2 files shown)


Gor027:
@Lorak-mmk currently #78 is blocked by this PR.

mykaul:
Why aren't we using Ubuntu latest in our github actions? (actually, looks like we do, so I'm unsure why are we getting xenial stuff? Some backward compat?)

constantfold:
@mykaul We are using xenial repo for `libssl1.0.0` (but off top of my head I don't remember if it's for Rust's TLS or some "legacy" C++ Driver testing code - tests copied from original C++ Driver).

Gor027:
> Why aren't we using Ubuntu latest in our github actions? (actually, looks like we do, so I'm unsure why are we getting xenial stuff? Some backward compat?)

The tests require `libssl1.0.0` which was dropped from the default Ubuntu repos after Ubuntu 18.04, that is why `xenial-security` was added. Maybe `libssl1.1` can be installed instead, which will allow to remove `xenial-security` from the setup step.

Gor027:
I can confirm that the tests successfully pass with `libssl1.1`, so we can get rid of `xenial-security`.

Lorak-mmk (approved):

mykaul:
> I can confirm that the tests successfully pass with `libssl1.1`, so we can get rid of `xenial-security`.

Good - thanks. 1.0 is not maintained even anymore (I think? https://www.openssl.org/policies/releasestrat.html )
I guess at some point we'll need to move to OpenSSL 3 (is that libssl3 in Ubuntu?), but I assume there's somewhere in the Rust code some work needed to support it?

piodul:
> I guess at some point we'll need to move to OpenSSL 3 (is that libssl3 in Ubuntu?), but I assume there's somewhere in the Rust code some work needed to support it?

The Rust driver should support it out of the box. Documentation of the crate that we use to bind to OpenSSL states that  "OpenSSL versions 1.0.1 through 3.x.x and LibreSSL versions 2.5 through 3.4.1 are supported."

https://docs.rs/openssl/latest/openssl/ , first paragraph

PR #95, opened 2022-12-22
Fix memory leak error afer adding schema metadata

Gor027:
`Build` and `Cassandra` workflows did not report definitely lost memory leaks as errors and support for fetching schema metadata was merged with memory leaks.

Fixed memory leaking error and added appropriate options to valgrind to report only definitely lost memory as error.

## Pre-review checklist


Gor027:
@Lorak-mmk While debugging and testing #76 to prevent memory leaks I found that the schema metadata related PR #72 was merged still having some memory leaks that the GitHub Actions workflows did not detect and report. I hope this PR will prevent such cases in the future.

Lorak-mmk (approved):

PR #86, opened 2022-10-13
Add fetching tracing ID associated with a request

Gor027:
Added implementation for `cass_future_tracing_id`

Enabled `TracingTests` in GitHub actions for ScyllaDB and Cassandra

## Pre-review checklist

- [x] I have split my patch into logically separate commits.
- [x] All commit messages clearly explain what they change and why.
- [x] PR description sums up the changes and reasons why they should be introduced.
- [x] I have enabled appropriate tests in `.github/workflows/build.yml` in `gtest_filter`.
- [x] I have enabled appropriate tests in `.github/workflows/cassandra.yml` in `gtest_filter`.

PR #96, opened 2023-01-04
Update README about fetching tracing id

Gor027:
## Pre-review checklist

After merging fetching tracing id associated with a request (see https://github.com/scylladb/cpp-rust-driver/pull/86)
`cass_future_tracing_id` should be removed from README as it is already
implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,6 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
             <td rowspan="3">Unimplemented</td>
         </tr>
         <tr>
-            <td>cass_future_tracing_id</td>
-        </tr>
-        <tr>
             <td>cass_future_coordinator</td>
         </tr>
         <tr>


README.md
```diff
@@ -310,4 +307,4 @@ ___
 # License
 ___
 
-This project is licensed under the GNU Lesser General Public License v2.1
\ No newline at end of file
+This project is licensed under the GNU Lesser General Public License v2.1
```
constantfold: Nit: this shouldn't be a part of patch.

constantfold (approved):
LGTM, but I'd appreciate adding more details to the commit message (add which commit added the support for cass_future_tracing_id).

README.md
Gor027: Actually, the patch just adds a new line at the end of the line, should we add it in a separate PR?

constantfold: I wouldn't bother changing it at all.

PR #97, opened 2023-01-09
Avoid creating references to uninitialized data

Lorak-mmk:
Creating a reference to uninitialized data is most of the time undefined behaviour in Rust - even if data under this reference is never read.

This commit removes all occurences of this problem (in accordance with cpp-driver's semantics), instead using `std::ptr::write` - safer and more explicit way to achieve the same goal.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -170,8 +170,6 @@ pub unsafe extern "C" fn cass_future_error_message(
     message: *mut *const ::std::os::raw::c_char,
     message_length: *mut size_t,
 ) {
-    let message = ptr_to_ref_mut(message);
-    let message_length = ptr_to_ref_mut(message_length);
     ptr_to_ref(future).with_waited_state(|state: &mut CassFutureState| {
         let value = &state.value;
         let msg = state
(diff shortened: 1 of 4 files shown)


Gor027:
I think it is not safe to just remove `ptr_to_ref_mut` and leave `std::ptr::write`, cause `std::ptr::write` anticipates a non-null pointer and it is not obvious from the documentation what will happen if it gets a null pointer as a destination location. The `ptr_to_ref_mut` did the check previously and returned `None` so it would just panic on an unwrap if a null pointer was provided. So, perhaps it is a good idea to add those null checks before `std::ptr::write` and panic in case the provided pointer is null.

Gor027 (commented):
I think it is not safe to just remove `ptr_to_ref_mut` and leave `std::ptr::write`, cause `std::ptr::write` anticipates a non-null pointer and it is not obvious from the documentation what will happen if it gets a null pointer as a destination location. The `ptr_to_ref_mut` did the check previously and returned `None` so it would just panic on an unwrap if a null pointer was provided. So, perhaps it is a good idea to add those null checks before `std::ptr::write` and panic in case the provided pointer is null.

Lorak-mmk:
> it is not obvious from the documentation what will happen if it gets a null pointer as a destination location

Documentation states it is undefined behavior. 

cppdriver also doesn't check for null outputs and would also exhibit UB in that case, so there is no regression on that front. I can add null checks, but it would be just a hardening measure, not required for corectness of this project.

Gor027 (approved):

constantfold (approved):

piodul (approved):

PR #98, opened 2023-01-09
Put content of CassDataType into UnsafeCell

Lorak-mmk:
Previously `CassDataType` was just an enum, held inside `Arc`. User was given a pointer to `CassDataType` using `Arc::as_ptr` or `Arc::into_ptr`. There are however some functions that mutate the data - and they were given the very same pointers.
Current code was most likely sound - but I'm not completely sure, Rust reference is very confusing in this aspect.
It was however very confusing - when a programmer reads or writes a function that that *mut CassDataType it is not obivious that this data lies inside Arc and so has shared ownership.

To make this more explicit this commit puts `CassDataType` inside UnsafeCell. Now each access needs to use `.get_unchecked()` and `.get_mut_unchecked()` methods and an unsafe block / function, so it will be easier to spot aliasing ^ mutability problems in the future.

In the future we can use `Arc::get_mut_unchecked()` for this purpose, but it's not yet stabilised.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -4,6 +4,7 @@ use crate::types::*;
 use scylla::batch::{BatchType, Consistency, SerialConsistency};
 use scylla::frame::response::result::ColumnType;
 use scylla::transport::topology::{CollectionType, CqlType, NativeType};
+use std::cell::UnsafeCell;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::os::raw::c_char;
@@ -89,7 +90,7 @@ impl Default for UDTDataType {
 }
 
 #[derive(Clone, Debug)]
-pub enum CassDataType {
+pub enum CassDataTypeInner {
     Value(CassValueType),
     UDT(UDTDataType),
     List(Option<Arc<CassDataType>>),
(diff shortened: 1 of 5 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -99,8 +100,37 @@ pub enum CassDataType {
...
+    }
+
+    pub const fn new(inner: CassDataTypeInner) -> CassDataType {
+        CassDataType(UnsafeCell::new(inner))
+    }
+
+    pub fn new_alloc(inner: CassDataTypeInner) -> Arc<CassDataType> {
+        Arc::new(CassDataType(UnsafeCell::new(inner)))
+    }
+}
```
piodul, line 130: This impl looks quite generic, i.e. nothing here seems to be related to `CassDataTypeInner`. Would it make sense to make a generic container out of `CassDataType`? `CassDataType` would be just a typedef for the container. Maybe it could be used for other types as well.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -99,6 +100,35 @@ pub enum CassDataType {
     Custom(String),
 }
 
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct CassDataType(UnsafeCell<CassDataTypeInner>);
+
+unsafe impl Sync for CassDataType {}
```
Gor027, line 107: I think it'd be better to have an explanatory comment about why `Sync` was not automatically implemented and which parameters in `CassDataTypeInner` were not `Sync`.

Lorak-mmk:
Closing for now - I want to include this in the whole safety refactor after it's finished.

PR #100, opened 2023-01-12
Increment Cassandra version in CI tests

Gor027:
The `TracingTests.Simple` test was flaky, supposedly because of
Cassandra version 3.0.8. A select all query from
`system_traces.sessions` sometimes returned an empty result, however,
a manual check showed that the table was not empty during the execution
of the query. The issue is fixed in 3.0.9 which waits for the tracing
events before returning response and queries at the same consistency level
client side.

Ref to the changelog: https://raw.githubusercontent.com/apache/cassandra/cassandra-3.0.9/CHANGES.txt

The version is incremented to 3.0.9 for now, as it allows to
avoid the issue and not introduce new tests which were skipped
previously because of the version.

## Pre-review checklist

### DIFF
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Build
         run: cmake -DCASS_BUILD_INTEGRATION_TESTS=ON . && make
 
-      - name: Run integration tests on Cassandra 3.0.8
+      - name: Run integration tests on Cassandra 3.0.9
         env:
           #        Ignored tests are added in the end, after the "-" sign.
           Tests: "ClusterTests.*\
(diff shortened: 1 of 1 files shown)


constantfold:
Please add in the commit message that `3.0.9` is fine, since it fixed some issue around tracing ID (https://raw.githubusercontent.com/apache/cassandra/cassandra-3.0.9/CHANGES.txt): "Wait for tracing events before returning response and query at same consistency level client side (CASSANDRA-11465)" and this was the reason for failing test. Refs: https://issues.apache.org/jira/browse/CASSANDRA-11465

constantfold:
Generally, it would be best to bump the version up to the latest Cassandra 3.x: 3.11.14.

Gor027:
> Generally, it would be best to bump the version up to the latest Cassandra 3.x: 3.11.14.

Sure, I guess it should be done in another PR, as it will introduce a lot of new tests which were skipped previously and may not be fixed easily because of some dependencies on the Rust driver.

constantfold:
OK, so merging the current version, but we should remember to bump the version to the latest Cassandra 3.x version in a follow up PR.

PR #88, opened 2022-11-22
Add RPM/DEB packaging

Gor027:
## Pre-review checklist

This PR adds GitHub Actions workflows to create rpm/deb packages on push or pull request events. The generated packages will be saved in artifacts if the workflow runs end successfully.

This is a draft PR as the packaging metadata is yet to be updated. 

- [x] I have split my patch into logically separate commits.
- [x] All commit messages clearly explain what they change and why.
- [x] PR description sums up the changes and reasons why they should be introduced.
- [ ] I have enabled appropriate tests in `.github/workflows/build.yml` in `gtest_filter`.
- [ ] I have enabled appropriate tests in `.github/workflows/cassandra.yml` in `gtest_filter`.

PR #101, opened 2023-01-17
Add install capabilities to top-level CMake config

constantfold:
Before this change, the generated Makefile by CMake lacked the capability to run `make install`. This PR makes it possible
to run `make install`, for example:

```bash
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=RELEASE -DCASS_INSTALL_PKG_CONFIG=OFF -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib64 ..
make DESTDIR=/tmp/example install
```

The names and contents of installed files try to mimic the original C++ Driver files. For example, installation of C++ Driver generates the following `.so` files/symlinks:

```
usr/lib64/libscylla-cpp-driver.so -> libscylla-cpp-driver.so.2
usr/lib64/libscylla-cpp-driver.so.2 -> libscylla-cpp-driver.so.2.16.2
usr/lib64/libscylla-cpp-driver.so.2.16.2
```

This particular behavior is replicated in a new `build-cmake-compat.sh` script, which is executed as a `BUILD_COMMAND` in the Rust external project in the top-level `CMakeLists.txt`.

Unfortunately, I have not been able to configure CMake to create such symlinks on its own. One of the reasons is an unfixed bug in CMake: https://gitlab.kitware.com/cmake/cmake/-/issues/23249#note_1277271. Therefore, the `build-cmake-compat.sh` is introduced to create the necessary files in a way it's easy to install for CMake.

The third commit includes the static library in the installed files. 

Those are the contents of `DESTDIR` (as in `make DESTDIR=... install`) in the original C++ driver:
```
DESTDIR
└── usr
    ├── include
    │   ├── cassandra.h
    │   └── dse.h
    └── lib64
        ├── libscylla-cpp-driver.so -> libscylla-cpp-driver.so.2
        ├── libscylla-cpp-driver.so.2 -> libscylla-cpp-driver.so.2.16.2
        ├── libscylla-cpp-driver.so.2.16.2
        └── libscylla-cpp-driver_static.a
```

and after this PR in the bindings:
```
DESTDIR
└── usr
    ├── include
    │   └── cassandra.h
    └── lib64
        ├── libscylla-cpp-driver.so -> libscylla-cpp-driver.so.0
        ├── libscylla-cpp-driver.so.0 -> libscylla-cpp-driver.so.0.0.1
        ├── libscylla-cpp-driver.so.0.0.1
        └── libscylla-cpp-driver_static.a
```

This PR is a prerequisite for a follow-up PR reworking #88.

## Pre-review checklist

- [X] I have split my patch into logically separate commits.
- [X] All commit messages clearly explain what they change and why.
- [X] PR description sums up the changes and reasons why they should be introduced.

### DIFF
--- a/scylla-rust-wrapper/build-cmake-compat.sh
+++ b/scylla-rust-wrapper/build-cmake-compat.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -e
+
+# The version number should be in the third line
+# of Cargo.toml file (e.g. version = "0.0.1")
+VERSION_NUMBER=$(sed -n '3p' < Cargo.toml)
+REGEX="version = \"([0-9]+)\.([0-9]+)\.([0-9]+)\""
+if [[ $VERSION_NUMBER =~ $REGEX ]]
+then
+  MAJOR=${BASH_REMATCH[1]}
+  MINOR=${BASH_REMATCH[2]}
+  PATCH=${BASH_REMATCH[3]}
+else
+  echo "Could not parse the version number in Cargo.toml"
+  echo "Tried to parse a third line of Cargo.toml: $VERSION_NUMBER"
+  echo 'but it did not contain a valid version number (e.g. version = "0.0.1").'
+  exit 1
+fi
+
+create_symlinks() {
+  rm -rf target/$1/cmake-compat/
+  mkdir target/$1/cmake-compat/
+
+  cp target/$1/libscylla_cpp_driver.so target/$1/cmake-compat/libscylla-cpp-driver.so.$MAJOR.$MINOR.$PATCH
+  ln -s libscylla-cpp-driver.so.$MAJOR.$MINOR.$PATCH target/$1/cmake-compat/libscylla-cpp-driver.so.$MAJOR
+  ln -s libscylla-cpp-driver.so.$MAJOR target/$1/cmake-compat/libscylla-cpp-driver.so
+}
+
+if [[ $1 == "DEBUG" ]]
+then
+  cargo build
+  create_symlinks "debug"
+elif [[ $1 == "RELEASE" ]]
+then
+  cargo build --release
+  create_symlinks "release"
+else
+  echo 'Invalid build mode: $1 (only DEBUG or RELEASE permitted)'
+  exit 1
+fi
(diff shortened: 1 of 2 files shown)


Gor027 (commented):
I manually enabled the flag `CASS_BUILD_STATIC` in `CMakeLists.txt`, however, after the build finished `.a` file was not created in the destination directory. I built it with the same commands as in the PR description. Should I build it in another way?

scylla-rust-wrapper/build-cmake-compat.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -e
+
+# The version number should be in the third line
+# of Cargo.toml file (e.g. version = "0.0.1")
+VERSION_NUMBER=$(sed -n '3p' < Cargo.toml)
```
wprzytula, line 6: Isn't this an assumption that is prone to change?

scylla-rust-wrapper/build-cmake-compat.sh
```diff
@@ -17,11 +17,31 @@ else
...
+fi
+
+
 create_symlinks() {
   rm -rf target/$1/cmake-compat/
   mkdir target/$1/cmake-compat/
 
   cp target/$1/libscylla_cpp_driver.so target/$1/cmake-compat/libscylla-cpp-driver.so.$MAJOR.$MINOR.$PATCH
+  
+  if [[ "$COPY_STATIC" == true ]]
```
wprzytula, line 40: Is comparison to `true` necessary in Bash??

wprzytula (approved):
I've tried it out and the results were very good. The `-DCASS_BUILD_STATIC` option worked as well, respecting correctly options: `=false`, `=OFF` as non-static build and `=true`, `=ON` as static. The `.a` file was present in the target dir.

roydahan:
@avelanarius is this one still needed?
If so, what's preventing it from being merged?

roydahan:
Closing in favor of #88

Lorak-mmk:
Did you mean #122 ?

roydahan:
Yes

PR #102, opened 2023-01-25
Upgrade cassandra version

Gor027:
## Pre-review checklist

### DIFF
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Build
         run: cmake -DCASS_BUILD_INTEGRATION_TESTS=ON . && make
 
-      - name: Run integration tests on Cassandra 3.0.9
+      - name: Run integration tests on Cassandra 4.0.7
         env:
           #        Ignored tests are added in the end, after the "-" sign.
           Tests: "ClusterTests.*\
@@ -51,12 +51,13 @@ jobs:
 :CompressionTests.*\
 :LoggingTests.*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
+:PreparedTests.Integration_Cassandra_FailFastWhenPreparedIDChangesDuringReprepare\
 :*5.Integration_Cassandra_*\
 :*19.Integration_Cassandra_*\
-:CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_UDT\
+:*7.Integration_Cassandra_*\
 :SslTests.Integration_Cassandra_ReconnectAfterClusterCrashAndRestart\
 :ExecutionProfileTest.InvalidName"
-        run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --version=3.0.9 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
+        run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --version=4.0.7 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
 
       - name: Upload test logs
         uses: actions/upload-artifact@v3
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -12,9 +12,11 @@ use scylla::frame::value::MaybeUnset::{Set, Unset};
 use scylla::query::Query;
 use scylla::statement::prepared_statement::PreparedStatement;
 use scylla::statement::SerialConsistency;
-use scylla::Bytes;
+use scylla::{BufMut, Bytes, BytesMut};
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::os::raw::{c_char, c_int};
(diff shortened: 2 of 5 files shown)


scylla-rust-wrapper/src/statement.rs
```diff
@@ -221,6 +223,30 @@ pub unsafe extern "C" fn cass_statement_set_paging_state(
...
+        statement_from_raw.paging_state = None;
+        return CassError::CASS_ERROR_LIB_NULL_VALUE;
+    }
+
+    let paging_state_usize: usize = paging_state_size.try_into().unwrap();
+    let mut b = BytesMut::with_capacity(paging_state_usize);
+    let paging_state_bytes = slice::from_raw_parts(paging_state, paging_state_usize);
+    for byte in paging_state_bytes {
+        b.put_i8(*byte);
+    }
```
piodul, line 247: Would `b.put(&paging_state_bytes)` work here? Documentation for `BytesMut` suggests that it is possible:

https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.with_capacity

Gor027: I believe it should work, will fix it. Indeed, it will be better to put it without specifying whether it is `i8` or `u8`, as `c_char` is considered to be `u8` on some processors (e.g. ARM) and it may cause a compilation error.

tests/src/integration/tests/test_tracing.cpp
```diff
@@ -34,6 +34,7 @@ CASSANDRA_INTEGRATION_TEST_F(TracingTests, Simple) {
   { // Validate tracing ID by attempting to get the associated tracing data.
     Statement statement("SELECT * FROM system_traces.sessions WHERE session_id = ?", 1);
     statement.bind(0, tracing_id);
+    msleep(10000);
```
Lorak-mmk: I don't like this 10s sleep, and in general we should avoid timing-dependent tests.
If there is no way to solve this in a different way (how does Cassandra 4x solve it? Did they remove this test? Modify it somehow?), we should execute this block in a loop (and this 10s would be a time limit instead of unconditional sleep) until it succeeds or time limit is reached.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/statement.rs
Gor027: Unfortunately, `b.put(&paging_state_bytes)` is not implemented for `&[u8]` or `&[i8]`, so I think either we should add an explicit cast to `i8` and ignore the warnings or use the `cfg!` macro with some configuration flags to put `c_char` as `i8` or `u8` depending on the processor type.

tests/src/integration/tests/test_tracing.cpp
Gor027: I have changed the unconditional sleep to a loop with a 10s limit after which it can fail. The mentioned issue in the commit message https://issues.apache.org/jira/browse/CASSANDRA-11465 is a fix for versions up to version 3.8. I think the original cpp-driver is not being tested on newer Cassandra versions 4.0 or 4.1, so perhaps we should wait for a fix to be added for those versions too.

mykaul:
By now 4.1.2 is already available. What's the plan to get this in?

Gor027:
> By now 4.1.2 is already available. What's the plan to get this in?

Upgrading to 4.1.2 will require more changes related to the recent renaming of the parameters in `cassandra.yaml`:
https://cassandra.apache.org/doc/latest/cassandra/configuration/configuration.html
I think it is better to do that in a separate PR.

Gor027:
This will also require some changes in `scylla-ccm` also, to generate proper configuration before starting a cluster.

scylla-rust-wrapper/src/statement.rs
wprzytula: @piodul WDYT?

.github/workflows/cassandra.yml
```diff
@@ -51,12 +51,13 @@ jobs:
 :CompressionTests.*\
 :LoggingTests.*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
+:PreparedTests.Integration_Cassandra_FailFastWhenPreparedIDChangesDuringReprepare\
 :*5.Integration_Cassandra_*\
 :*19.Integration_Cassandra_*\
-:CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_UDT\
+:*7.Integration_Cassandra_*\
 :SslTests.Integration_Cassandra_ReconnectAfterClusterCrashAndRestart\
```
wprzytula, line 58: Ad commit message: 

> * CassandraTypes tests for Duration type, as binding duration to
  statement is not yet supported. Iterator based deserialization in the
  Rust driver will allow to add this feature.

Is this feature added in #112?

scylla-rust-wrapper/src/statement.rs
piodul: Let's keep the current version for simplicity and hope that the compiler optimizes it.

.github/workflows/cassandra.yml
Gor027: Yes.

wprzytula (approved):

PR #103, opened 2023-02-13
Update rust driver rev in Cargo.toml

Gor027:
Imported scylla-rust-driver by commit hash 9dc6f4871f2. This will include all pre-execution-profile changes in the Rust driver and will resolve the following conflicts:

* Removed `BadQuery::ValueLenMismatch` error, as it was also removed from the Rust driver to allow removing `len` from `BatchValues`.

* Resolved conflicts after custom authentication feature was added to the Rust driver.

It was requested by @wprzytula and it aims to make the addition of execution profiles easier.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -39,6 +39,8 @@ pub struct CassCluster {
     child_load_balancing_policy: CassClusterChildLoadBalancingPolicy,
     token_aware_policy_enabled: bool,
     use_beta_protocol_version: bool,
+    auth_username: Option<String>,
+    auth_password: Option<String>,
 }
 
 pub struct CassCustomPayload;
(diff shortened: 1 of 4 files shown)


Gor027:
This PR is dependent upon the fix https://github.com/scylladb/cpp-rust-driver/pull/104 to pass the CI successfully.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -75,6 +77,15 @@ pub fn build_session_builder(cluster: &CassCluster) -> SessionBuilder {
...
+            .session_builder
+            .clone()
+            .known_nodes(&known_nodes)
+            .load_balancing(load_balancing)
+            .user(username, password);
+    }
+
     cluster
         .session_builder
         .clone()
```
wprzytula: This code looks better without repetition and in a more functional style, like this:
```rust
let builder = cluster
  .session_builder
  .clone()
  .known_nodes(&known_nodes)
  .load_balancing(load_balancing);
if let (Some(username), Some(password)) = (&cluster.auth_username, &cluster.auth_password) {
  builder.user(username, password)
} else {
  builder
}
```

Gor027: Done.

wprzytula (approved):

PR #106, opened 2023-03-23
CurrentThread runtime support - fix

wprzytula:
So far, the implementation of CassFuture caused deadlock when the only thread called `cass_future_wait()`. This happened because the thread went asleep on a Condvar until the future is resolved, but there were no
other executors that would resolve the future.
Although a runtime with no worker threads is not recommended for production usage, it is helpful e.g. for debugging (as it makes execution serialised). Therefore, we fix the problem by changing the way the thread waits for the future's result: instead of sleeping while the future is not yet resolved, it calls runtime.block_on(fut), which makes
it the executor of that future.
This not only fixes the problem, but possibly boosts efficiency.
More specifically, it saves a sleep and a wakeup in the case of having a non-empty worker thread pool (the main thread does not wake a worker, falls asleep until the worker thread resolves the future and gets woken up in the end, but instead simply does the work by itself).

To test that the changes did solve the problem, change `RUNTIME` in `lib.rs` into a "current thread" variant:
from: 
```rust
```
into:
```rust
pub static ref RUNTIME: Runtime =
        tokio::runtime::Builder::new_current_thread()
                .enable_all()
                .build()
                .unwrap();
```

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -75,12 +75,12 @@ impl CassFuture {
         let cass_fut_clone = cass_fut.clone();
         let join_handle = RUNTIME.spawn(async move {
             let r = fut.await;
-            let mut lock = cass_fut_clone.state.lock().unwrap();
-            lock.value = Some(r);
-
-            // Take the callback and call it after realeasing the lock
-            let maybe_cb = lock.callback.take();
-            mem::drop(lock);
+            let maybe_cb = {
+                let mut guard = cass_fut_clone.state.lock().unwrap();
+                guard.value = Some(r);
+                // Take the callback and call it after releasing the lock
+                guard.callback.take()
+            };
             if let Some(bound_cb) = maybe_cb {
                 bound_cb.invoke(cass_fut_clone.as_ref());
             }
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -132,6 +132,9 @@ impl CassFuture {
     }
 }
 
+trait CheckSendSync: Send + Sync {}
+impl CheckSendSync for CassFuture {}
+
 #[no_mangle]
 pub unsafe extern "C" fn cass_future_set_callback(
     future_raw: *const CassFuture,
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -58,8 +58,8 @@ pub struct CassFuture {
 impl CassFuture {
     pub fn make_raw(
         fut: impl Future<Output = CassFutureResult> + Send + 'static,
-    ) -> *const CassFuture {
-        Self::new_from_future(fut).into_raw()
+    ) -> *mut CassFuture {
+        Self::new_from_future(fut).into_raw() as *mut _
     }
 
     pub fn new_from_future(
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -249,3 +249,44 @@ pub unsafe extern "C" fn cass_future_tracing_id(
         _ => CassError::CASS_ERROR_LIB_INVALID_FUTURE_TYPE,
     })
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{os::raw::c_char, thread, time::Duration};
+
+    // This is not a particularly smart test, but if some thread is granted access the value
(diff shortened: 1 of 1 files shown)


scylla-rust-wrapper/src/future.rs
```diff
@@ -102,11 +107,11 @@ impl CassFuture {
...
-        let mut guard = self
-            .wait_for_value
-            .wait_while(self.state.lock().unwrap(), |s| s.value.is_none())
-            .unwrap();
-        f(&mut guard)
+        let handle = self.state.lock().unwrap().join_handle.take();
+        if let Some(handle) = handle {
+            RUNTIME.block_on(handle).unwrap();
+        }
+        f(&mut self.state.lock().unwrap())
```
piodul: It looks like we are taking the lock twice here, but in the case when the future was already waited on (when e.g. calling `cass_future_error_code`) we could get away with just one lock acquisition.

wprzytula: I can't think of any simple solutions for this. Could you share yours?

piodul: For example:

```rust
    pub(self) fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
        let mut lock = self.state.lock().unwrap();
        let handle = lock.join_handle.take();
        if let Some(handle) = handle {
            std::mem::drop(lock);
            RUNTIME.block_on(handle).unwrap();
            lock = self.state.lock().unwrap();
        }
        f(&mut lock)
    }
```

Lorak-mmk: Isn't there a race condition, in both @piodul suggestion and original PR changes? 
We have 2 threads.
1. Thread 1 performs `with_waited_state`, `.take()`'s the handle, drops a lock.
2. Before Thread 1 can perform `block_on`, Thread 2 enters `with_waited_state`, takes a lock.
3. There is no `join_handle` (as it was already taken by Thread 1), so it moves into calling the closure (before @piodul 's changes this requires taking a lock again, but it doesn't matter).
4. Closure is called before future is resolved.

Lorak-mmk: And we do need to handle such scenario:
1. `CassFuture` is documented as being thread safe: https://cpp-driver.docs.scylladb.com/master/topics/#thread-safety
2. cpp-driver handles such scenario correctly (as far as I can tell by looking at the code).
3. Our current implementation also handles it correctly

piodul: Ouch, you are definitely right. Thread 2 needs to wait for Thread 1 to complete before it can continue.

I think we still need a condvar. If thread 2 notices that both `join_handle` and `value` are `None`, then it should wait on the condvar until `value` is not `None`, just like before.

wprzytula:
v2: I've fixed the race condition and hardened CassFuture a bit:
- added a compile-time assertion of `Send + Sync`
- written a smoke test of thread safety

scylla-rust-wrapper/src/future.rs
```diff
@@ -249,3 +249,43 @@ pub unsafe extern "C" fn cass_future_tracing_id(
...
+                cass_future_error_message(cass_fut, &mut message, &mut msg_len);
+                assert_eq!(ptr_to_cstr_n(message, msg_len), Some(ERROR_MSG));
+            });
+
+            let mut message: *const c_char = std::ptr::null();
+            let mut msg_len: size_t = 0;
+            cass_future_error_message(cass_fut, &mut message, &mut msg_len);
+            assert_eq!(ptr_to_cstr_n(message, msg_len), Some(ERROR_MSG));
+
+            cass_future_free(cass_fut);
```
Lorak-mmk, line 304: Shouldn't you wait for thread to finish before freeing future?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/future.rs
```diff
@@ -58,8 +58,8 @@ pub struct CassFuture {
 impl CassFuture {
     pub fn make_raw(
         fut: impl Future<Output = CassFutureResult> + Send + 'static,
-    ) -> *const CassFuture {
-        Self::new_from_future(fut).into_raw()
+    ) -> *mut CassFuture {
+        Self::new_from_future(fut).into_raw() as *mut _
```
Gor027, line 65: The changes are made to the `CassFuture::make_raw` function, but the commit message includes `make_mut()`.

Lorak-mmk (approved):

PR #107, opened 2023-03-24
Execution profiles support

### LINKED ISSUE #38: CassExecProfile
Useful feature of cppdriver that lets user create predefined set of options that can be then used with queries.

Rust driver doesn't support them - we need to decide if we want to implement it there, or fully in our wrapper.
I'd be leaning towards implementation in the driver - the feature is useful in cppdriver, and could be useful in Rust driver, and could bring performance benefits.

Functions:

- [x] `cass_execution_profile_new`
- [x] `cass_execution_profile_free`
- [x] `cass_execution_profile_set_request_timeout`
- [x] `cass_execution_profile_set_consistency`
- [x] `cass_execution_profile_set_serial_consistency`
- [x] `cass_execution_profile_set_load_balance_round_robin`
- [x] `cass_execution_profile_set_load_balance_dc_aware`
- [x] `cass_execution_profile_set_load_balance_dc_aware_n`
- [x] `cass_execution_profile_set_token_aware_routing`
- [ ] `cass_execution_profile_set_token_aware_routing_shuffle_replicas`
- [x] `cass_execution_profile_set_latency_aware_routing`
- [x] `cass_execution_profile_set_latency_aware_routing_settings`
- [ ] `cass_execution_profile_set_whitelist_filtering`
- [ ] `cass_execution_profile_set_whitelist_filtering_n`
- [ ] `cass_execution_profile_set_blacklist_filtering`
- [ ] `cass_execution_profile_set_blacklist_filtering_n`
- [ ] `cass_execution_profile_set_whitelist_dc_filtering`
- [ ] `cass_execution_profile_set_whitelist_dc_filtering_n`
- [ ] `cass_execution_profile_set_blacklist_dc_filtering`
- [ ] `cass_execution_profile_set_blacklist_dc_filtering_n`
- [x] `cass_execution_profile_set_retry_policy`
- [x] `cass_execution_profile_set_constant_speculative_execution_policy`
- [ ] `cass_execution_profile_set_no_speculative_execution_policy`
- [x] `cass_batch_set_execution_profile`
- [x] `cass_batch_set_execution_profile_n`
- [x] `cass_statement_set_execution_profile`
- [x] `cass_statement_set_execution_profile_n`
- [x] `cass_cluster_set_execution_profile`
- [x] `cass_cluster_set_execution_profile_n`

wprzytula:
The Rust driver has had recently added support for Execution Profiles. As that concept is present in the CPP driver, this PR implements the bridge between them (as there are, of course, some differences to be handled).
As a side effect, this PR raises the supported Rust driver version to a recent one, featuring a small UDT refactor, a major load balancing refactor and support for Serverless Cloud connections. 

This PR contains a considerable number of tests, including unit tests and integration tests (to which end proxy is used). The existing integration tests inherited from CPP driver are (for various reasons) not suitable for use in bindings (see listing below).

Existing integration tests:
- `ExecutionProfileTest.Integration_Cassandra_*`
  These tests assume that it is possible to set Consistency as SerialConsistency and vice versa on an execution profile and that such error will be reported only when querying with that profile. Our implementation reports error already on setting bad LegacyConsistency variant on a profile, so in order to run any of those tests, it is required to swap those consistencies (in the source code) not to throw exceptions while preparing for tests. 
  - `InvalidName` - valid, works (after the swap mentioned above).
  - `RequestTimeout` - fails due to `tokio` being unable to handle such small delays in sleeps.
  - `Consistency` and `SerialConsistency` - fails to the mentioned above different error semantics (expects error on query, not on exec profile definition).
  - `RoundRobin` - fails due to unimplemented `get_host_from_future()` (and we aren't able to implement it in Rust driver architecture).
  - `LatencyAwareRouting`, `RetryPolicy` - fail due to nonsupported logging (at least in my case, I was unable to turn on tracing and see any traces coming from the Rust driver).
  - `SpeculativeExecutionPolicy` - fails due to unimplemented `set_record_attempted_hosts()`.
  - `BlacklistFiltering`, `WhitelistFiltering` - nonsupported in Rust driver.
- `LatencyAwarePolicyTest.Integration_Cassandra_IsEnabled` - fails due to nonsupported logging.

Fixes: #38 

### DIFF
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -9,9 +9,8 @@ keywords = ["database", "scylla", "cql", "cassandra"]
 categories = ["database"]
 license = "MIT OR Apache-2.0"
 
-
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "9dc6f4871f2477c8ebb5bd911a0ce8ddc05ba8db", features = ["ssl"]}
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "1a913a19", features = ["ssl"]}
 tokio = { version = "1.1.0", features = ["full"] }
 lazy_static = "1.4.0"
 uuid = "1.1.2"
@@ -23,11 +22,16 @@ libc = "0.2.108"
 openssl-sys = "0.9.75"
 openssl = "0.10.32"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
-tracing = "0.1.25"
+tracing = "0.1.37"
 
 [build-dependencies]
 bindgen = "0.65"
 chrono = "0.4.20"
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+ntest = "0.9.0"
+scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "1a913a19"}
 
 [lib]
 name = "scylla_cpp_driver"
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -26,7 +26,7 @@ tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
 tracing = "0.1.25"
 
 [build-dependencies]
-bindgen = "0.59.1"
+bindgen = "0.65"
 chrono = "0.4.20"
 
 [lib]
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -31,6 +31,7 @@ chrono = "0.4.20"
 [dev-dependencies]
 assert_matches = "1.5.0"
 ntest = "0.9.0"
+rusty-fork = "0.3.0"
 scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "1a913a19"}
 
 [lib]
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -73,3 +73,24 @@ pub unsafe fn strlen(ptr: *const c_char) -> size_t {
     }
     libc::strlen(ptr) as size_t
 }
+
+#[cfg(test)]
+pub fn str_to_c_str_n(s: &str) -> (*const c_char, size_t) {
+    let mut c_str = std::ptr::null();
+    let mut c_strlen = size_t::default();
+
+    // SAFETY: The pointers that are passed to `write_str_to_c` are compile-checked references.
+    unsafe { write_str_to_c(s, &mut c_str, &mut c_strlen) };
+
+    (c_str, c_strlen)
+}
+
+#[cfg(test)]
+macro_rules! make_c_str {
+    ($str:literal) => {
+        concat!($str, "\0").as_ptr() as *const c_char
+    };
+}
+
+#[cfg(test)]
+pub(crate) use make_c_str;
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -2,6 +2,7 @@ use crate::argconv::{free_boxed, ptr_to_ref, ptr_to_ref_mut};
 use crate::cass_error::CassError;
 use crate::cass_types::CassConsistency;
 use crate::cass_types::{make_batch_type, CassBatchType};
+use crate::exec_profile::PerStatementExecProfile;
 use crate::retry_policy::CassRetryPolicy;
 use crate::statement::{CassStatement, Statement};
 use crate::types::*;
@@ -14,6 +15,8 @@ use std::sync::Arc;
 pub struct CassBatch {
     pub state: Arc<CassBatchState>,
     pub batch_request_timeout_ms: Option<cass_uint64_t>,
+
+    pub(crate) exec_profile: Option<PerStatementExecProfile>,
 }
 
 #[derive(Clone)]
@@ -31,6 +34,7 @@ pub unsafe extern "C" fn cass_batch_new(type_: CassBatchType) -> *mut CassBatch 
                 bound_values: Vec::new(),
             }),
             batch_request_timeout_ms: None,
+            exec_profile: None,
         }))
     } else {
         std::ptr::null_mut()
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -2,6 +2,7 @@ use crate::argconv::{free_boxed, ptr_to_ref, ptr_to_ref_mut};
 use crate::cass_error::CassError;
 use crate::cass_types::CassConsistency;
 use crate::cass_types::{make_batch_type, CassBatchType};
+use crate::retry_policy::CassRetryPolicy;
 use crate::statement::{CassStatement, Statement};
 use crate::types::*;
 use scylla::batch::Batch;
@@ -71,6 +72,29 @@ pub unsafe extern "C" fn cass_batch_set_serial_consistency(
     Arc::make_mut(&mut batch.state)
         .batch
         .set_serial_consistency(Some(serial_consistency));
+
+    CassError::CASS_OK
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_batch_set_retry_policy(
+    batch: *mut CassBatch,
+    retry_policy: *const CassRetryPolicy,
+) -> CassError {
+    let batch = ptr_to_ref_mut(batch);
+
+    let maybe_arced_retry_policy: Option<Arc<dyn scylla::retry_policy::RetryPolicy>> =
+        retry_policy.as_ref().map(|policy| match policy {
+            CassRetryPolicy::DefaultRetryPolicy(default) => {
+                default.clone() as Arc<dyn scylla::retry_policy::RetryPolicy>
+            }
+            CassRetryPolicy::FallthroughRetryPolicy(fallthrough) => fallthrough.clone(),
+            CassRetryPolicy::DowngradingConsistencyRetryPolicy(downgrading) => downgrading.clone(),
+        });
+
+    Arc::make_mut(&mut batch.state)
+        .batch
+        .set_retry_policy(maybe_arced_retry_policy);
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -592,12 +592,58 @@ pub unsafe extern "C" fn cass_cluster_set_serial_consistency(
     CassError::CASS_OK
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_execution_profile(
+    cluster: *mut CassCluster,
+    name: *const c_char,
+    profile: *const CassExecProfile,
+) -> CassError {
+    cass_cluster_set_execution_profile_n(cluster, name, strlen(name), profile)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_execution_profile_n(
+    cluster: *mut CassCluster,
+    name: *const c_char,
+    name_length: size_t,
+    profile: *const CassExecProfile,
+) -> CassError {
+    let cluster = ptr_to_ref_mut(cluster);
+    let name = if let Some(name) =
+        ptr_to_cstr_n(name, name_length).and_then(|name| name.to_owned().try_into().ok())
+    {
+        name
+    } else {
+        // Got NULL or empty string, which is invalid name for a profile.
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+    let profile = if let Some(profile) = profile.as_ref() {
+        profile.clone()
+    } else {
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
+    cluster.execution_profile_map.insert(name, profile);
+
+    CassError::CASS_OK
+}
+
 #[cfg(test)]
 mod tests {
     use crate::testing::assert_cass_error_eq;
 
     use super::*;
+    use crate::{
+        argconv::make_c_str,
+        cass_error::CassError,
+        exec_profile::{cass_execution_profile_free, cass_execution_profile_new},
+    };
     use assert_matches::assert_matches;
+    use std::{
+        collections::HashSet,
+        convert::{TryFrom, TryInto},
+        os::raw::c_char,
+    };
 
     #[test]
     fn test_load_balancing_config() {
@@ -663,6 +709,118 @@ mod tests {
                 }
             }
 
+            cass_cluster_free(cluster_raw);
+        }
+    }
+
+    #[test]
+    fn test_register_exec_profile() {
+        unsafe {
+            let cluster_raw = cass_cluster_new();
+            let exec_profile_raw = cass_execution_profile_new();
+            {
+                /* Test valid configurations */
+                let cluster = ptr_to_ref(cluster_raw);
+                {
+                    assert!(cluster.execution_profile_map.is_empty());
+                }
+                {
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            make_c_str!("profile1"),
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_OK
+                    );
+                    assert!(cluster.execution_profile_map.len() == 1);
+                    let _profile = cluster
+                        .execution_profile_map
+                        .get(&"profile1".to_owned().try_into().unwrap())
+                        .unwrap();
+
+                    let (c_str, c_strlen) = str_to_c_str_n("profile1");
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile_n(
+                            cluster_raw,
+                            c_str,
+                            c_strlen,
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_OK
+                    );
+                    assert!(cluster.execution_profile_map.len() == 1);
+                    let _profile = cluster
+                        .execution_profile_map
+                        .get(&"profile1".to_owned().try_into().unwrap())
+                        .unwrap();
+
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            make_c_str!("profile2"),
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_OK
+                    );
+                    assert!(cluster.execution_profile_map.len() == 2);
+                    let _profile = cluster
+                        .execution_profile_map
+                        .get(&"profile2".to_owned().try_into().unwrap())
+                        .unwrap();
+                }
+
+                /* Test invalid configurations */
+                {
+                    // NULL name
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            std::ptr::null(),
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
+                }
+                {
+                    // empty name
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            make_c_str!(""),
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
+                }
+                {
+                    // NULL profile
+                    assert_cass_error_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            make_c_str!("profile1"),
+                            std::ptr::null()
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
+                }
+                // Make sure that invalid configuration did not influence the profile map
+                assert_eq!(
+                    cluster
+                        .execution_profile_map
+                        .keys()
+                        .cloned()
+                        .collect::<HashSet<_>>(),
+                    vec![
+                        ExecProfileName::try_from("profile1".to_owned()).unwrap(),
+                        "profile2".to_owned().try_into().unwrap()
+                    ]
+                    .into_iter()
+                    .collect::<HashSet<ExecProfileName>>()
+                );
+            }
+
+            cass_execution_profile_free(exec_profile_raw);
             cass_cluster_free(cluster_raw);
         }
     }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -10,6 +10,7 @@ use openssl::ssl::SslContextBuilder;
 use openssl_sys::SSL_CTX_up_ref;
 use scylla::execution_profile::ExecutionProfileBuilder;
 use scylla::frame::Compression;
+use scylla::load_balancing::LatencyAwarenessBuilder;
 use scylla::load_balancing::{DefaultPolicyBuilder, LoadBalancingPolicy};
 use scylla::retry_policy::RetryPolicy;
 use scylla::speculative_execution::SimpleSpeculativeExecutionPolicy;
@@ -25,6 +26,8 @@ include!(concat!(env!("OUT_DIR"), "/cppdriver_compression_types.rs"));
 pub(crate) struct LoadBalancingConfig {
     pub(crate) token_awareness_enabled: bool,
     pub(crate) dc_awareness: Option<DcAwareness>,
+    pub(crate) latency_awareness_enabled: bool,
+    pub(crate) latency_awareness_builder: LatencyAwarenessBuilder,
 }
 impl LoadBalancingConfig {
     // This is `async` to prevent running this function from beyond tokio context,
@@ -36,6 +39,9 @@ impl LoadBalancingConfig {
                 .prefer_datacenter(dc_awareness.local_dc.clone())
                 .permit_dc_failover(true)
         }
+        if self.latency_awareness_enabled {
+            builder = builder.latency_awareness(self.latency_awareness_builder);
+        }
         builder.build()
     }
 }
@@ -44,6 +50,8 @@ impl Default for LoadBalancingConfig {
         Self {
             token_awareness_enabled: true,
             dc_awareness: None,
+            latency_awareness_enabled: false,
+            latency_awareness_builder: Default::default(),
         }
     }
 }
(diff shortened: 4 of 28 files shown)


scylla-rust-wrapper/src/statement.rs
```diff
@@ -146,7 +146,10 @@ pub unsafe extern "C" fn cass_statement_new_n(
 
     // Set Cpp Driver default configuration for queries:
     query.disable_paging();
-    query.set_consistency(Consistency::One);
+    // TODO: Consider moving this configuration to the default execution profile.
+    // With the current way, any changes to consistency on exec profile level will be
+    // anyway overriden with this setting.
+    query.set_consistency(Consistency::LocalOne);
```
Gor027: > Contrary to what the commit eb6b70d138752c5e6a487eeaf69e6baa4d6652a4
says

I think the hash in the commit message is invalid, perhaps you meant to add this (?):
https://github.com/scylladb/cpp-rust-driver/commit/fcedb509

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -73,3 +73,24 @@ pub unsafe fn strlen(ptr: *const c_char) -> size_t {
...
+    // SAFETY: The pointers that are passed to `write_str_to_c` are compile-checked references.
+    unsafe { write_str_to_c(s, &mut c_str, &mut c_strlen) };
+
+    (c_str, c_strlen)
+}
+
+#[cfg(test)]
+macro_rules! make_c_str {
+    ($str:literal) => {
+        concat!($str, "\0").as_ptr() as *const i8
```
Gor027: Why not use `c_char` instead of `i8`?

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -0,0 +1,153 @@
...
+    free_boxed(profile);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_blacklist_dc_filtering(
+    profile: *mut CassExecProfile,
+    dcs: *const c_char,
+) -> CassError {
+    unimplemented!()
+}
```
Gor027: As the Rust driver does not support the filtering policies, I think it is preferable to leave the `set_*_filtering(_n)` functions unimplemented which will cause a linkage error in the user's application noting that the feature is not supported, instead of having an empty implementation. If these functions are used in tests, they can be commented out with a message that the filtering policy is not supported.

scylla-rust-wrapper/Cargo.toml
```diff
@@ -23,12 +22,17 @@ libc = "0.2.108"
 openssl-sys = "0.9.75"
 openssl = "0.10.32"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
-tracing = "0.1.25"
+tracing = "0.1.37"
```
Gor027, line 25: Why'd you need to update the `tracing` crate version?

scylla-rust-wrapper/Cargo.toml
```diff
@@ -23,12 +22,17 @@ libc = "0.2.108"
...
 
 [build-dependencies]
 bindgen = "0.59.1"
 chrono = "0.4.20"
 
+[dev-dependencies]
+assert_matches = "1.5.0"
+ntest = "0.9.0"
+scylla-proxy = { git = "https://github.com/wprzytula/scylla-rust-driver.git", rev = "1abd7a27"}
```
Gor027: Why the `scylla-proxy` version is on your fork, will there be any changes in the crate that are related to this PR?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -485,3 +485,40 @@ pub unsafe extern "C" fn cass_cluster_set_compression(
...
+            assert_eq!(
+                ___x,
+                crate::cass_error::CassError::CASS_OK,
+                "expected CASS_OK, instead got \"{}\"",
+                ptr_to_cstr(cass_error_desc(___x)).unwrap()
+            );
+        }};
+    }
+    pub(crate) use {assert_cass_error_eq, assert_cass_ok};
+}
```
Gor027: Perhaps it is better to move these macros to the `testing.rs` file, as these are generic macros used for assertions in tests.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -0,0 +1,153 @@
...
+    consistency: CassConsistency,
+) -> CassError {
+    unimplemented!()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_constant_speculative_execution_policy(
+    profile: *mut CassExecProfile,
+    constant_delay_ms: cass_int64_t,
+    max_speculative_executions: i32,
```
Gor027: To stick to the convention let's use `cass_int32_t` instead of `i32`.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -0,0 +1,153 @@
...
+    enabled: cass_bool_t,
+) -> CassError {
+    unimplemented!()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_load_balance_dc_aware(
+    profile: *mut CassExecProfile,
+    local_dc: *const c_char,
+    used_hosts_per_remote_dc: u32,
```
Gor027: `cass_uint32_t` for `u32`

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -0,0 +1,153 @@
...
+    unimplemented!()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_whitelist_filtering(
+    profile: *mut CassExecProfile,
+    hosts: *const c_char,
+) -> CassError {
+    unimplemented!()
+}
```
Gor027: Same as above. Please note that for functions (particularly setters) with string parameters, the API includes twin functions with `_n` suffix in the name meaning that the function signature includes lengths for string parameters.

scylla-rust-wrapper/src/cluster.rs
Gor027: The `assert_cass_error_eq` seems to be a generic one, so I think `assert_cass_ok` is redundant here. There is no difference to use `assert_cass_error_eq` and to pass `CassError::CASS_OK` instead of `assert_cass_ok`.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -258,4 +307,69 @@ mod tests {
...
+                    assert_eq!(
+                        cass_execution_profile_set_load_balance_dc_aware(
+                            profile_raw,
+                            "eu\0".as_ptr() as *const i8,
+                            0,
+                            1
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
+                }
```
Gor027: Can't you use `assert_cass_error_eq` here instead of `assert_eq`?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -690,4 +737,109 @@ mod tests {
...
+                    ));
+                    assert!(cluster.execution_profile_map.len() == 1);
+                    let _profile = cluster
+                        .execution_profile_map
+                        .get(&"profile1".to_owned().try_into().unwrap())
+                        .unwrap();
+
+                    let mut c_str: *const c_char = std::ptr::null();
+                    let mut c_strlen: size_t = 0;
+                    write_str_to_c("profile1", &mut c_str, &mut c_strlen);
```
Gor027: Why not use `argconv::str_to_c_str_n`?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -690,4 +737,109 @@ mod tests {
...
+                        cluster_raw,
+                        c_str,
+                        c_strlen,
+                        exec_profile_raw
+                    ));
+                    assert!(cluster.execution_profile_map.len() == 1);
+                    let _profile = cluster
+                        .execution_profile_map
+                        .get(&"profile1".to_owned().try_into().unwrap())
+                        .unwrap();
```
Gor027, line 758: Wouldn't it be better to use `assert_ne!(..., None)` instead of panicking?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -690,4 +737,109 @@ mod tests {
...
+                {
+                    // NULL name
+                    assert_eq!(
+                        cass_cluster_set_execution_profile(
+                            cluster_raw,
+                            std::ptr::null(),
+                            exec_profile_raw
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
```
Gor027: Again, why not use `assert_cass_error_eq`?

scylla-rust-wrapper/src/session.rs
```diff
@@ -85,9 +128,15 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
     let batch_from_raw = ptr_to_ref(batch_raw);
-    let state = batch_from_raw.state.clone();
+    let mut state = batch_from_raw.state.clone();
     let request_timeout_ms = batch_from_raw.batch_request_timeout_ms;
 
+    // DO NOT refer to `batch_from_raw` inside the async block, as I've done just to face a segfault.
+    let batch_exec_profile_name = batch_from_raw.exec_profile_name.clone();
+    let batch_exec_profile_handle = batch_from_raw.exec_profile_handle.clone();
+    #[allow(unused)]
+    let batch_from_raw = (); // Hardening shadow to avoid use-after-free.
```
Gor027: I guess it is a trick to avoid using the `batch_from_raw` inside the async block. I think just the above comment is enough to prevent it.

scylla-rust-wrapper/src/session.rs
```diff
@@ -100,6 +149,17 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
         let session = &cass_session_inner.session;
 
+        let handle = cass_session_inner
+            .get_or_resolve_profile_handle(
+                batch_exec_profile_handle.as_deref(),
+                batch_exec_profile_name.as_ref(),
+            )
+            .await?;
+
+        Arc::make_mut(&mut state)
```
Gor027, line 202: According to the CPP driver documentation, functions that modify `CassSession` and `CassFuture` object's state should be thread-safe. Are you sure that the usage of `Arc::make_mut` does not break that?

scylla-rust-wrapper/src/session.rs
```diff
@@ -158,6 +224,20 @@ pub unsafe extern "C" fn cass_session_execute(
...
+            .get_or_resolve_profile_handle(
+                statement_exec_profile_handle.as_deref(),
+                statement_exec_profile_name.as_ref(),
+            )
+            .await?;
+
+        match &mut statement {
+            Statement::Simple(query) => query.query.set_execution_profile_handle(handle),
+            Statement::Prepared(prepared) => {
+                Arc::make_mut(prepared).set_execution_profile_handle(handle)
```
Gor027, line 276: Same question as above.

scylla-rust-wrapper/src/batch.rs
```diff
@@ -2,17 +2,34 @@ use crate::argconv::{free_boxed, ptr_to_ref, ptr_to_ref_mut};
...
+    // it can't be set eagerly is simple: the mentioned function does not provide any access
+    // to the Session.
+    pub(crate) exec_profile_name: Option<ExecProfileName>,
+
+    // This is the handle to be cloned as per-statement handle upon each query.
+    // We have to do that due to limitations of CPP driver API:
+    // `cass_batch_set_execution_profile()` does not have access to exec profiles map,
+    // whereas `cass_session_execute_batch()` can't mutate the shared statement
+    // (but only its own copy of the statement).
+    pub(crate) exec_profile_handle: Option<Arc<OnceCell<ExecutionProfileHandle>>>,
```
Gor027: I think I do not understand why `OnceCell<ExecutionProfileHandle>` should be wrapped under `Arc`.

Gor027 (changes requested):
* Some of the commits do not compile successfully (e.g. updating the Rust driver version). * The PR description mentions that new tests are added to the `build` and `cassandra` workflows but the files are unmodified. 
* Please add a "fixes" annotation to the description that the PR will fix #38.
* Have you tried to run tests in `test_exec_profile`, can we fix and enable them in CI? I think we can also add Rust tests in CI, however, `cargo test` generated a lot of warnings that should be suppressed. 
* In the examples catalog we have the `execution_profiles` example, have you tried to run it with valgrind to check memory corruption errors?

scylla-rust-wrapper/Cargo.toml
wprzytula: Without that, I was unable to compile `scylla` crate. Although I don't fully understand why it is the case, I decided to wave it aside.

scylla-rust-wrapper/Cargo.toml
wprzytula: Yes, and I'll open a PR with those changes soon.

scylla-rust-wrapper/src/cluster.rs
wprzytula: This is debatable. I don't consider `assert_ne!` option anyhow better than an `unwrap()`, which as well conveys meaning of "verify that this value is a `Some`", which is btw even more to the point than `assert_ne!`'s "verify that this value is not `None`.

scylla-rust-wrapper/src/session.rs
wprzytula: Rust is all about relying on compile-time errors to avoid faulty code. Why not use such a simple trick to achieve that?

scylla-rust-wrapper/src/batch.rs
wprzytula: The purpose of `exec_profile_handle` field, being a `OnceCell`, is to make it possible to resolve the profile only once after its alterations (e.g. re-setting an exec profile name) take place, and then (on further executions) used straight from that field without resolution. In order to achieve that, we want the resolution's result to be saved into the `Batch` struct that is shared between calls to `cass_session_execute_batch()`.
Please note that we can't use references to that `Batch` struct inside of a `CassFuture` returned from `cass_session_execute_batch()` because of lifetime issues (possible _use-after-free_). Therefore, we have to clone `Batch`'es contents to inside the future and perform the resolution there. If the `OnceCell` wasn't shared under `Arc` and we cloned it into the future, then the resolution inside the future would never propagate into the shared `Batch` struct.
The same is true for `Arc`ed `exec_profile_handle` field in `Statement`.

Is it clear now?

Gor027: Yes, thanks for the explanation.

wprzytula: I think I'll put this explanation into a comment.

scylla-rust-wrapper/src/session.rs
wprzytula: I can't think of any way in which `Arc::make_mut()` can violate thread safety. Can you elaborate?

wprzytula:
> Some of the commits do not compile successfully (e.g. updating the Rust driver version). * The PR description mentions that new tests are added to the `build` and `cassandra` workflows but the files are unmodified.

I see no way to have both small, logically separate commits **and** commits that all compile. The bump of Rust driver version is a monolith, which changes API a lot and hence we can't expect the next commits to compile - they have to fix many issues before.

> Please add a "fixes" annotation to the description that the PR will fix [CassExecProfile #38](https://github.com/scylladb/cpp-rust-driver/issues/38).

Done.
 
> Have you tried to run tests in `test_exec_profile`, can we fix and enable them in CI? I think we can also add Rust tests in CI, however, `cargo test` generated a lot of warnings that should be suppressed.

About `cargo test` in CI: done.
About existing tests: I've described them in detail in the PR's description.
 
> In the examples catalog we have the `execution_profiles` example, have you tried to run it with valgrind to check memory corruption errors?

Yes, these are the results, so I assume everything's all right:
```
LEAK SUMMARY:
    definitely lost: 0 bytes in 0 blocks
    indirectly lost: 0 bytes in 0 blocks
    possibly lost: 123,671 bytes in 1,162 blocks
    still reachable: 548,741 bytes in 676 blocks
    suppressed: 0 bytes in 0 blocks
```

piodul:
> > Some of the commits do not compile successfully (e.g. updating the Rust driver version). * The PR description mentions that new tests are added to the `build` and `cassandra` workflows but the files are unmodified.
> 
> I see no way to have both small, logically separate commits **and** commits that all compile. The bump of Rust driver version is a monolith, which changes API a lot and hence we can't expect the next commits to compile - they have to fix many issues before.

if rust driver version bump is such a big effort then it should be done in a separate PR. If the adjustments required for the new version are straightforward and/or independent from each other (I's expect them to be), then it's fine to do it in a single commit.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -8,85 +8,97 @@ use crate::types::*;
...
+#[derive(Clone, Debug)]
+pub(crate) struct LoadBalancingConfig {
+    pub(crate) token_awareness_enabled: bool,
+    pub(crate) dc_awareness: Option<DcAwareness>,
+    pub(crate) latency_awareness_enabled: bool,
+    pub(crate) latency_awareness_builder: LatencyAwarenessBuilder,
+}
+impl LoadBalancingConfig {
+    // This is `async` to prevent running this function from beyond tokio context,
+    // as it results in panic due to DefaultPolicy::new() spawning a tokio task.
```
Lorak-mmk: Did you mean `DefaultPolicyBuilder::new()`?

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -65,12 +73,56 @@ pub unsafe extern "C" fn cass_execution_profile_free(profile: *mut CassExecProfi
...
+    unimplemented!("This feature is missing from Rust driver that operates underneath.");
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_blacklist_filtering(
+    _profile: *mut CassExecProfile,
+    _hosts: *const c_char,
+) -> CassError {
+    unimplemented!("This feature is missing from Rust driver that operates underneath.");
+}
```
Lorak-mmk: If those functions are unimplemented, wouldn't it be better to actually not implement them, so that user will see the error at link time instead of runtime?

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -123,23 +188,44 @@ pub unsafe extern "C" fn cass_execution_profile_set_request_timeout(
...
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_execution_profile_set_serial_consistency(
     profile: *mut CassExecProfile,
     serial_consistency: CassConsistency,
 ) -> CassError {
-    unimplemented!()
+    let profile_builder = ptr_to_ref_mut(profile);
+    // TODO: consider CassConsistency::UNKNOWN case
```
Lorak-mmk: How is it handled now? Will it return error from this function?

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -152,7 +160,31 @@ pub unsafe extern "C" fn cass_execution_profile_set_latency_aware_routing(
...
+        .latency_awareness_enabled = enabled != 0;
+
+    CassError::CASS_OK
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_execution_profile_set_latency_aware_routing_settings(
+    profile: *mut CassExecProfile,
+    exclusion_threshold: cass_double_t,
+    _scale_ms: cass_uint64_t, // Currently ignored, TODO: add this parameter to Rust driver
```
Lorak-mmk, line 290: If it's not used, it would be better to return error if it's used instead of silently failing.

scylla-rust-wrapper/src/session.rs
```diff
@@ -4,6 +4,7 @@ use crate::cass_error::*;
 use crate::cass_types::{get_column_type, CassDataType, UDTDataType};
```
Lorak-mmk, line 4: Regarding commit message in commit 8b59e27a36ccff64b7764749040adf0f3babd87a: did you mean `LatencyAwarenessBuilder::new()`?

scylla-rust-wrapper/src/session.rs
```diff
@@ -45,7 +48,7 @@ pub unsafe extern "C" fn cass_session_connect(
     cluster_raw: *const CassCluster,
 ) -> *const CassFuture {
     let session_opt = ptr_to_ref(session_raw);
-    let cluster: CassCluster = (*ptr_to_ref(cluster_raw)).clone();
+    let cluster: CassCluster = ptr_to_ref(cluster_raw).clone();
```
Lorak-mmk: This looks like an unrelated change, shouldn't it be in it's own commit?

scylla-rust-wrapper/src/batch.rs
```diff
@@ -2,17 +2,46 @@ use crate::argconv::{free_boxed, ptr_to_ref, ptr_to_ref_mut};
 use crate::cass_error::CassError;
```
Lorak-mmk, line 2: This comment is regarding commit e997b77f48c9124cc0463ef679dfe2a6b07e02e6.
If I understand correctly, only one of `exec_profile_name` and `exec_profile_handle` is actually relevant - if name is set, it will be resolved to handle on next execute, `exec_profile_handle` will be set to None and `exec_profile_handle` replaced with new handle - so why are they 2 members?? Shouldn't it be an enum with 2 variants (name or a handle)?

scylla-rust-wrapper/src/session.rs
```diff
@@ -22,16 +22,61 @@ use scylla::transport::execution_profile::ExecutionProfileHandle;
...
+                CassError::CASS_ERROR_LIB_EXECUTION_PROFILE_INVALID,
+                format!("{} does not exist", name.deref()),
+            )
+        })
+    }
+
+    // Clippy claims it is possible to make this `async fn`, but it's terribly wrong,
+    // because async fn can't have its future bound to a specific lifetime, which is
+    // required in this case.
+    #[allow(clippy::manual_async_fn)]
```
Lorak-mmk, line 53: So this doesn't compile if you make it an async function? If you have some spare time you could try to open an issue in clippy repo about this

scylla-rust-wrapper/src/session.rs
```diff
@@ -617,11 +617,13 @@ mod tests {
         cass_types::CassBatchType,
```
Lorak-mmk, line 636: Commit 1f3819a95cc58089fdafe2e1590d07486ef001f3 needs a proper message instead of TODO

scylla-rust-wrapper/src/session.rs
```diff
@@ -127,11 +127,15 @@ pub unsafe extern "C" fn cass_session_connect(
         }
         let mut exec_profile_map = HashMap::with_capacity(exec_profile_builder_map.len());
         for (name, builder) in exec_profile_builder_map {
-            exec_profile_map.insert(name, builder.build().await.into_handle());
+            exec_profile_map.insert(name.clone(), builder.clone().build().await.into_handle());
```
Lorak-mmk: The function already takes ownership of the hashmap - why are those clones necessary?

Lorak-mmk (changes requested):
Wow, really awesome PR, thanks.
I did not review tests yet. I also did not review commits that add per statement retry policy - afaik those commits will be significantly changed because this feature was re-added to Rust driver.

scylla-rust-wrapper/src/session.rs
wprzytula: Do you think it actually deserves its own commit?

scylla-rust-wrapper/src/cluster.rs
wprzytula: Actually, I meant an action equivalent to `DefaultPolicy::new()` if such nonpublic method existed. As such does not exist, though, let's refer to `DefaultPolicyBuilder::build()`.

scylla-rust-wrapper/src/batch.rs
wprzytula: You're right, an enum would suit the use semantics best. Will change it.

wprzytula: On a second though, this is not feasible.

In the scenario of an enum, the transition from variant `ToResolve(ExecProfileName)` into `Resolved(Arc<OnceCell<ExecProfileHandle>>)` would have to be done inside of `cass_session_execute()`, which is a function that takes CassSession by immutable reference (in Rust terms) or that is thread-safe (in C terms). Such transition is not thread-safe, whereas the mere initialisation of a `OnceCell` indeed is: if a empty `OnceCell` is created in a non-thread-safe `cass_statement_set_execution_profile()`, then it can be safely initialised in a non-thread-safe `cass_session_execute()`.

Am I convincing?

scylla-rust-wrapper/src/session.rs
Lorak-mmk: Let's leave it as is

scylla-rust-wrapper/src/batch.rs
Lorak-mmk: Why not store `Arc<OnceCell<OurNewEnum>>`  (`OurNewEnum` wouldn't have `Arc` / `OnceCell` inside it in this scenario`)? In this case:
- New struct will have empty `OnceCell` - this means no execution profile has been set.
- `cass_statement_set_execution_profile` will set contents of `OnceCell` to enum variant with profile name.
- `cass_session_execute()` will work similarly to how it does now - clone the Arc, update the contents if the profile is not resolved.

wprzytula: As the name explains, `OnceCell` allows for only one assignment. This means that once we set the `OnceCell`'s content to one variant of our enum, we will no longer be able to change it into another.

wprzytula: The whole thing is about making exec profile resolution happen only once (not upon every call to `cass_session_execute()`. This is done with `OnceCell` (shared in an `Arc`), which is initialised with the profile _handle_, not name.

Lorak-mmk: > As the name explains, `OnceCell` allows for only one assignment. This means that once we set the `OnceCell`'s content to one variant of our enum, we will no longer be able to change it into another.

This is not true, it has `take` method: https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.take
But of course a different primitive  could be used here, like `Mutex` if you feel it fits better.

wprzytula: Why would I use Mutex when OnceCell has exactly the semantics I would like to have here? Namely, resolving and setting an exec profile handle should happen exactly once.

scylla-rust-wrapper/Cargo.toml
wprzytula: They've been already merged.

scylla-rust-wrapper/src/exec_profile.rs
wprzytula: The problem is, this function returns `void`...

scylla-rust-wrapper/src/batch.rs
wprzytula: I've got convinced to use `RwLock` as you proposed. The new version, which I will push soon, will contain the `PerStatementExecutionProfile` holding an `RwLock` to either `Unresolved` or `Resolved` variant.

scylla-rust-wrapper/src/exec_profile.rs
wprzytula: Yes. Therefore, I change this code to map `CASS_CONSISTENCY::UNKNOWN` into `None` in case of `SerialConsistency`

wprzytula:
v2:
- all commits now compile and pass tests. All changes related to Rust driver's version update are squashed in one commit, but I hope they're still readable,
- the driver used now is a rev from the main repo's `main` branch,
- retry policy emulation is no longer needed (after a change in Rust driver), so this feature was removed and replaced with the plain settings on `Statement` and `Batch`,
- `OnceCell` was replaced with `RwLock` in order to have only one field instead of two distinct (`exec_profile_name` and `exec_profile_handle`, which anyway couldn't coexist at one `Statement/Batch`).

wprzytula:
v2.1 - fixed memory leak due to an oversight during updating Rust driver's version

scylla-rust-wrapper/src/statement.rs
```diff
@@ -255,17 +255,19 @@ pub unsafe extern "C" fn cass_statement_set_retry_policy(
     statement: *mut CassStatement,
     retry_policy: *const CassRetryPolicy,
 ) -> CassError {
-    let retry_policy_from_raw: &dyn scylla::retry_policy::RetryPolicy =
+    let arced_retry_policy: Arc<dyn scylla::retry_policy::RetryPolicy> =
+        // Shouldn't this accept NULL ptr for unsetting retry policy on statement?
```
Lorak-mmk: I think we discussed on Slack that it should be supported. Can you add support, or make a TODO comment?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -508,3 +516,105 @@ pub unsafe extern "C" fn cass_cluster_set_compression(
...
+) {
+    let cluster = ptr_to_ref_mut(cluster);
+    cluster.load_balancing_config.latency_awareness_enabled = enabled != 0;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_latency_aware_routing_settings(
+    cluster: *mut CassCluster,
+    exclusion_threshold: cass_double_t,
+    _scale_ms: cass_uint64_t, // Currently ignored, TODO: add this parameter to Rust driver
```
Lorak-mmk, line 546: Please add a warning when this is set to a value different than a "default" - where default is the value that corresponds with current rust driver behavior.

scylla-rust-wrapper/src/testing.rs
```diff
@@ -1 +1,15 @@
...
+        let ___x = $expr;
+        assert_eq!(
+            ___x,
+            $error,
+            "expected \"{}\", instead got \"{}\"",
+            ptr_to_cstr(cass_error_desc($error)).unwrap(),
+            ptr_to_cstr(cass_error_desc(___x)).unwrap()
+        );
+    }};
+}
```
Lorak-mmk, line 14: I think this commit is in the wrong order - you used this macro 2 commits ago.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -9,11 +14,50 @@ use crate::retry_policy::CassRetryPolicy;
...
+        self.inner.build()
+    }
+}
+
+/// Represents a non-empty execution profile name.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ExecProfileName(String);
+#[derive(Debug, PartialEq, Eq)]
+pub struct EmptyProfileName;
+impl TryFrom<String> for ExecProfileName {
```
Lorak-mmk: Add some empty lines maybe?

scylla-rust-wrapper/src/exec_profile.rs
Lorak-mmk: So I guess that the TODO comment can be removed now?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -444,9 +445,9 @@ pub unsafe extern "C" fn cass_cluster_set_constant_speculative_execution_policy(
         retry_interval: Duration::from_millis(constant_delay_ms as u64),
     };
 
-    cluster.default_execution_profile_builder =
-        std::mem::take(&mut cluster.default_execution_profile_builder)
-            .speculative_execution_policy(Some(Arc::new(policy)));
+    exec_profile_builder_modify(&mut cluster.default_execution_profile_builder, |builder| {
```
Lorak-mmk, line 459: I thought there was a method on exec profile for that (which called `exec_profile_builder_modify` underneath).

Lorak-mmk (changes requested):
Finally managed to get trough it all properly. Really awesome PR, I love the tests and refactors you did. Left some minor comments.

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: I see that support is added later, nvm.

scylla-rust-wrapper/src/cluster.rs
wprzytula: I've opened a PR to Rust driver which introduces the missing `scale` parameter. This should fix the problem.

scylla-rust-wrapper/src/cluster.rs
wprzytula: `CassExecProfile::modify_in_place()` works, unsurprisingly, only for `CassExecProfile`, which is not involved here.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -299,4 +357,72 @@ mod tests {
...
+                    );
+                    assert_eq!(
+                        cass_execution_profile_set_load_balance_dc_aware(
+                            profile_raw,
+                            "eu\0".as_ptr() as *const i8,
+                            0,
+                            1
+                        ),
+                        CassError::CASS_ERROR_LIB_BAD_PARAMS
+                    );
```
Gor027: You can use `aseert_cass_error_eq` here as well.

scylla-rust-wrapper/src/session.rs
Gor027: Unresolving as it is not fixed(?).

scylla-rust-wrapper/src/session.rs
```diff
@@ -565,6 +565,7 @@ pub unsafe extern "C" fn cass_session_get_schema_meta(
```
Gor027, line 619: Nit: typo in the commit message: "Default or Fallthrough".

scylla-rust-wrapper/src/session.rs
wprzytula: As I can see, it is resolved. Before `0bf96b5 CassSession: Elide CassCluster clone in connect()`, the clone is needed, and beginning with that commit `clone()` is removed.

scylla-rust-wrapper/src/session.rs
```diff
@@ -1262,4 +1263,43 @@ mod tests {
...
+                    cass_cluster_free(cluster_raw);
+
+                    cass_future_wait(cass_future);
+                    // The exact outcome is not important, we only test that we don't segfault.
+                }
+                cass_execution_profile_free(profile_raw);
+                cass_session_free(session_raw);
+            }
+        }
+    }
```
Gor027, line 1358: Thanks for this commit, it helped a lot to implement serverless support without memory corruption errors.

Gor027 (approved):
LGTM. Minor comments are left in some commits as their fixes were added in further commits. Overall, thanks for this PR as besides adding support for execution profiles it also increases the quality of the driver with "all-inclusive" tests.

scylla-rust-wrapper/src/cluster.rs
constantfold: This won't be addressed in this PR, created an issue to not forget about it later: https://github.com/scylladb/cpp-rust-driver/issues/114

PR #108, opened 2023-04-12
Add serverless support

### LINKED ISSUE #105: Serverless support
The ScyllaDB Rust driver will soon be enhanced with support for serverless https://github.com/scylladb/scylla-rust-driver/pull/614. The C++ bindings should also support serverless and this will imply to:
* remove existing DBaaS-related functions from API
* add appropriate API functions for serverless support
* modify `DbaasTests` tests to properly test serverless support

It is also necessary to add code examples in https://github.com/scylladb/serverless-examples.

Gor027:
## Pre-review checklist

This PR is dependent upon #107.
Unit tests and integration tests for serverless are supported in the Rust driver.

Fixes #105


scylla-rust-wrapper/Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = ["database"]
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "1a913a19", features = ["ssl"]}
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "1a913a19", features = ["ssl", "cloud"]}
 tokio = { version = "1.1.0", features = ["full"] }
```
wprzytula, line 14: We should rather mimic the Rust driver's behaviour (not to put unnecessary burden on non-cloud users of the driver) and provide cloud functionalities only if they request it explicitly (use features).

scylla-rust-wrapper/src/cloud.rs
```diff
@@ -0,0 +1,62 @@
...
+    let config_path = ptr_to_cstr_n(path, path_length);
+
+    if let Some(config_path) = config_path {
+        if config_path == "invalid_filename" {
+            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+        }
+
+        cluster.connection_bundle_path = Some(config_path.to_string());
+        return CassError::CASS_OK;
+    }
```
wprzytula: What is this mysterious hardcoded `"invalid_filename"` ?

scylla-rust-wrapper/extern/cassandra.h
```diff
@@ -646,7 +646,8 @@ typedef enum  CassErrorSource_ {
   CASS_ERROR_SOURCE_LIB,
   CASS_ERROR_SOURCE_SERVER,
   CASS_ERROR_SOURCE_SSL,
```
wprzytula, line 648: Oh, so we have duplicated `cassandra.h` in our repo? Perhaps we could somehow avoid this, e.g. by use of soft links.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -65,6 +66,8 @@ pub(crate) struct DcAwareness {
     pub(crate) local_dc: String,
 }
 
+pub type SessionConfigError = CloudConfigError;
```
wprzytula: Why such name?

scylla-rust-wrapper/extern/cassandra.h
Gor027: Yes, but better to do that in a separate PR. Will create an issue to not forget about it.

scylla-rust-wrapper/src/cluster.rs
Gor027: What would you suggest instead? Currently, there may be errors when a cloud session is created, but that may change in the future.

scylla-rust-wrapper/src/cloud.rs
Gor027: I just kept the previous implementation of the function. Will fix it, tnx.

scylla-rust-wrapper/Cargo.toml
Gor027: I will see if it is possible to dynamically enable features of a certain dependency.

wprzytula: This is definitely possible, as we do this in Rust driver itself. Excerpt from Rust driver's `Cargo.toml`:
```toml
[features]
default = []
ssl = ["dep:tokio-openssl", "dep:openssl"]
cloud = ["ssl", "scylla-cql/serde", "dep:serde_yaml", "dep:serde", "dep:url", "dep:base64"]
secret = ["scylla-cql/secret"] # <- this is the most suitable for your situation
```

scylla-rust-wrapper/src/cluster.rs
wprzytula: Then wouldn't an enum convey the intention better?
```rust
pub enum SessionConfigError {
  Cloud(CloudConfigError),
  // here new variants can be added in the future
}
```

Gor027: Done.

scylla-rust-wrapper/Cargo.toml
Gor027: Fixed it in a separate commit: [4ad0ede](https://github.com/scylladb/cpp-rust-driver/pull/108/commits/4ad0ededc8de73b81344abbeaaf31f5db6d33f3c)

CMakeLists.txt
```diff
@@ -240,7 +240,7 @@ ExternalProject_Add(
     scylla-rust-cpp-driver
     DOWNLOAD_COMMAND ""
     CONFIGURE_COMMAND ""
-    BUILD_COMMAND cargo build $<IF:$<CONFIG:Debug>,,--release>
+    BUILD_COMMAND cargo build $<IF:$<CONFIG:Debug>,,--release> --features cloud,ssl
     BINARY_DIR "${CMAKE_SOURCE_DIR}/scylla-rust-wrapper"
     INSTALL_COMMAND ""
```
wprzytula: Are we going to allow building bindings without `cloud` or `ssl` feature?

scylla-rust-wrapper/Cargo.toml
Gor027: After discussing it with @Lorak-mmk, decided to enable SSL and cloud features.

src/driver_config.hpp
```diff
@@ -1,7 +1,7 @@
 #ifndef DATASTAX_INTERNAL_DRIVER_CONFIG_HPP
 #define DATASTAX_INTERNAL_DRIVER_CONFIG_HPP
 
-#define HAVE_KERBEROS
+/* #undef HAVE_KERBEROS */
 #define HAVE_OPENSSL
```
wprzytula: Please explain this change.

scylla-rust-wrapper/src/external.rs
```diff
@@ -70,6 +70,12 @@ pub unsafe extern "C" fn cass_error_desc(error: CassError) -> *const c_char {
         }
         CassError::CASS_ERROR_SSL_PROTOCOL_ERROR => "Protocol error\0",
         CassError::CASS_ERROR_SSL_CLOSED => "Connection closed\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_OPEN_ERROR => "Unable to open config file\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_PARSE_ERROR => "Unable to parse config file\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_DECODE_ERROR => "Unable to decode base64 key/cert\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_VALIDATION_ERROR => "Invalid cloud config\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_BAD_SSL_PARAMS => "Unable to parse SSL key/cert\0",
+        CassError::CASS_ERROR_CLOUD_CONFIG_UNKNOWN_ERROR => "Unknown error\0",
```
wprzytula: Please make sure to mention "cloud" in each of these messages.

scylla-rust-wrapper/src/cloud.rs
```diff
@@ -0,0 +1,63 @@
...
+    if let Some(config_path) = config_path {
+        if !Path::new(config_path).exists() {
+            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+        }
+
+        cluster.connection_bundle_path = Some(config_path.to_string());
+        return CassError::CASS_OK;
+    }
+
+    CassError::CASS_ERROR_LIB_BAD_PARAMS
```
wprzytula: Please don't put error-case returns on the default path and ok-case returns inside an `if`. This is counterintuitive.
My suggestion:
```suggestion
    if let Some(config_path) = config_path {
        if !Path::new(config_path).exists() {
            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
        }

        cluster.connection_bundle_path = Some(config_path.to_string());
        CassError::CASS_OK
    } else {
        CassError::CASS_ERROR_LIB_BAD_PARAMS
    }
```

wprzytula: Another way, to keep the ok-case linear:
```suggestion
    let config_path = if let Some(path) = config_path {
        if !Path::new(config_path).exists() {
            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
        }
        path
    } else {
       return CassError::CASS_ERROR_LIB_BAD_PARAMS;
    };
    
    cluster.connection_bundle_path = Some(config_path.to_string());
    CassError::CASS_OK
```

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -65,6 +68,72 @@ pub(crate) struct DcAwareness {
...
+}
+
+impl CassSessionBuilder {
+    async fn execution_profile_handle(
+        self,
+        mut execution_profile_builder: ExecutionProfileBuilder,
+        load_balancing_config: LoadBalancingConfig,
+    ) -> Result<Self, SessionConfigError> {
+        let load_balancing = load_balancing_config.clone().build().await;
+        execution_profile_builder = execution_profile_builder.load_balancing_policy(load_balancing);
```
wprzytula: This `load_balancing_config.clone()` seems redundant.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -65,6 +68,72 @@ pub(crate) struct DcAwareness {
...
+                CassSessionBuilder::DefaultSessionBuilder(
+                    session_builder.default_execution_profile_handle(
+                        execution_profile_builder.build().into_handle(),
+                    ),
+                )
+            }
+            CassSessionBuilder::CloudSessionBuilder(cloud_session_builder) => {
+                CassSessionBuilder::CloudSessionBuilder(
+                    cloud_session_builder.default_execution_profile_handle(
+                        execution_profile_builder.build().into_handle(),
```
wprzytula: Let's extract `handle`:
```rust
let handle = execution_profile_builder.build().into_handle();
```
and then use it instead of repeating the code that creates the handle.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -93,23 +164,36 @@ pub struct CassCustomPayload;
...
+            match cloud_session_builder {
+                Ok(builder) => Ok(CassSessionBuilder::CloudSessionBuilder(builder)),
+                Err(err) => Err(SessionConfigError::Cloud(err)),
+            }
+        } else {
+            let known_nodes: Vec<_> = cluster
+                .contact_points
+                .iter()
+                .map(|cp| format!("{}:{}", cp, cluster.port))
+                .collect();
```
wprzytula: I had extended `known_nodes()` in Rust driver specially for this case to avoid collecting, and then got rid of it. I'm unhappy that you revert this `collect()` back.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -93,23 +164,36 @@ pub struct CassCustomPayload;
...
-        .map(|cp| format!("{}:{}", cp, cluster.port));
-    let mut execution_profile_builder = cluster.default_execution_profile_builder.clone();
+) -> impl Future<Output = Result<CassSessionBuilder, SessionConfigError>> + 'static {
+    let session_builder: Result<CassSessionBuilder, SessionConfigError> =
+        if let Some(ref cloud_config_path) = cluster.connection_bundle_path {
+            let cloud_session_builder = CloudSessionBuilder::new(Path::new(cloud_config_path));
+            match cloud_session_builder {
+                Ok(builder) => Ok(CassSessionBuilder::CloudSessionBuilder(builder)),
+                Err(err) => Err(SessionConfigError::Cloud(err)),
+            }
```
wprzytula, line 171: In the cloud case, cloud-agnostic settings are not preserved. Consider a sequence of calls:
```C
cass_cluster_set_cloud_secure_connection_bundle(/*...*/);
cass_cluster_set_tcp_nodelay(/*...*/); // This is not respected
cass_session_connect(/*...*/);
// The resulting session has TCP nodelay set to default, not respecting the setting above 
```

The problem is that Rust sees different instantiations of `GenericSessionBuilder` as distinct types, even though they share the same underlying `SessionConfig` and most of API. One possible solution is to only hold `SessionConfig` in `CassCluster` (instead of `SessionBuilder`) and:
- either construct appropriate instantiation of `GenericSessionBuilder` on-demand (upon call to `cass_session_connect()`, in `build_session_builder()`),
- or skip `GenericSessionBuilder` API at all and use bare `SessionConfig` together with `Session::connect()`.

wprzytula (changes requested):
As I've found some serious errors, it is clear to me that specific tests in bindings for cloud mode are a must. Please add them to this PR.

src/driver_config.hpp
Gor027: I did not notice that perhaps this was auto-modified after a build.

scylla-rust-wrapper/src/cluster.rs
Gor027: Sorry, applying stashed changes reverted this, will fix it.

scylla-rust-wrapper/src/cluster.rs
Gor027: * Using bare `SessionConfig` *maybe* would require huge refactoring to get rid of building the default execution profile and other configurations in `build_session_builder`. But if it is possible, I believe we should follow this direction. 
* It is easy with `SessionBuilder` to just set a previously created session config and then set up other configurations before building a session object. However, the same cannot be done with `CloudSessionBuilder` as its constructor sets some configs which we should not override. 

I think the easiest way is to keep all common config params in the `CassCluster` object (e.g. tcp_nodelay, connect_tmeout, etc.) and after constructing an appropriate instantiation of `GenericSessionBuilder`, manually set these parameters.

wprzytula: On the other hand, _the easiest way_ that you mention involves duplication of many config params and is prone to accidentally forgetting about some of them. Using bare `SessionConfig` lets us have all those params already set in the proper place.

Gor027: I agree, but having bare `SessionConfig` implies having duplication of some `SessionBuilder` setters' code, e.g. `user()`, and remembering to apply changes to these codes in both drivers.

wprzytula: What is the solution that you decide to settle on?

Gor027: To replace `SessionBuilder` with `SessionConfig`. I have added some necessary changes to the Rust driver to achieve this. https://github.com/scylladb/scylla-rust-driver/pull/750

wprzytula: Bear in mind disadvantages that you yourself pointed out: 
> having bare SessionConfig implies having duplication of some SessionBuilder setters' code, e.g. `user()`

to avoid code duplication, i suggest moving logic from `SessionBuilder` to correspoding `SessionConfig` setters. This way, the logic for creating `CloudConfig`, as well as e.g. `user()`'s logic would be accessible for us even if bypassing `SessionBuilder` as you prefer.

Lorak-mmk:
I think we can close this as serverless is no longer a thing

PR #111, opened 2023-04-26
Logging: set event subscriber globally for all threads

### LINKED ISSUE #110: Logging: propogate tracing subscriber for all threads
Currently, the `tracing::subscriber::set_default` function sets our custom subscriber as the default subscriber for the duration of the lifetime of the returned `DefaultGuard` value. However, it does not propagate the current thread's default subscriber to all threads because of which some log events (including Rust driver's logs) are not captured. To check that, run a simple test example with `RUNTIME` in `lib.rs`  built with `new_current_thread`:

```rust
pub static ref RUNTIME: Runtime =
            tokio::runtime::Builder::new_current_thread()
                    .enable_all()
                    .build()
                    .unwrap();
```

As the integration tests set a custom log callback, all logs from the test example will be written to a file in `build/log/{TEST}` directory which is not the case with a multithreaded environment.

One possible solution is to use the `set_global_default` function which will set the subscriber as default for all threads for the lifetime of the program. However, it can be called once and the subscriber cannot be changed later, for example, to change the log level during the runtime of the application. In fact, the C++ driver does not support dynamic configuration of logging, and the configuration must not be changed after any other driver function is called, e.g. `cass_session_connect`. So, the solution does not contradict to C++ driver's API requirements.
Unfortunately, I could not find a way to have a dynamic logging configuration with the current `tracing_subscriber` API, so if you have an idea to accomplish that, please share it in the scope of this issue.

Gor027:
## Pre-review checklist

The previous attempt for adding logging support in the driver was to set a custom event subscriber with an option to replace it with a new subscriber if the logging configuration is changed during the program's runtime. The `tracing::subscriber::set_default` function was used to set a scoped subscriber and be able to drop it gracefully. However, the documentation of the function fails to mention that the default subscriber is set only for the current thread hence making it unsuitable for the needs of multi-threaded applications. So, as of now, the `tracing::subscriber::set_global_default` is the only reliable way to set a default subscriber available for all threads within an application. However, it can be set only once and cannot be gracefully dropped later, so multiple attempts to set a global default subscriber will fail.
This PR replaces `set_default` with `set_global_default`, which implies that `cass_log_set_level` must be called only once and later attempts to modify the logging configuration will be ignored. For informational purposes, a logging section is added to the README to explain the details of the implementation and usage. Additionally, the test logs from `build` and `cassandra` jobs will be uploaded as an artifact in order to help in some debugging processes in the future.

Fixes #110

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,3 +54,10 @@ jobs:
 :*19.Integration_Cassandra_*\
 :ExecutionProfileTest.InvalidName"
       run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --scylla --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
+
+    - name: Upload test logs
+      uses: actions/upload-artifact@v3
+      if: success() || failure()
+      with:
+        name: test-logs-scylla
+        path: ./log/*
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -57,3 +57,10 @@ jobs:
 :SslTests.Integration_Cassandra_ReconnectAfterClusterCrashAndRestart\
 :ExecutionProfileTest.InvalidName"
         run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --version=3.0.9 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
+
+      - name: Upload test logs
+        uses: actions/upload-artifact@v3
+        if: success() || failure()
+        with:
+          name: test-logs-cassandra
+          path: ./log/*
--- a/README.md
+++ b/README.md
@@ -90,6 +90,31 @@ int main(int argc, char* argv[]) {
 }
 ```
 
+# Logging
+___
+
+The logging API and implementation are compatible with the C++ driver, for more details please refer to the [logging documentation](https://cpp-driver.docs.scylladb.com/master/topics/logging/index.html).
+As the `tracing` framework is used under the hood to instrument the collection of logs from the Rust driver and the Cpp-Rust wrapper,
+the logging level and callback are passed through a custom event subscriber which is globally set as default when `cass_log_set_level` is called.
+So, `cass_log_set_level` *must* be called only once as subsequent attempts trying to modify the globally set event subscriber will be ignored.
+Also, Rust programs using Cpp-Rust driver under the hood must avoid calling `tracing::subscriber::set_global_default` as this will cause conflicts.    
+
+##### Note: The logging configuration must be done before any other driver function is called, otherwise, the default logging callback will be used, and logs will appear on stderr.
+
+```c++
+void on_log(const CassLogMessage* message, void* data) {
+  /* Handle logging */
+}
+
+int main() {
+  void* log_data = NULL /* Custom log resource */;
+  cass_log_set_callback(on_log, log_data);
+  cass_log_set_level(CASS_LOG_INFO);
+
+  /* Create cluster and connect session */
+}
+```
+
 # Features
 ___
 
--- a/scylla-rust-wrapper/src/lib.rs
+++ b/scylla-rust-wrapper/src/lib.rs
@@ -1,12 +1,10 @@
 #![allow(clippy::missing_safety_doc)]
 
-use crate::logging::set_tracing_subscriber_with_level;
 use crate::logging::stderr_log_callback;
 use crate::logging::Logger;
 use lazy_static::lazy_static;
 use std::sync::RwLock;
 use tokio::runtime::Runtime;
-use tracing::dispatcher::DefaultGuard;
 
 #[macro_use]
 mod binding;
@@ -42,9 +40,6 @@ lazy_static! {
         cb: Some(stderr_log_callback),
         data: std::ptr::null_mut(),
     });
-    pub static ref LOG: RwLock<Option<DefaultGuard>> = RwLock::new(Some(
-        set_tracing_subscriber_with_level(tracing::Level::WARN)
-    ));
 }
 
 // To send a Rust object to C:
--- a/scylla-rust-wrapper/src/logging.rs
+++ b/scylla-rust-wrapper/src/logging.rs
@@ -1,6 +1,5 @@
 use crate::argconv::{arr_to_cstr, ptr_to_cstr, ptr_to_ref, str_to_arr};
 use crate::types::size_t;
-use crate::LOG;
 use crate::LOGGER;
 use std::convert::TryFrom;
 use std::fmt::Debug;
@@ -8,7 +7,6 @@ use std::fmt::Write;
 use std::os::raw::{c_char, c_void};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::debug;
-use tracing::dispatcher::DefaultGuard;
 use tracing::field::Field;
 use tracing::Level;
 use tracing_subscriber::layer::Context;
@@ -143,10 +141,10 @@ where
     }
 }
 
-// Sets tracing subscriber with specified `level` and returns `DefaultGuard`.
-// The subscriber is valid for the duration of the lifetime of the returned `DefaultGuard`.
-pub fn set_tracing_subscriber_with_level(level: Level) -> DefaultGuard {
-    tracing::subscriber::set_default(
+// Sets tracing subscriber with specified `level`.
+// The subscriber is valid for the duration of the entire program.
+pub fn set_tracing_subscriber_with_level(level: Level) {
+    tracing::subscriber::set_global_default(
         tracing_subscriber::registry()
             .with(
                 tracing_subscriber::EnvFilter::from_default_env()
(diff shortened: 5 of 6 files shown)


Lorak-mmk:
I'm not that familiar with `tracing` crate, but can't we do the same thing as in rust driver did in https://github.com/scylladb/scylla-rust-driver/pull/668? I mean using `.with_current_subscriber()` on each future that is passed to `tokio::spawn()`. The restriction that `cass_log_set_level` can be called only once is quite restrictive, I don't think cpp-driver has such limitation.

piodul:
> I'm not that familiar with `tracing` crate, but can't we do the same thing as in rust driver did in [scylladb/scylla-rust-driver#668](https://github.com/scylladb/scylla-rust-driver/pull/668)? I mean using `.with_current_subscriber()` on each future that is passed to `tokio::spawn()`. The restriction that `cass_log_set_level` can be called only once is quite restrictive, I don't think cpp-driver has such limitation.

Yes, this sounds better - however, this approach requires rust driver 0.8.0 to work properly.

Lorak-mmk:
So in order to do that we need to merge https://github.com/scylladb/cpp-rust-driver/pull/107 first, right?

piodul:
> So in order to do that we need to merge #107 first, right?

Probably. Or, if a PR that just migrates to the new rust driver version is extracted out of that, then merge that first. I guess it's up to @wprzytula.

Gor027:
> > I'm not that familiar with `tracing` crate, but can't we do the same thing as in rust driver did in [scylladb/scylla-rust-driver#668](https://github.com/scylladb/scylla-rust-driver/pull/668)? I mean using `.with_current_subscriber()` on each future that is passed to `tokio::spawn()`. The restriction that `cass_log_set_level` can be called only once is quite restrictive, I don't think cpp-driver has such limitation.
> 
> Yes, this sounds better - however, this approach requires rust driver 0.8.0 to work properly.

IIUC, this implies having a scoped logger for each future, so logs outside of futures will not be recorded and modified as C++ logs as there is no global default subscriber.

Gor027:
As I see, what is necessary to properly support logging is not yet supported in the tracing crate: https://github.com/tokio-rs/tracing/issues/1517

Lorak-mmk:
> > > I'm not that familiar with `tracing` crate, but can't we do the same thing as in rust driver did in [scylladb/scylla-rust-driver#668](https://github.com/scylladb/scylla-rust-driver/pull/668)? I mean using `.with_current_subscriber()` on each future that is passed to `tokio::spawn()`. The restriction that `cass_log_set_level` can be called only once is quite restrictive, I don't think cpp-driver has such limitation.
> > 
> > 
> > Yes, this sounds better - however, this approach requires rust driver 0.8.0 to work properly.
> 
> IIUC, this implies having a scoped logger for each future, so logs outside of futures will not be recorded and modified as C++ logs as there is no global default subscriber.

What are `logs outside of futures`? Do you mean logs that are logged in `cass_*` functions directly? If so, how many places where such logs are possible do we have? Is it possible to use scoped logger in those places?

Gor027:
> > > > I'm not that familiar with `tracing` crate, but can't we do the same thing as in rust driver did in [scylladb/scylla-rust-driver#668](https://github.com/scylladb/scylla-rust-driver/pull/668)? I mean using `.with_current_subscriber()` on each future that is passed to `tokio::spawn()`. The restriction that `cass_log_set_level` can be called only once is quite restrictive, I don't think cpp-driver has such limitation.
> > > 
> > > 
> > > Yes, this sounds better - however, this approach requires rust driver 0.8.0 to work properly.
> > 
> > 
> > IIUC, this implies having a scoped logger for each future, so logs outside of futures will not be recorded and modified as C++ logs as there is no global default subscriber.
> 
> What are `logs outside of futures`? Do you mean logs that are logged in `cass_*` functions directly? If so, how many places where such logs are possible do we have? Is it possible to use scoped logger in those places?

Yes, for example, logs that may appear in some setters. It is hard to guess where these logs are possible and we should always make sure that a newly added log in the Rust driver also appears in the bindings, so we should add a scoped subscriber to the appropriate function.

Lorak-mmk:
One solution to that could be a proc macro wrapping whole function body in a scoped subscriber. This macro would be placed on any function that calls Rust driver or does the logging itself.

wprzytula:
> One solution to that could be a proc macro wrapping whole function body in a scoped subscriber. This macro would be placed on any function that calls Rust driver or does the logging itself.

This is an interesting idea, could you write a PoC for this?

wprzytula:
@Gor027 
Using this PR's code, I'm still unable to get any logs from the Rust driver.
I'm using the `prepared` example, modified by adding these lines at the `main()` beginning:
```c
cass_log_set_level(CASS_LOG_TRACE);
cass_log_set_callback(callback, NULL);
```
with callback defined as:
```c
void callback(CassLogMessage const* msg, void *data) {
  printf("%s::%s:%i: %s\n", msg->file, msg->function, msg->line, msg->message);
}
```

There's no difference whether I call `cass_log_set_callback()` or not.

Gor027:
I tested with the modifications you mentioned and it worked for me:

```
1686564005967 [DEBUG] (scylla_cpp_driver::logging:169) message: Log level is set to TRACE
scylla::transport::connection_pool:::631: message: [127.0.0.1:9042] Started asynchronous pool worker
scylla::transport::connection_pool:::739: message: [127.0.0.1:9042] Will open the first connection to the node
scylla::transport::connection_pool:::680: pool_state: "[(0, [])]"
scylla::transport::connection_pool:::823: message: [127.0.0.1:9042] Failed to open connection to the non-shard-aware port: IoError(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
scylla::transport::connection_pool:::680: pool_state: "[(0, [])]"
scylla::transport::connection_pool:::692: message: [127.0.0.1:9042] Scheduling next refill in 100 ms
scylla::transport::connection_pool:::275: message: Selecting random connection
scylla::transport::topology:::438: message: Known peers: ContactPoint(ContactPoint { address: 127.0.0.1:9042, datacenter: None })
scylla::transport::topology:::491: message: Could not fetch metadata, error: IO Error: No connections in the pool; last connection failed with: IO Error: Connection refused (os error 111)
Error: IO Error: No connections in the pool; last connection failed with: IO Error: Connection refused (os error 111)
```

Did you make sure that you link an updated version of `libscylla_cpp_driver.so` while building the example?

Gor027:
> One solution to that could be a proc macro wrapping whole function body in a scoped subscriber. This macro would be placed on any function that calls Rust driver or does the logging itself.

The problem is that `with_current_subscriber` returns a future that needs to be awaited, so what about logging in non-async functions?

wprzytula:
> Did you make sure that you link an updated version of `libscylla_cpp_driver.so` while building the example?

Apparently not... Now, my output is the same as yours. Great!

wprzytula (approved):
Although the restrictions imposed by the global subscriber are indeed limiting, it is better to have working tracing with some limitations that nonworking tracing. Possible enhancements should be considered for a follow-up.

Gor027:
@Lorak-mmk @piodul what do you think?

Lorak-mmk:
If we were to merge it with current limitations:
1. This PR needs to document those limitations:
  - Being able to call `cass_set_log_level` only once
  - That this library ignores warning in `set_global_default` documentation https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html "Note: Libraries should NOT call set_global_default()! That will cause conflicts when executables try to set them later." -  so if you're using it from Rust application, you won't be able to set global default subscriber.
3. Issue needs to be open after merge. You can also put my idea for a solution in the issue (Creating macro that uses https://docs.rs/tracing/latest/tracing/subscriber/fn.with_default.html + adding `.with_current_subscriber()` to futures passed to tokio::spawn)

Gor027:
> 1. This PR needs to document those limitations:

Details about the logging limitations are documented in the README.

> * That this library ignores warning in `set_global_default` documentation https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html "Note: Libraries should NOT call set_global_default()! That will cause conflicts when executables try to set them later." -  so if you're using it from Rust application, you won't be able to set global default subscriber.

Do we plan to release the bindings as a library so that other crates can depend on it?

wprzytula:
> > - That this library ignores warning in set_global_default documentation https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html "Note: Libraries should NOT call set_global_default()! That will cause conflicts when executables try to set them later." - so if you're using it from Rust application, you won't be able to set global default subscriber.

> Do we plan to release the bindings as a library so that other crates can depend on it?

IIUC, the problem stated by @Lorak-mmk is about an (IMO) unlikely situation when someone uses the C library (driven by the bindings) using FFI in another Rust crate. But... why would anyone do that?

Lorak-mmk:
One possible scenario is using our library underneath something like https://github.com/Metaswitch/cassandra-rs

Gor027:
Ping. I have updated README about logging limitations.

piodul:
> One solution to that could be a proc macro wrapping whole function body in a scoped subscriber. This macro would be placed on any function that calls Rust driver or does the logging itself.

The CPP driver API is large and it sounds very easy to make a mistake and forget about some functions that need it. I don't see how we could test something like that, I think we need something which doesn't require per-function action.

One idea that comes to my mind is a per-file attribute macro which automatically annotates `extern "C"` functions whose names start with `cass_`.

> One possible scenario is using our library underneath something like https://github.com/Metaswitch/cassandra-rs

Maybe there is no problem? Rust libraries are linked statically, so our bindings will contain a copy of the tracing library, including the symbol for the global subscriber. If you have an application which uses tracing itself and links to our bindings, then perhaps there will be two separate instances of tracing. This is just a hypothesis, but I think it's worth checking out - perhaps there is no problem at all with the current solution.

wprzytula:
> > One solution to that could be a proc macro wrapping whole function body in a scoped subscriber. This macro would be placed on any function that calls Rust driver or does the logging itself.
> 
> The CPP driver API is large and it sounds very easy to make a mistake and forget about some functions that need it. I don't see how we could test something like that, I think we need something which doesn't require per-function action.
> 
> One idea that comes to my mind is a per-file attribute macro which automatically annotates `extern "C"` functions whose names start with `cass_`.
> 
> > One possible scenario is using our library underneath something like https://github.com/Metaswitch/cassandra-rs
> 
> Maybe there is no problem? Rust libraries are linked statically, so our bindings will contain a copy of the tracing library, including the symbol for the global subscriber. If you have an application which uses tracing itself and links to our bindings, then perhaps there will be two separate instances of tracing. This is just a hypothesis, but I think it's worth checking out - perhaps there is no problem at all with the current solution.

This is definitely promising. Are you going to check the hypothesis out?

PR #112, opened 2023-05-26
New deserialization api

### LINKED ISSUE #94: Add support for `cass_value_get_decimal` and `cass_value_get_bytes`
Add support for two missing functions: `cass_value_get_decimal` and `cass_value_get_bytes`. This functionality is dependent on Rust Driver providing ability to see "raw bytes" - unparsed response received from Scylla. This Rust Driver issue tracks the changes that will allow such functionality: https://github.com/scylladb/scylla-rust-driver/issues/462

Gor027:
## Pre-review checklist

Fixes #94 

This PR Integrates the new deserialization API from the Rust driver which is a [WIP](https://github.com/scylladb/scylla-rust-driver/pull/665).
It is dependent upon the execution profiles' PR #107 and can be merged only after the latter is already done.
The lazy deserialization allows to also implement binder makers for decimal and duration types. The `test_cassandra_types` tests for these types are also enabled.


scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -96,8 +97,8 @@ impl From<&BadKeyspaceName> for CassError {
     }
 }
 
-impl From<ParseError> for CassError {
-    fn from(error: ParseError) -> Self {
+impl From<&ParseError> for CassError {
+    fn from(error: &ParseError) -> Self {
```
wprzytula: What is the reason behind this change? Do you want to avoid consuming the error somewhere?

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -149,6 +169,18 @@ impl CassErrorMessage for ParseError {
...
+
+impl CassErrorMessage for FirstRowError {
+    fn msg(&self) -> String {
+        self.to_string()
+    }
+}
+
 impl CassErrorMessage for str {
     fn msg(&self) -> String {
         self.to_string()
```
wprzytula, line 186: I'm wondering why do we even need this `msg()` method. Isn't it virtually equivalent to `ToString::to_string()` functionality? And `ToString::to_string()` is blanket implemented for every `T: Display`, which means that we only have to make `CassErrorMessage` require `Display` and we will be able to remove `msg()` alltogether.

scylla-rust-wrapper/src/future.rs
```diff
@@ -254,7 +254,7 @@ pub unsafe extern "C" fn cass_future_tracing_id(
     tracing_id: *mut CassUuid,
 ) -> CassError {
     ptr_to_ref(future).with_waited_result(|r: &mut CassFutureResult| match r {
-        Ok(CassResultValue::QueryResult(result)) => match result.metadata.tracing_id {
+        Ok(CassResultValue::QueryResult(result)) => match result.result.tracing_id() {
```
wprzytula: Consider renaming the outer `result` into `cass_result`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -138,8 +139,8 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
 
             result_iterator.position = Some(new_pos);
 
-            match &result_iterator.result.rows {
-                Some(rs) => (new_pos < rs.len()) as cass_bool_t,
+            match result_iterator.result.result.rows_num() {
+                Some(rs) => (new_pos + 1 < rs) as cass_bool_t, // TODO: add next row decoding
```
wprzytula: Why is this `+1` added? Isn't new `rs` (result of `rows_num()`) equivalent to old `rs.len()` ?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -858,16 +859,21 @@ pub unsafe extern "C" fn cass_row_get_column_by_name_n(
     let row_from_raw = ptr_to_ref(row);
     let mut name_str = ptr_to_cstr_n(name, name_length).unwrap();
     let mut is_case_sensitive = false;
+    let result = row_from_raw.result.upgrade().unwrap();
```
wprzytula: Why is this `unwrap()` safe? Please write in a comment.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -858,16 +859,21 @@ pub unsafe extern "C" fn cass_row_get_column_by_name_n(
...
-        .col_specs
+    if col_specs.is_none() {
+        return std::ptr::null();
+    }
+
+    return col_specs
+        .unwrap()
         .iter()
         .enumerate()
         .find(|(_, spec)| {
```
wprzytula: Can we drop this explicit `return` in the last expression of the function?

wprzytula: Also, please make this more Rusty: instead of excluding the `None` case by returning early and later calling  `unwrap()`, use pattern matching. In case of this particular piece of code, I'd recommend such a refactor:
```rust
col_specs
    .and_then(|col_specs| col_specs
        .iter()
        .enumerate()
        .find(|(_, spec)| {
            is_case_sensitive && spec.name == name_str
                || !is_case_sensitive && spec.name.eq_ignore_ascii_case(name_str)
        })
        .map(|(index, _)| {
            if let Some(value) = row_from_raw.columns.get(index) {
                value as *const CassValue
            } else {
                std::ptr::null()
            }
        }))
    .unwrap_or(std::ptr::null())
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -892,12 +898,13 @@ pub unsafe extern "C" fn cass_result_column_name(
 ) -> CassError {
     let result_from_raw = ptr_to_ref(result);
     let index_usize: usize = index.try_into().unwrap();
+    let col_specs = result_from_raw.result.column_specs();
 
-    if index_usize >= result_from_raw.metadata.col_specs.len() {
+    if col_specs.is_none() || index_usize >= col_specs.unwrap().len() {
         return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
     }
```
wprzytula: Is the error relevant for the case of `None` `col_specs`? What does it mean that `col_specs` are `None`?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -892,12 +898,13 @@ pub unsafe extern "C" fn cass_result_column_name(
...
 
-    let column_spec: &ColumnSpec = result_from_raw.metadata.col_specs.get(index_usize).unwrap();
+    let column_spec: &ColumnSpec = col_specs.unwrap().get(index_usize).unwrap();
```
wprzytula: Please harden this by avoiding `unwrap()`ping:
```rust
let col_specs = if let Some(specs) = result_from_raw.result.column_specs() {
    specs
} else {
    return CassError::SOME_ERROR;
};
// use `col_specs` without need to `unwrap()`
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1197,26 +1204,28 @@ pub unsafe extern "C" fn cass_value_secondary_sub_type(
 pub unsafe extern "C" fn cass_result_row_count(result_raw: *const CassResult) -> size_t {
     let result = ptr_to_ref(result_raw);
 
-    if result.rows.as_ref().is_none() {
-        return 0;
+    match result.result.rows_num().as_ref() {
+        Some(rows_count) => *rows_count as size_t,
+        None => 0,
```
wprzytula: Or, more consisely:
```rust
result.result.rows_num().as_ref().copied().unwrap_or(0) as size_t
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -94,3 +94,14 @@ macro_rules! make_c_str {
```
wprzytula, line 94: Ad 72e8da9a931cd29952472c3b37d5a351e47a6385:
in commit message: `means to stores` -> `means to store`

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -94,3 +94,14 @@ macro_rules! make_c_str {
...
 pub(crate) use make_c_str;
+
+macro_rules! unwrap_or_return {
+    ( $e:expr ) => {
+        match $e {
+            Ok(x) => x,
+            Err(_) => return false,
+        }
+    };
+}
```
wprzytula: Naming: either rename this macro to `unwrap_or_return_false`, or make it accept the return value in the second parameter (i.e. make its parameter list look like this: `( $e:expr, $retval:expr )`

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -123,6 +124,33 @@ pub enum CassIterator {
     CassViewMetaIterator(CassViewMetaIterator),
 }
 
+fn decode_next_row(result: &'static CassResult, row: &mut Option<CassRow>) -> bool {
+    if let Some(row) = row {
+        // Errors are ignored, but logging them may come handy in the future.
```
wprzytula: come handy -> come in handy

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -123,6 +124,33 @@ pub enum CassIterator {
...
+        let mut rows_iter = unwrap_or_return!(result.result.rows::<ColumnIterator>());
+        let next_cols_iter = unwrap_or_return!(rows_iter.next().unwrap());
+
+        for (i, raw_col) in next_cols_iter.into_iter().enumerate() {
+            let raw_col = unwrap_or_return!(raw_col);
+            let value_type = get_column_type(&raw_col.spec.typ);
+            let frame_slice = raw_col.slice;
+            let is_null = frame_slice.is_none() || raw_col.slice.unwrap().is_empty();
```
wprzytula: You can avoid `unwrap()` with:
```rust
let is_null = frame_slice.map_or(true, |f| f.is_empty());
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -131,17 +159,20 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
 // After creating an iterator we have to call next() before accessing the value
 #[no_mangle]
 pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass_bool_t {
-    let mut iter = ptr_to_ref_mut(iterator);
+    let iter: &mut CassIterator = ptr_to_ref_mut(iterator);
 
-    match &mut iter {
+    match iter {
```
wprzytula, line 276: This change seems to be a logical no-op (which is good, because otherwise it would be suspicious): the type inferred in the previous version was `&mut CassIterator` anyway, so the `mut` and `match &mut` were simply redundant.

wprzytula: (this is just for other reviewers to clarify)

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -131,17 +159,20 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
             result_iterator.position = Some(new_pos);
 
             match result_iterator.result.result.rows_num() {
-                Some(rs) => (new_pos + 1 < rs) as cass_bool_t, // TODO: add next row decoding
-                None => false as cass_bool_t,
+                Some(rs) if new_pos < rs => {
+                    decode_next_row(result_iterator.result.as_ref(), &mut result_iterator.row)
+                        as cass_bool_t
+                }
+                _ => false as cass_bool_t,
```
wprzytula: The `+ 1` has been removed in this commit, so I believe its addition in a previous commit was a bug.

scylla-rust-wrapper/src/session.rs
```diff
@@ -301,21 +302,55 @@ pub unsafe extern "C" fn cass_session_execute(
...
             result: weak_res.clone(),
-            columns: Vec::new(),
+            columns: Vec::with_capacity(cols_num),
         };
         cass_result.first_row = Some(first_row);
 
         cass_result
-    })
+    });
+    decode_first_row(Arc::as_ptr(&cass_result) as *mut CassResult);
```
wprzytula, line 314: Using `get_mut()` here would be arguably safer, and it would convey the intention better:
```rust
decode_first_row(Arc::get_mut(&mut cass_result).unwrap());
```

scylla-rust-wrapper/src/session.rs
```diff
@@ -301,21 +302,55 @@ pub unsafe extern "C" fn cass_session_execute(
...
+
+    if result.first_row.is_none() {
+        return false;
+    }
+
+    for raw_col in rows_iter.into_iter() {
+        let raw_col = unwrap_or_return!(raw_col);
+        let value_type = get_column_type(&raw_col.spec.typ);
+        let frame_slice = raw_col.slice;
+        let is_null = frame_slice.is_none() || raw_col.slice.unwrap().is_empty();
```
wprzytula: ditto: use `map_or()`

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -957,224 +963,191 @@ pub unsafe extern "C" fn cass_value_data_type(value: *const CassValue) -> *const
...
+);
+
+cass_value_get_numeric_type!(
+    cass_value_get_int64,
+    cass_int64_t,
+    CassValueType::CASS_VALUE_TYPE_BIGINT
+        | CassValueType::CASS_VALUE_TYPE_COUNTER
+        | CassValueType::CASS_VALUE_TYPE_TIMESTAMP
+        | CassValueType::CASS_VALUE_TYPE_TIME,
+    ColumnType::BigInt // or `Counter` types can be deserialized as i64 in Rust driver
```
wprzytula, line 1343: Can you elaborate on this comment?
Do you mean that both `BigInt` and `Counter` can be deserialized as `i64`?
If so, you could probably mention `Timestamp` and `Time` as well, as I can see them above in the pattern.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -957,224 +963,191 @@ pub unsafe extern "C" fn cass_value_data_type(value: *const CassValue) -> *const
...
+        std::ptr::write(output, val.into());
+        CassError::CASS_OK
+    }
+);
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_value_is_null(value: *const CassValue) -> cass_bool_t {
+    if value.is_null() {
+        return true as cass_bool_t;
+    }
```
wprzytula: Shorter:
```rust
value.as_ref().map_or(true, |val| val.is_null) as cass_bool_t
```

scylla-rust-wrapper/src/future.rs
```diff
@@ -289,6 +289,9 @@ mod tests {
         let wrapped_cass_fut = PtrWrapper(cass_fut);
         unsafe {
             let handle = thread::spawn(move || {
+                // dummy let to force the closure to capture the entire variable
+                // Migration guide for 2021 edition: https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html#migration
+                let _ = &wrapped_cass_fut;
```
wprzytula, line 294: Interesting change! @Lorak-mmk @piodul @havaker @cvybhu

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -141,11 +136,17 @@ fn decode_next_row(result: &'static CassResult, row: &mut Option<CassRow>) -> bo
             let value_type = get_column_type(&raw_col.spec.typ);
             let frame_slice = raw_col.slice;
             let is_null = frame_slice.is_none() || raw_col.slice.unwrap().is_empty();
+            let mut count = 0;
+
+            if let (Some(frame), true) = (frame_slice, value_type.is_collection()) {
+                let mut mem = frame.as_slice();
+                count = unwrap_or_return!(types::read_int_length(&mut mem));
+            }
```
wprzytula: Unnecessary `mut`. You can take advantage of Rust's amazing block evaluation:
```rust
  let count = if let (Some(frame), true) = (frame_slice, value_type.is_collection()) {
      let mut mem = frame.as_slice();
      unwrap_or_return!(types::read_int_length(&mut mem))
  } else {
      0
  };
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -317,29 +325,17 @@ pub unsafe extern "C" fn cass_iterator_get_value(
...
+            if collection_iterator.value.is_some() =>
+        {
+            collection_iterator.value.as_ref().unwrap()
         }
-
-        return value.unwrap() as *const CassValue;
+        CassIterator::CassTupleIterator(tuple_iterator) if tuple_iterator.value.is_some() => {
+            tuple_iterator.value.as_ref().unwrap()
+        }
+        _ => std::ptr::null(), // null is returned if value in iterator is not set
```
wprzytula: Please avoid `unwrap()`s:
```rust
match iter {
    CassIterator::CassCollectionIterator(CassCollectionIterator { value: Some(value), .. }) => value,
    CassIterator::CassTupleIterator(CassTupleIterator { value: Some(value), .. }) => value,
    _ => std::ptr::null(), // null is returned if value in iterator is not set
}
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -348,25 +344,16 @@ pub unsafe extern "C" fn cass_iterator_get_map_key(
...
-
-        if entry.is_none() {
-            return std::ptr::null();
+    match iter {
+        CassIterator::CassMapIterator(map_iterator) => {
+            assert!(map_iterator
+                .position
+                .map(|pos| pos < map_iterator.count)
+                .is_some()); // assertion copied from c++ driver
+            map_iterator.key.as_ref().unwrap()
```
wprzytula: Why is this `unwrap()` safe? Write in a comment.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -375,25 +362,16 @@ pub unsafe extern "C" fn cass_iterator_get_map_value(
...
-
-        if entry.is_none() {
-            return std::ptr::null();
+    match iter {
+        CassIterator::CassMapIterator(map_iterator) => {
+            assert!(map_iterator
+                .position
+                .map(|pos| pos < map_iterator.count)
+                .is_some()); // assertion copied from c++ driver
+            map_iterator.value.as_ref().unwrap()
```
wprzytula: ditto: justify `unwrap()`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -404,31 +382,21 @@ pub unsafe extern "C" fn cass_iterator_get_user_type_field_name(
...
-            }
-            None => return CassError::CASS_ERROR_LIB_BAD_PARAMS,
+    match iter {
+        CassIterator::CassUdtIterator(udt_iterator) if udt_iterator.field_name.is_some() => {
+            assert!(udt_iterator
+                .position
+                .map(|pos| pos < udt_iterator.count)
+                .is_some()); // assertion copied from c++ driver
+            write_str_to_c(
+                udt_iterator.field_name.as_ref().unwrap().as_str(),
```
wprzytula: ditto: justify `unwrap()`.

wprzytula: Cleaner:
```rust
match iter {
    CassIterator::CassUdtIterator(CassUdtIterator { field_name: Some(field_name), count, position, ..}) => {
        assert!(position
            .map(|pos| pos < *count)
            .is_some()); // assertion copied from c++ driver
        write_str_to_c(
            field_name.as_str(),
            name,
            name_length,
        );
        CassError::CASS_OK
    }
    _ => CassError::CASS_ERROR_LIB_BAD_PARAMS,
}
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -437,29 +405,16 @@ pub unsafe extern "C" fn cass_iterator_get_user_type_field_value(
...
-            },
-            None => std::ptr::null(),
-        };
+    match iter {
+        CassIterator::CassUdtIterator(udt_iterator) if udt_iterator.field_value.is_some() => {
+            assert!(udt_iterator
+                .position
+                .map(|pos| pos < udt_iterator.count)
+                .is_some()); // assertion copied from c++ driver
+            udt_iterator.field_value.as_ref().unwrap()
```
wprzytula: Please refactor this analogously to my suggestion for `cass_iterator_get_user_type_field_name()` (avoiding `unwrap()`s).

scylla-rust-wrapper/src/session.rs
```diff
@@ -339,11 +340,17 @@ unsafe fn decode_first_row(result: *mut CassResult) -> bool {
         let value_type = get_column_type(&raw_col.spec.typ);
         let frame_slice = raw_col.slice;
         let is_null = frame_slice.is_none() || raw_col.slice.unwrap().is_empty();
+        let mut count = 0;
+
+        if let (Some(frame), true) = (frame_slice, value_type.is_collection()) {
+            let mut mem = frame.as_slice();
+            count = unwrap_or_return!(types::read_int_length(&mut mem));
+        }
```
wprzytula: Unnecessary `mut`.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -238,6 +238,13 @@ impl CassDataType {
         }
     }
 
+    pub fn get_tuple_types(&self) -> &Vec<Arc<CassDataType>> {
+        match self {
+            CassDataType::Tuple(type_defs) => type_defs,
+            _ => panic!("Cannot get tuple out of non-tuple data type"),
+        }
+    }
+
```
wprzytula, line 247: This panic may unwind into caller C code, which means UB. What is our policy in this regard? @Lorak-mmk @piodul

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -133,31 +158,61 @@ fn decode_next_row(result: &'static CassResult, row: &mut Option<CassRow>) -> bo
...
+                slice: raw_col.slice,
             };
-            // Below assignment is safe from out of bounds panic as,
-            // first [decode_first_row] call already initialized the columns vec
-            row.columns[i] = cass_value;
+            let cass_value = decode_value(raw_value, &raw_col.spec.typ);
+            match cass_value {
+                Some(value) => {
+                    // Below assignment is safe from out of bounds panic as,
+                    // first [decode_first_row] call already initialized the columns vec
```
wprzytula: panic as, first -> panic, as first

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -133,31 +158,61 @@ fn decode_next_row(result: &'static CassResult, row: &mut Option<CassRow>) -> bo
...
     true
 }
 
+pub fn decode_value(
+    raw_value: RawValue<'static>,
+    val_type: &'static ColumnType,
+) -> Option<CassValue> {
+    let data_type = get_column_type(val_type);
+    let frame_slice = raw_value.slice;
+    let is_null = frame_slice.is_none() || raw_value.slice.unwrap().is_empty();
```
wprzytula: Please avoid `unwrap()` here.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -194,30 +249,88 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
...
+                    let value = decode_value(raw_value, value_type);
+                    map_iterator.key = key;
+                    map_iterator.value = value;
+
+                    return true as cass_bool_t;
+                }
+            }
 
-            (new_pos < map_iterator.count) as cass_bool_t
+            false as cass_bool_t
```
wprzytula: `false` being after an `if` suggests that this is the default path.
Instead, I suggest:
```rust
if new_pos < map_iterator.count {
    let raw_value = map_iterator.map_iterator.next().unwrap();
    if let Ok((raw_key, raw_value)) = raw_value {
        let key_type = raw_key.spec;
        let key = decode_value(raw_key, key_type);
        let value_type = raw_value.spec;
        let value = decode_value(raw_value, value_type);
        map_iterator.key = key;
        map_iterator.value = value;

        true
    }
} else {
  false
}  as cass_bool_t
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -194,30 +249,88 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
...
+                    let value = decode_value(raw, field_type);
+                    udt_iterator.field_value = value;
+                    udt_iterator.field_name = Some(name.clone());
+
+                    return true as cass_bool_t;
+                }
+            }
 
-            (new_pos < udt_iterator.count) as cass_bool_t
+            false as cass_bool_t
```
wprzytula: ditto (`false` after `if`)

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -639,8 +762,11 @@ pub unsafe extern "C" fn cass_iterator_from_tuple(value: *const CassValue) -> *m
 
     if !tuple.is_null && tuple.value_type.is_tuple() {
         let item_count = tuple.count;
+        let column_type = tuple.column_type;
+        let sequence_iterator =
+            SequenceIterator::new(column_type, item_count, tuple.frame_slice.unwrap()); // safe to unwrap as is_null is false
```
wprzytula: I'm unhappy with all these `unwrap()`s. Perhaps we can do that another way?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -680,8 +809,13 @@ pub unsafe extern "C" fn cass_iterator_fields_from_user_type(
 
     if !udt.is_null && udt.value_type.is_user_type() {
         let item_count = udt.count;
+        let fields = match udt.column_type {
+            ColumnType::UserDefinedType { field_types, .. } => field_types.as_slice(),
+            _ => panic!("Unexpected column type for map collection"),
+        };
+        let udt_iterator = UdtIterator::new(fields, udt.frame_slice.unwrap()); // safe to unwrap as is_null is false
         let iterator = CassUdtIterator {
```
wprzytula: ditto

scylla-rust-wrapper/src/session.rs
```diff
@@ -337,24 +336,17 @@ unsafe fn decode_first_row(result: *mut CassResult) -> bool {
...
-
-        result.first_row.as_mut().unwrap().columns.push(cass_value);
+        let cass_value = decode_value(raw_value, &raw_col.spec.typ);
+        match cass_value {
+            Some(value) => {
+                result.first_row.as_mut().unwrap().columns.push(value);
+            }
+            _ => return false,
+        }
     }
```
wprzytula: This is a great use case for `if let`:
```
if let Some(value) = cass_value {
    result.first_row.as_mut().unwrap().columns.push(value);
} else {
    return false;
}
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -439,11 +467,20 @@ pub unsafe extern "C" fn cass_iterator_get_value(
...
+        CassIterator::CassCollectionIterator(collection_iterator) => match collection_iterator {
+            CassCollectionIterator::SequenceIterator(seq_iter) => seq_iter.value.as_ref().unwrap(),
+            CassCollectionIterator::SeqMapIterator(seq_map_iter) => {
+                if let Some(pos) = seq_map_iter.position {
+                    if pos % 2 == 0 && seq_map_iter.key.is_some() {
+                        return seq_map_iter.key.as_ref().unwrap();
+                    } else if pos % 2 == 0 && seq_map_iter.value.is_some() {
+                        return seq_map_iter.value.as_ref().unwrap();
+                    }
+                }
```
wprzytula: It's unclear to me why `pos` is taken modulo 2.

scylla-rust-wrapper/src/binding.rs
```diff
@@ -137,11 +137,6 @@ macro_rules! make_appender {
     }
 }
 
-// TODO: Types for which binding is not implemented yet:
-// custom - Not implemented in Rust driver?
-// decimal
-// duration - DURATION not implemented in Rust Driver
-
```
wprzytula: What about `custom`? The TODO was removed but this case is probably not solved.

wprzytula (changes requested):

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -131,17 +159,20 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
             match result_iterator.result.result.rows_num() {
-                Some(rs) => (new_pos + 1 < rs) as cass_bool_t, // TODO: add next row decoding
-                None => false as cass_bool_t,
+                Some(rs) if new_pos < rs => {
+                    decode_next_row(result_iterator.result.as_ref(), &mut result_iterator.row)
+                        as cass_bool_t
+                }
+                _ => false as cass_bool_t,
             }
         }
```
wprzytula: Alternatively:
```rust
match result_iterator.result.result.rows_num() {
  Some(rs) if new_pos < rs => {
      decode_next_row(result_iterator.result.as_ref(), &mut result_iterator.row)
  }
  _ => false ,
} as cass_bool_t

**Review by wprzytula — CHANGES_REQUESTED** [review]
A general idea to make the logic more type-safe (which Rust promotes):
replace groups of `Option`s with custom types that encode the logic inside themselves (their names, variants, etc.). E.g. in case of `CassResultIterator`, `position` is `None` iff `row` is `None`. This dependency could be represented as an enum:

```rust
pub struct CassResultIterator {
    result: Arc<CassResult>,
    curr_row_info: CassResultIteratorCurrentRowInfo,
}

enum CassResultIteratorCurrentRowInfo {
    NoRows,
    Rows {
        current_row: CassRow,
        position: usize,
    }
}

```

scylla-rust-wrapper/src/cass_error.rs
Gor027: Yes. See `From` impl for `RowsError` and `FirstRowError`.

scylla-rust-wrapper/src/cass_error.rs
Gor027: The mapping of the Rust driver errors to the `CassError` enum values is not so obvious, and some of the mappings are not even correct (e.g. `TimeoutError => CassError::CASS_ERROR_LAST_ENTRY`). Having said that the msg trait may be helpful in the future for modifying certain error messages. I guess the trait was introduced with that intuition, too. I would not change it in this PR, maybe better to leave it for now.

scylla-rust-wrapper/src/future.rs
Gor027: Done.

scylla-rust-wrapper/src/query_result.rs
Gor027: You are right, fixed it.

scylla-rust-wrapper/src/query_result.rs
Gor027: The lifetime of `CassRow` is bound to `CassResult`, so it will be freed only when the `CassResult` is freed. I will add a short comment about that.

scylla-rust-wrapper/src/query_result.rs
Gor027: The `column_specs()` function may return `None` if "raw" rows of a single query are empty.

scylla-rust-wrapper/src/query_result.rs
Gor027: I think you will need to wrap the whole match block in `()` to make it work.

scylla-rust-wrapper/src/session.rs
Gor027: Thanks for the hint, fixed it.

scylla-rust-wrapper/src/query_result.rs
Gor027: Yes, both can be deserialized as `i64` in the Rust driver, but `Timestamp` and `Time` deserialization as `i64` is not implemented.

scylla-rust-wrapper/src/query_result.rs
Gor027: If `cass_iterator_next` succeeded it is safe to unwrap, it is wrapped in an `Option` for safe initialization purposes.

scylla-rust-wrapper/src/query_result.rs
Gor027: If `cass_iterator_next` succeeded it is safe to unwrap, it is wrapped in an `Option` for safe initialization purposes.

scylla-rust-wrapper/src/query_result.rs
Gor027: Done.

scylla-rust-wrapper/src/query_result.rs
Gor027: Done.

scylla-rust-wrapper/src/query_result.rs
Gor027: I think the whole if-else block should be wrapped in `()` in order to make the casting work.

scylla-rust-wrapper/src/query_result.rs
Gor027: I did not notice I check the same condition in the else branch. It is fixed now. If you iterate sequentially over a map, an even position means you will get a key and an odd position means you will get a value:
```
0   | 1     | 2   |....
key | value | key | ....
```

scylla-rust-wrapper/src/binding.rs
Gor027: As it is not implemented in the Rust driver (not even in ScyllaDB?), I have reverted the TODO comment.

scylla-rust-wrapper/src/session.rs
Gor027: I did not carefully read the documentation of `get_mut` which returns `None` in case there are `Weak` pointers to the same allocation which is the case here. Reverted back to `Arc::as_ptr()`.

scylla-rust-wrapper/src/query_result.rs
wprzytula: Or even (which can be a bit surprising):
```rust
let count;
if let (Some(frame), true) = (frame_slice, value_type.is_collection()) {
      let mut mem = frame.as_slice();
      count = unwrap_or_return!(types::read_int_length(&mut mem));
  } else {
      count = 0;
  };
```

wprzytula: I prefer the first option, though.

scylla-rust-wrapper/src/query_result.rs
wprzytula: What do you mean? What is an error that you get?

Gor027: The alternative that you suggested is not a valid expression until you wrap it into parenthesis or add an explicit `return`:

```rust
(match result_iterator.result.result.rows_num() {
  Some(rs) if new_pos < rs => {
      decode_next_row(result_iterator.result.as_ref(), &mut result_iterator.row)
  }
  _ => false ,
}) as cass_bool_t
```
or,
```rust
return match result_iterator.result.result.rows_num() {
  Some(rs) if new_pos < rs => {
      decode_next_row(result_iterator.result.as_ref(), &mut result_iterator.row)
  }
  _ => false ,
} as cass_bool_t;
```

Gor027:
@wprzytula @Lorak-mmk Ping for review.

scylla-rust-wrapper/src/cass_types.rs
wprzytula: ping @Lorak-mmk @piodul

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -237,17 +424,12 @@ pub unsafe extern "C" fn cass_iterator_get_row(iterator: *const CassIterator) ->
...
-            Some(row) => row,
-            None => return std::ptr::null(),
-        };
-
-        return row;
+        if let Some(rows_count) = result_iterator.result.result.rows_num() {
+            return match &result_iterator.row {
+                Some(row) if iter_position < rows_count => row,
+                _ => return std::ptr::null(),
+            };
```
wprzytula: This would be tidier if both returns were on the same level:
```rust
return match &result_iterator.row {
    Some(row) if iter_position < rows_count => row,
    _ => std::ptr::null(),
};
```

scylla-rust-wrapper/src/cass_types.rs
wprzytula: Isn't it the case that we should use `panic='abort'` ?

Gor027:
> A general idea to make the logic more type-safe (which Rust promotes): replace groups of `Option`s with custom types that encode the logic inside themselves (their names, variants, etc.). E.g. in case of `CassResultIterator`, `position` is `None` iff `row` is `None`. This dependency could be represented as an enum:
> 
> ```rust
> pub struct CassResultIterator {
>     result: Arc<CassResult>,
>     curr_row_info: CassResultIteratorCurrentRowInfo,
> }
> 
> enum CassResultIteratorCurrentRowInfo {
>     NoRows,
>     Rows {
>         current_row: CassRow,
>         position: usize,
>     }
> }
> ```

I like the idea to express the state of the fields of iterators through enums, but that will make this PR a lot more complicated. Do you mind doing this refactor in another PR? I will open an issue to not forget it.

wprzytula:
> > A general idea to make the logic more type-safe (which Rust promotes): replace groups of `Option`s with custom types that encode the logic inside themselves (their names, variants, etc.). E.g. in case of `CassResultIterator`, `position` is `None` iff `row` is `None`. This dependency could be represented as an enum:
> > ```rust
> > pub struct CassResultIterator {
> >     result: Arc<CassResult>,
> >     curr_row_info: CassResultIteratorCurrentRowInfo,
> > }
> > 
> > enum CassResultIteratorCurrentRowInfo {
> >     NoRows,
> >     Rows {
> >         current_row: CassRow,
> >         position: usize,
> >     }
> > }
> > ```
> 
> I like the idea to express the state of the fields of iterators through enums, but that will make this PR a lot more complicated. Do you mind doing this refactor in another PR? I will open an issue to not forget it.

I'd prefer to have the refactor done in this PR. You can put it in a separate commit.

scylla-rust-wrapper/src/query_result.rs
Gor027: Done.

Gor027:
> A general idea to make the logic more type-safe (which Rust promotes): replace groups of `Option`s with custom types that encode the logic inside themselves (their names, variants, etc.). E.g. in case of `CassResultIterator`, `position` is `None` iff `row` is `None`. This dependency could be represented as an enum:
> 
> ```rust
> pub struct CassResultIterator {
>     result: Arc<CassResult>,
>     curr_row_info: CassResultIteratorCurrentRowInfo,
> }
> 
> enum CassResultIteratorCurrentRowInfo {
>     NoRows,
>     Rows {
>         current_row: CassRow,
>         position: usize,
>     }
> }
> ```

I have applied the idea in separate commits. Perhaps this approach is more expressive and makes it easy to understand the iterator's state, although it looks overhead to me.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -68,15 +68,20 @@ pub struct CassValue {
     pub column_type: &'static ColumnType,
 }
 
+enum CassIteratorStateInfo<T> {
+    NoValue,
+    ValueNoPosition { value: T },
+    PositionNoValue { position: usize },
+    Value { value: T, position: usize },
+}
```
wprzytula, line 76: The semantics is unclear: Neither do names of the variants describe the semantics, nor are there any comments with such explanation.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -215,6 +220,30 @@ pub fn decode_value(
...
+                position: position + 1,
+            }
+        }
+        CassIteratorStateInfo::ValueNoPosition { value } => {
+            CassIteratorStateInfo::Value { value, position: 0 }
+        }
+        CassIteratorStateInfo::NoValue => CassIteratorStateInfo::PositionNoValue { position: 0 },
+    }
+}
+
```
wprzytula: What about making this a method and simply mutate instead of returning?

```suggestion
impl<T> CassIteratorStateInfo<T> {
    /// Update iterator's state and return new state info.
    /// Increments an existing position, otherwise sets it to 0.
    fn advance(
        &mut self,
    ) {
        // Store a dummy NoValue temporarily as we cannot move out StateInfo fields.
        let old_state_info = std::mem::replace(self, CassIteratorStateInfo::NoValue);
        *self = match old_state_info {
            CassIteratorStateInfo::Value { value, position } => CassIteratorStateInfo::Value {
                value,
                position: position + 1,
            },
            CassIteratorStateInfo::PositionNoValue { position } => {
                CassIteratorStateInfo::PositionNoValue {
                    position: position + 1,
                }
            }
            CassIteratorStateInfo::ValueNoPosition { value } => {
                CassIteratorStateInfo::Value { value, position: 0 }
            }
            CassIteratorStateInfo::NoValue => CassIteratorStateInfo::PositionNoValue { position: 0 },
        };
    }
}


```

wprzytula: Then using it is as simple as
```rust
result_iterator.state_info.advance();
```

scylla-rust-wrapper/src/query_result.rs
wprzytula: The question is: what behaviour do these states represent? What properties are represented by such?

wprzytula (commented):
I'm very happy to see the enum-based iterator states. This is much more readable than combinations of `Option`s.

scylla-rust-wrapper/src/query_result.rs
Gor027: Applied the suggestion.

scylla-rust-wrapper/src/query_result.rs
wprzytula: Perhaps this is obvious provided one knows the inner working of Cpp driver. I believe, however, that we should strive towards having bindings' code documented well enough so that this becomes unnecessary. After all, nobody really wants to peruse its code.

Gor027: The enum is supposed to represent the state of two optional variables: position and value. Let me explain the details and feel free to suggest better naming for the variants.
{NoValue, ValueNoPosition}: The iterator may be created empty (without a value and a position) or with an initial value, e.g. `RowIterator`.
{PositionNoValue, Value}: A call to `cass_iterator_next` will first increment the position or initialize it to 0 while the value may not be present yet. The first call to `cass_iterator_next` is supposed to make the state "complete", with present value and position.

Gor027: Maybe it is better to distinguish the case when the position exists and the value is not and in that case `advance()` should fail (maybe panic as the value cannot be retrieved from the iterator on a certain position).

Gor027: Fixed it.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -68,15 +68,44 @@ pub struct CassValue {
...
+    fn advance(&mut self) {
+        // Store a dummy NoValue temporarily as we cannot move out StateInfo fields.
+        let old_state_info = std::mem::replace(self, CassIteratorStateInfo::NoValue);
+        *self = match old_state_info {
+            CassIteratorStateInfo::Value { value, position } => CassIteratorStateInfo::Value {
+                value,
+                position: position + 1,
+            },
+            CassIteratorStateInfo::PositionNoValue { .. } => {
+                panic!("Cannot advance the iterator. Value is empty!")
```
wprzytula, line 90: Are we even allowed to panic in bindings? I still haven't got any answer from @piodul and @Lorak-mmk, and I believe that letting an unwinding panic cross the C/Rust boundary is awful UB. IIUC, we are only allowed to panic if we call `catch_unwind()` on any path from bindings' entrypoint to any panic.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -282,109 +301,130 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
...
+                                let value_type = raw_value.spec;
+                                let new_value = decode_value(raw_value, value_type);
+
+                                if let (Some(k), Some(v)) = (new_key, new_value) {
+                                    seq_map_iterator.state_info.update_value((k, v));
+                                    return true as cass_bool_t;
+                                }
+                            }
                         }
+                        return true as cass_bool_t;
```
wprzytula: Isn't the first `return` redundant?

Gor027: Good catch! When `if let` evaluates to false, we still return true which is wrong.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -282,109 +301,132 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
...
+                                };
+                            }
                         }
+                        return true as cass_bool_t;
                     }
-
-                    return true as cass_bool_t;
                 }
 
                 false as cass_bool_t
```
wprzytula: As there are a lot of nested `if let`s, please take use of `else` for each.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -282,109 +301,132 @@ pub unsafe extern "C" fn cass_iterator_next(iterator: *mut CassIterator) -> cass
...
+                            if let Some(val) = new_value {
+                                tuple_iterator.state_info.update_value(val);
+                                return true as cass_bool_t;
+                            }
+                        }
                     }
                 }
             }
 
             false as cass_bool_t
```
wprzytula: ditto

scylla-rust-wrapper/src/query_result.rs
Gor027: Basically, we use aborting panics in bindings, those are not unwinding panics and hence cannot be even caught by `catch_unwind` if I am not mistaken.

wprzytula: I can't find any code in our codebase that would turn panics into aborting. Can you?

scylla-rust-wrapper/src/query_result.rs
Gor027: I think that will add a lot of redundant true/false. The current version is simpler. WDYT?

scylla-rust-wrapper/src/query_result.rs
Gor027: For example, `cass_types:244` where for a wrong match arm Rust will stop the execution and refuse to continue.

.github/workflows/cassandra.yml
```diff
@@ -52,9 +52,6 @@ jobs:
 :LoggingTests.*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :PreparedTests.Integration_Cassandra_FailFastWhenPreparedIDChangesDuringReprepare\
-:*5.Integration_Cassandra_*\
-:*19.Integration_Cassandra_*\
-:*7.Integration_Cassandra_*\
 :SslTests.Integration_Cassandra_ReconnectAfterClusterCrashAndRestart\
```
wprzytula, line 55: Why are these enabled now? (IIUC)

scylla-rust-wrapper/src/query_result.rs
wprzytula: Hmm, how can I know that such behaviour occur?
Moreover, aborting on unrecognized pattern is something completely different that aborting on explicit panic. Isn't it?

.github/workflows/cassandra.yml
Gor027: Cause lazy deserialization allows to append, bind and get decimal, duration and raw bytes from the value. These tests were disabled cause appropriate functions were not implemented (and could not have been properly implemented before lazy deserialization).

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -128,48 +271,166 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+
+                                return if let (Some(k), Some(v)) = (new_key, new_value) {
+                                    seq_map_iterator.state_info.update_value((k, v));
+                                    true as cass_bool_t
+                                } else {
+                                    false as cass_bool_t
+                                };
+                            }
+                        }
+                        return true as cass_bool_t;
```
wprzytula: So the `if let` at `336` should return `true` if `raw_value` is `Err` ? I really can't grasp it. Perhaps more comments could explain these `if`s better.

Gor027: No, you are right it should not return true.

wprzytula: Then please, add explicit `else`s, even if they seen redundant. And comments explaining why so.

Gor027: Sequential iteration over a map was a tricky case, as the Rust driver deserializes in pairs and while iterating only on even positions (key, value) should be deserialized. Fixed it.

wprzytula (changes requested):
Looking at the maze of `if`s, I'm convinced that a set of unit tests for deserialization is absolutely required.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -128,48 +271,168 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+                            } else {
+                                false as cass_bool_t
+                            };
+                        }
+                        return true as cass_bool_t;
+                    }
+                }
 
-            collection_iterator.position = Some(new_pos);
+                false as cass_bool_t
```
wprzytula: Last idea to make it clearer for the future readers, and I won't push this anymore: for every return (either true or false), explain shortly why it is that value.

Gor027: I applied this suggestion. Hope it is clearer now.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -128,48 +271,197 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+                        } else {
+                            false as cass_bool_t // Raw value is empty
+                        }
+                    } else {
+                        false as cass_bool_t // Iterator reached the end
+                    }
+                } else {
+                    false as cass_bool_t // Iterator position is unknown
+                }
+            }
```
wprzytula, line 330: This looks good now!

Gor027:
> Looking at the maze of `if`s, I'm convinced that a set of unit tests for deserialization is absolutely required.

I added some relevant tests for collection iterators as Cassandra tests may not cover everything.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1544,6 +1544,304 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+            let sequence_iterator =
+                SequenceIterator::deserialize(column_type.as_ref().unwrap(), Some(slice_frame))
+                    .unwrap();
+            let mut collection_iterator = CassIterator::CassCollectionIterator(
+                CassCollectionIterator::SequenceIterator(CassSequenceIterator {
+                    sequence_iterator,
+                    count: 1,
+                    state_info: CassIteratorStateInfo::NoValue,
+                }),
+            );
```
wprzytula, line 1603: This code shows the problem we discussed before: as `CassSequenceIterator` requires `'static` lifetime out the held references, the lifetime of the `&Bytes` reference must by extended to `'static` in a hacky way (by casting it to pointer and back again).
The solution is to make the iterators parametrised by a lifetime.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1544,6 +1544,304 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+            let has_next = cass_iterator_next(&mut collection_iterator);
+            assert_ne!(has_next, 0);
+            // Reached the end
+            let has_next = cass_iterator_next(&mut collection_iterator);
+            assert!(has_next.is_zero());
```
wprzytula: Wouldn't `assert_eq!(has_next, 0)` fit better here?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1544,6 +1544,304 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+            let mut bytes_mut = BytesMut::new();
+            let text = "key";
+            let true_bytes: &[u8] = &[0x01];
+            bytes_mut.put_i32(1); // Number of values
+
+            // Put serialized (string, bool) pair in bytes
+            bytes_mut.put_i32(text.len() as i32);
+            bytes_mut.put_slice(text.as_bytes());
+            bytes_mut.put_i32(1);
+            bytes_mut.put(true_bytes);
```
wprzytula: Is such by-hand crafting necessary? Couldn't we use our serialisation framework to create deserialisation tests by asserting identity of the types fed and got back?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1544,6 +1544,304 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+            let has_next = cass_iterator_next(&mut collection_iterator);
+            assert_ne!(has_next, 0);
+
+            // Hacking the iterator to not change the value but increment the position
+            if let CassIterator::CassCollectionIterator(CassCollectionIterator::SeqMapIterator(
+                CassMapIterator { state_info, .. },
+            )) = &mut collection_iterator
+            {
+                state_info.advance();
+            }
```
wprzytula: Is this `if let` condition going to ever not be satisfied? If not, please add `else` branch with explicit `unreachable!()`.

Gor027: Done.

scylla-rust-wrapper/src/query_result.rs
Gor027: It would require adding `scylla-cql` in dependencies, I do not want to add it for a couple of serialization purposes.

scylla-rust-wrapper/src/query_result.rs
Gor027: Done.

scylla-rust-wrapper/src/query_result.rs
Gor027: Let's create an issue out of it to not forget #119

scylla-rust-wrapper/src/query_result.rs
wprzytula: Rust driver already has `scylla-cql` as dependency, so I believe this shouldn't hurt us. @piodul Can you confirm that? Will Cargo be intelligent enough to compile `scylla-cql ` only once?

piodul: Yes, provided that we use the same version of `scylla-cql` as the driver uses.

wprzytula: Then I insist on using scylla-cql for aid here.

Gor027: @piodul Are we sure that the following way of serializing and deserializing should always work for all types:
```rust
// Serialize
let mut bytes = Vec::new();
let mut map = HashMap::new()
map.insert("key", "value");
map.serialize(&mut bytes);
let frame_slice = FrameSlice::new(bytes);
// Deserialize
let column_type = ColumnType::Map(Box::new(ColumnType::Text), Box::new(ColumnType::Text));
let (key, value) = MapIterator::deserialize(column_type, frame_slice).unwrap().next().unwrap();

assert_eq!(key, "key");
assert_eq!(value, "value");
```
This is a pseudocode example so it may not compile.

piodul: Yes, I think it should work. By the way, the impls for the `Value` trait are in `scylla-cql/src/frame/value.rs`, looks like they do the same thing.

Gor027: Ok, what was missing is that before the serialized value the buffer contains also the number of bytes of it, but the deserialization API expects a frame slice without the number of bytes in the beginning.

Lorak-mmk:
I think we can safely close this.

Lorak-mmk:
Oops, wrong PR

Lorak-mmk:
After all I think we can close this - applying new deserialization to cpp-rust-driver will require at least some changes to this PR, so it will have to be done by someone else.

PR #128, opened 2024-06-03
Support homebrew

syuu1228:
Add homebrew script to support installing cpp-rust-driver to macOS.


syuu1228:
Implemented homebrew support since cpp-driver has it too.

wprzytula (approved):

Lorak-mmk (approved):

PR #135, opened 2024-07-02
types: Implement API functions for `duration` type

muzarski:
This PR implements following set of API functions related to `duration` type:
- cass_statement_bind_duration_*
- cass_user_type_set_duration_*
- cass_tuple_set_duration
- cass_value_get_duration
- cass_value_is_duration
- cass_collection_append_duration

## Pre-review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,13 +197,10 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
         </tr>
         <tr>
             <td>cass_user_type_set_custom[by_name]</td>
-            <td rowspan="3">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the value being set for a field of the UDT is compatible with the field's actual type.</td>
+            <td rowspan="2">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the value being set for a field of the UDT is compatible with the field's actual type.</td>
         </tr>
         <tr>
             <td>cass_user_type_set_decimal[by_name]</td>
-        </tr>
-        <tr>
-            <td>cass_user_type_set_duration[by_name]</td>
         </tr>
         <tr>
             <td colspan=2 align="center" style="font-weight:bold">Value</td>
--- a/README.md
+++ b/README.md
@@ -160,13 +160,10 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
         </tr>
         <tr>
             <td>cass_statement_bind_custom[by_name]</td>
-            <td rowspan="3">Binding is not implemented for custom types in the Rust driver. <br> Binding Decimal and Duration types requires encoding raw bytes into BigDecimal and CqlDuration types in the Rust driver. <br> <b>Note</b>: The driver does not validate the types of the values passed to queries.</td>
+            <td rowspan="2">Binding is not implemented for custom types in the Rust driver. <br> Binding Decimal type requires encoding raw bytes into BigDecimal type in the Rust driver. <br> <b>Note</b>: The driver does not validate the types of the values passed to queries.</td>
         </tr>
         <tr>
             <td>cass_statement_bind_decimal[by_name]</td>
-        </tr>
-        <tr>
-            <td>cass_statement_bind_duration[by_name]</td>
         </tr>
         <tr>
             <td colspan=2 align="center" style="font-weight:bold">Future</td>
--- a/README.md
+++ b/README.md
@@ -190,13 +190,10 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
         </tr>
         <tr>
             <td>cass_collection_append_custom[_n]</td>
-            <td rowspan="3">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the appended value is compatible with the type of the collection items.</td>
+            <td rowspan="2">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the appended value is compatible with the type of the collection items.</td>
         </tr>
         <tr>
             <td>cass_collection_append_decimal</td>
-        </tr>
-        <tr>
-            <td>cass_collection_append_duration</td>
         </tr>
         <tr>
             <td colspan=2 align="center" style="font-weight:bold">User Defined Type</td>
--- a/README.md
+++ b/README.md
@@ -215,15 +215,8 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
             <td colspan=2 align="center" style="font-weight:bold">Value</td>
         </tr>
         <tr>
-            <td>cass_value_is_duration</td>
-            <td>Unimplemented</td>
-        </tr>
-        <tr>
             <td>cass_value_get_decimal</td>
-            <td rowspan="2">Getting raw bytes of Decimal and Duration values requires lazy deserialization feature in the Rust driver.</td>
-        </tr>
-        <tr>
-            <td>cass_value_get_duration</td>
+            <td>Getting raw bytes of Decimal values requires lazy deserialization feature in the Rust driver.</td>
         </tr>
         <tr>
             <td>cass_value_get_bytes</td>
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -137,9 +137,8 @@ macro_rules! make_appender {
 }
 
 // TODO: Types for which binding is not implemented yet:
-// custom - Not implemented in Rust driver?
+// custom - Not implemented in Rust driver
 // decimal
-// duration - DURATION not implemented in Rust Driver
 
 macro_rules! invoke_binder_maker_macro_with_type {
     (null, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
@@ -275,6 +274,21 @@ macro_rules! invoke_binder_maker_macro_with_type {
                 }
             },
             [v @ crate::inet::CassInet]
+        );
+    };
+    (duration, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
+        $macro_name!(
+            $this,
+            $consume_v,
+            $fn,
+            |m, d, n| {
+                Ok(Some(Duration(scylla::frame::value::CqlDuration {
+                    months: m,
+                    days: d,
+                    nanoseconds: n
+                })))
+            },
+            [m @ cass_int32_t, d @ cass_int32_t, n @ cass_int64_t]
         );
     };
     (collection, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1165,12 +1165,12 @@ pub unsafe extern "C" fn cass_value_is_null(value: *const CassValue) -> cass_boo
 pub unsafe extern "C" fn cass_value_is_collection(value: *const CassValue) -> cass_bool_t {
     let val = ptr_to_ref(value);
 
-    match val.value {
-        Some(Value::CollectionValue(Collection::List(_))) => true as cass_bool_t,
-        Some(Value::CollectionValue(Collection::Set(_))) => true as cass_bool_t,
-        Some(Value::CollectionValue(Collection::Map(_))) => true as cass_bool_t,
-        _ => false as cass_bool_t,
-    }
+    matches!(
+        val.value_type.get_value_type(),
+        CassValueType::CASS_VALUE_TYPE_LIST
+            | CassValueType::CASS_VALUE_TYPE_SET
+            | CassValueType::CASS_VALUE_TYPE_MAP
+    ) as cass_bool_t
 }
 
 #[no_mangle]
@@ -1544,9 +1544,6 @@ extern "C" {
 }
 extern "C" {
     pub fn cass_value_type(value: *const CassValue) -> CassValueType;
-}
-extern "C" {
-    pub fn cass_value_is_collection(value: *const CassValue) -> cass_bool_t;
 }
 extern "C" {
     pub fn cass_value_item_count(collection: *const CassValue) -> size_t;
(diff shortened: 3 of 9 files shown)


muzarski:
@Lorak-mmk @wprzytula Do you know how to enable the tests for the duration type? I see there is a `tests/src/integration/values/duration.hpp` file but I'm not sure how to enable it in CI.

muzarski:
> @Lorak-mmk @wprzytula Do you know how to enable the tests for the duration type? I see there is a `tests/src/integration/values/duration.hpp` file but I'm not sure how to enable it in CI.

I see that the duration tests were run in CI: `12 tests from CassandraTypes/CassandraTypesTests/7, where TypeParam = test::driver::NullableValue<test::driver::values::Duration>`. I'm curious why `Decimal` tests are not being run (decimal is unimplemented type).

Lorak-mmk (approved):

Lorak-mmk:
> @Lorak-mmk @wprzytula Do you know how to enable the tests for the duration type? I see there is a `tests/src/integration/values/duration.hpp` file but I'm not sure how to enable it in CI.

To be honest I'm not sure how those tests work. I can see that both duration and decimal have some special handling.
Those tests are defined in `tests/src/integration/tests/test_cassandra_types.cpp`
I can see that there is some specialization for Duration: https://github.com/scylladb/cpp-rust-driver/blob/805ef4c3a3f35e35d9e0736a86e47b2c600c3c4e/tests/src/integration/tests/test_cassandra_types.cpp#L98
And later some registration for types excluding duration: https://github.com/scylladb/cpp-rust-driver/blob/805ef4c3a3f35e35d9e0736a86e47b2c600c3c4e/tests/src/integration/tests/test_cassandra_types.cpp#L884-L891
There is also some instantiation that includes decimal: https://github.com/scylladb/cpp-rust-driver/blob/805ef4c3a3f35e35d9e0736a86e47b2c600c3c4e/tests/src/integration/tests/test_cassandra_types.cpp#L925-L930

Lorak-mmk:
> > @Lorak-mmk @wprzytula Do you know how to enable the tests for the duration type? I see there is a `tests/src/integration/values/duration.hpp` file but I'm not sure how to enable it in CI.
> 
> I see that the duration tests were run in CI: `12 tests from CassandraTypes/CassandraTypesTests/7, where TypeParam = test::driver::NullableValue<test::driver::values::Duration>`. I'm curious why `Decimal` tests are not being run (decimal is unimplemented type).

Where do you see that? I don't see any occurence of decimal or duration in CI logs.

muzarski:
> > > @Lorak-mmk @wprzytula Do you know how to enable the tests for the duration type? I see there is a `tests/src/integration/values/duration.hpp` file but I'm not sure how to enable it in CI.
> > 
> > 
> > I see that the duration tests were run in CI: `12 tests from CassandraTypes/CassandraTypesTests/7, where TypeParam = test::driver::NullableValue<test::driver::values::Duration>`. I'm curious why `Decimal` tests are not being run (decimal is unimplemented type).
> 
> Where do you see that? I don't see any occurence of decimal or duration in CI logs.

build job, step `Run integration tests on Scylla 5.0.0`, line 577

Lorak-mmk:
I see. Ctrl + f seems to be quite unreliable in those logs...

muzarski:
v2: rebased on master

README.md
```diff
@@ -160,14 +160,11 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
         </tr>
         <tr>
             <td>cass_statement_bind_custom[by_name]</td>
-            <td rowspan="3">Binding is not implemented for custom types in the Rust driver. <br> Binding Decimal and Duration types requires encoding raw bytes into BigDecimal and CqlDuration types in the Rust driver. <br> <b>Note</b>: The driver does not validate the types of the values passed to queries.</td>
+            <td rowspan="2">Binding is not implemented for custom types in the Rust driver. <br> Binding Decimal type requires encoding raw bytes into BigDecimal type in the Rust driver. <br> <b>Note</b>: The driver does not validate the types of the values passed to queries.</td>
```
wprzytula, line 163: > The driver does not validate the types of the values passed to queries.

Is this still true after the driver version was bumped to include the serialization refactor?

README.md
```diff
@@ -190,40 +187,27 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
         </tr>
         <tr>
             <td>cass_collection_append_custom[_n]</td>
-            <td rowspan="3">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the appended value is compatible with the type of the collection items.</td>
+            <td rowspan="2">Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the appended value is compatible with the type of the collection items.</td>
```
wprzytula, line 190: Why can we decrement the span?

scylla-rust-wrapper/src/binding.rs
```diff
@@ -139,7 +139,6 @@ macro_rules! make_appender {
 // TODO: Types for which binding is not implemented yet:
 // custom - Not implemented in Rust driver?
 // decimal
```
wprzytula: Let's remove the question mark. Custom types are not supported in Rust driver, and are not planned to be supported in the future.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1102,6 +1102,28 @@ pub unsafe extern "C" fn cass_value_get_string(
...
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_get_duration(
+    value: *const CassValue,
+    months: *mut cass_int32_t,
+    days: *mut cass_int32_t,
+    nanos: *mut cass_int64_t,
+) -> CassError {
+    let val: &CassValue = ptr_to_ref(value);
+
```
wprzytula: `cass_value_get_bytes()` additionally checks if the pointer is NULL, and returns an error in such case. Please double check that cpp-driver does it not for `cass_value_get_duration`, too, and gain insight why it does so for `bytes`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1146,6 +1168,16 @@ pub unsafe extern "C" fn cass_value_is_collection(value: *const CassValue) -> ca
...
 
+#[no_mangle]
+pub unsafe extern "C" fn cass_value_is_duration(value: *const CassValue) -> cass_bool_t {
+    let val = ptr_to_ref(value);
+
+    match val.value {
+        Some(Value::RegularValue(CqlValue::Duration(_))) => true as cass_bool_t,
+        _ => false as cass_bool_t,
+    }
+}
```
wprzytula: Shouldn't we rather use a consise `matches!` macro? I recall that clippy tends to complain about writing the fully-fledged `match` statement in such cases. Probably it's the cast to `cass_bool_t` that keeps it silent here.

scylla-rust-wrapper/src/value.rs
```diff
@@ -95,6 +101,31 @@ impl CassCqlValue {
     }
 }
 
+// These are copied from rust-driver. Maybe we should make them public there.
+fn zig_zag_encode(v: i64) -> u64 {
+    ((v >> 63) ^ (v << 1)) as u64
+}
```
wprzytula: Definitely, let's not copy them!
Let's make `vint_encode` and `vint_decode` public in `scylla_cql`. `cpp-rust-driver` can depend on `scylla-cql` directly, too, and then we don't even have to expose them in `scylla` crate.

scylla-rust-wrapper/src/value.rs
```diff
@@ -191,6 +222,13 @@ fn_serialize_via_writer!(serialize_inet, IpAddr, |me, writer| {
         IpAddr::V6(ip) => writer.set_value(&ip.octets()).unwrap(),
     }
 });
+fn_serialize_via_writer!(serialize_duration, CqlDuration, |me, writer| {
+    let mut buf = Vec::with_capacity(27); // worst case size is 27
```
wprzytula: Wait, why? This 27 seems like magic to me.

README.md
muzarski: It won't be true once we merge: https://github.com/scylladb/cpp-rust-driver/pull/143.

The PR that bumped driver's version implemented a naive serialization (with no typechecks). In cpp-driver, the typechecks are performed during binding, so I decided to implement it in separate PR.

README.md
muzarski: Because we removed `cass_collection_append_duration` row:
```
        <tr>
            <td>cass_collection_append_duration</td>
        </tr>
```

This leaves us with just `cass_collection_apend_custom` and `cass_collection_append_decimal` rows.

scylla-rust-wrapper/src/query_result.rs
muzarski: The cpp-driver actually does that for all of `cass_value_get_[...]` functions. I'll adjust it in cpp-rust-driver in this PR. To be quite honest, I'm not sure when `CassValue` pointer can be null, but it's better to verify that it's non-null anyway.

wprzytula: That's yet another spot in cpp-rust-driver where we are hazardously close to triggering UB. For the next releases, we undoubtedly have to focus on stability and memory safety. Let's start with refactoring the binding framework, as discussed, and then we should probably proceed with thorough and meticulous code inspection to detect potential trouble before it arises in production.

muzarski: What's interesting is that other `cass_value_*` functions in cpp-driver do not have that check - e.g. `cass_value_is_null` does not check whether the pointer itself is null.

scylla-rust-wrapper/src/query_result.rs
muzarski: Good thing that you mentioned that! Thanks to that, I realized that the logic itself is not correct. Even if the value is `None`, it can be a duration value (null). We need to check the type, not the value.

The bug is a result of me basing on implementation of `cass_value_is_collection`, which has this bug as well...

scylla-rust-wrapper/src/value.rs
muzarski: This is rust driver's code. I'll try to investigate it.

scylla-rust-wrapper/src/value.rs
muzarski: Opened a PR: https://github.com/scylladb/scylla-rust-driver/pull/1047

scylla-rust-wrapper/src/value.rs
muzarski: I think I found an explanation. A fragment of `unsigned_vint_encode`:
```rust
    let mut number_of_bytes = (639 - 9 * v.leading_zeros()) >> 6;
```

The maximum value of `number_of_bytes` is 9 (`639 >> 6`). The value is `mut`, but it is only decreased in some cases. We have 3 values (months, days, nanos), 3 x 9 = 27.

wprzytula: Let's put the explanation in the code. It should land in Rust driver's code, too.

piodul: Can't we just invoke `<CqlDuration as SerializeValue>::serialize()` here? Is this necessary to reimplement this in such low level way?

muzarski: We unfortunately can't, since rust-driver does the typechecks in builtin `SerializeValue` implementations. cpp-driver handles typechecks in a very specific way. We should not do the typechecks for elements of an untyped collection (yes, there is such thing in cpp-driver) - users are allowed to bind a value of any type to such collection (thus, the typecheck should not be performed).

piodul: The implementation in scylla-rust-driver indeed performs type checking, but can't we just pass the expected type when serializing? i.e. `my_cql_duration_value.serialize(&CqlType::Duration, writer)`?

muzarski: Hmmm, I like this idea. We could do it for other types excluding udts/collections/tuples as well.

@Lorak-mmk @wprzytula WDYT?

muzarski:
v3:
- rebased on master
- addressed minor issues
- fixed `cass_value_is_duration` and `cass_value_is_collection`
- making use of rust-driver's builtin serialization for simple types (first commit).

scylla-rust-wrapper/src/value.rs
```diff
@@ -73,18 +77,53 @@ impl CassCqlValue {
...
+            // will never fail. Thanks to that, we do not have to reimplement low-level serialization
+            // for each type.
+            CassCqlValue::TinyInt(v) => {
+                <i8 as SerializeCql>::serialize(v, &ColumnType::TinyInt, writer)
+            }
+            CassCqlValue::SmallInt(v) => {
+                <i16 as SerializeCql>::serialize(v, &ColumnType::SmallInt, writer)
+            }
+            CassCqlValue::Int(v) => <i32 as SerializeCql>::serialize(v, &ColumnType::Int, writer),
+            CassCqlValue::BigInt(v) => {
```
wprzytula: I'd prefer to have these changes in a separate PR. Refactors should be separate from new features (duration support).

wprzytula (changes requested):

scylla-rust-wrapper/src/value.rs
muzarski: Ok, I'll move it. This PR will base on the PR introducing these changes.

muzarski: https://github.com/scylladb/cpp-rust-driver/pull/152

muzarski:
v3.1: rebased on master

wprzytula (approved):

Lorak-mmk (approved):

scylla-rust-wrapper/src/value.rs
Lorak-mmk: Those are no longer required, so I'm marking as resolved

Lorak-mmk:
One more question: before you mentioned some tests that were incorrectly skipped. Is that fixed now?

muzarski:
> One more question: before you mentioned some tests that were incorrectly skipped. Is that fixed now?

Yes, there is an issue for that: https://github.com/scylladb/cpp-rust-driver/issues/144. Do you want me to fix it before merging this PR?

Lorak-mmk:
It can be a separate PR - but I don't want to postpone it too long.

PR #136, opened 2024-07-03
cass_result: implement missing functions

### LINKED ISSUE #35: CassResult missing functions
Both _count and cass_result_column_type and  cass_result_column_name should be easy (if Rust driver has support, but it probably has).
Not sure about cass_result_column_data_type and cass_result_paging_state_token.

- [x] `cass_result_row_count`
- [x] `cass_result_column_count`
- [x] `cass_result_column_name`
- [ ] `cass_result_column_type`
- [ ] `cass_result_column_data_type`
- [x] `cass_result_first_row`
- [x] `cass_result_paging_state_token`

API docs:
https://docs.datastax.com/en/developer/cpp-driver/2.16/api/struct.CassResult/

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/35

Implemented two missing CassResult functions:
- cass_result_column_type
- cass_result_column_data_type

Also fixed `cass_result_first_row` implementation.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/prepared.rs
+++ b/scylla-rust-wrapper/src/prepared.rs
@@ -12,6 +12,11 @@ use scylla::prepared_statement::PreparedStatement;
 pub struct CassPrepared {
     // Data types of columns from PreparedMetadata.
     pub variable_col_data_types: Vec<Arc<CassDataType>>,
+    // Data types of columns from ResultMetadata.
+    //
+    // Arc<CassDataType> -> to share each data type with other structs such as `CassValue`
+    // Arc<Vec<...>> -> to share the whole vector with `CassResultData`.
+    pub result_col_data_types: Arc<Vec<Arc<CassDataType>>>,
     pub statement: PreparedStatement,
 }
 
@@ -23,8 +28,17 @@ impl CassPrepared {
             .map(|col_spec| Arc::new(get_column_type(&col_spec.typ)))
             .collect();
 
+        let result_col_data_types: Arc<Vec<Arc<CassDataType>>> = Arc::new(
+            statement
+                .get_result_set_col_specs()
+                .iter()
+                .map(|col_spec| Arc::new(get_column_type(&col_spec.typ)))
+                .collect(),
+        );
+
         Self {
             variable_col_data_types,
+            result_col_data_types,
             statement,
         }
     }
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1385,6 +1385,197 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
     CassError::CASS_OK
 }
 
+#[cfg(test)]
+mod tests {
+    use std::{ffi::c_char, ptr::addr_of_mut, sync::Arc};
+
+    use scylla::{
+        frame::response::result::{ColumnSpec, ColumnType, CqlValue, Row, TableSpec},
+        transport::PagingStateResponse,
+    };
+
+    use crate::{
+        cass_error::CassError,
+        cass_types::{CassDataType, CassValueType},
+        query_result::{
+            cass_result_column_data_type, cass_result_column_name, cass_result_first_row,
+            ptr_to_cstr_n, ptr_to_ref, size_t,
+        },
+        session::create_cass_rows_from_rows,
+    };
+
+    use super::{cass_result_column_count, cass_result_column_type, CassResult, CassResultData};
+
+    fn col_spec(name: &str, typ: ColumnType) -> ColumnSpec {
+        ColumnSpec {
+            table_spec: TableSpec::borrowed("ks", "tbl"),
+            name: name.to_owned(),
+            typ,
+        }
+    }
+
+    const FIRST_COLUMN_NAME: &str = "bigint_col";
+    const SECOND_COLUMN_NAME: &str = "varint_col";
+    const THIRD_COLUMN_NAME: &str = "list_double_col";
+    fn create_cass_rows_result() -> CassResult {
+        let metadata = Arc::new(CassResultData::from_result_payload(
+            PagingStateResponse::NoMorePages,
+            vec![
+                col_spec(FIRST_COLUMN_NAME, ColumnType::BigInt),
+                col_spec(SECOND_COLUMN_NAME, ColumnType::Varint),
+                col_spec(
+                    THIRD_COLUMN_NAME,
+                    ColumnType::List(Box::new(ColumnType::Double)),
+                ),
+            ],
+            None,
+            None,
+        ));
+
+        let rows = create_cass_rows_from_rows(
+            Some(vec![Row {
+                columns: vec![
+                    Some(CqlValue::BigInt(42)),
+                    None,
+                    Some(CqlValue::List(vec![
+                        CqlValue::Float(0.5),
+                        CqlValue::Float(42.42),
+                        CqlValue::Float(9999.9999),
+                    ])),
+                ],
+            }]),
+            &metadata,
+        );
+
+        CassResult { rows, metadata }
+    }
+
+    unsafe fn cass_result_column_name_rust_str(
+        result_ptr: *const CassResult,
+        column_index: u64,
+    ) -> Option<&'static str> {
+        let mut name_ptr: *const c_char = std::ptr::null();
+        let mut name_length: size_t = 0;
+        let cass_err = cass_result_column_name(
+            result_ptr,
+            column_index,
+            addr_of_mut!(name_ptr),
+            addr_of_mut!(name_length),
+        );
+        assert_eq!(CassError::CASS_OK, cass_err);
+        ptr_to_cstr_n(name_ptr, name_length)
+    }
+
+    #[test]
+    fn rows_cass_result_api_test() {
+        let result = create_cass_rows_result();
+
+        unsafe {
+            let result_ptr = std::ptr::addr_of!(result);
+
+            // cass_result_column_count test
+            {
+                let column_count = cass_result_column_count(result_ptr);
+                assert_eq!(3, column_count);
+            }
+
+            // cass_result_column_name test
+            {
+                let first_column_name = cass_result_column_name_rust_str(result_ptr, 0).unwrap();
+                assert_eq!(FIRST_COLUMN_NAME, first_column_name);
+                let second_column_name = cass_result_column_name_rust_str(result_ptr, 1).unwrap();
+                assert_eq!(SECOND_COLUMN_NAME, second_column_name);
+                let third_column_name = cass_result_column_name_rust_str(result_ptr, 2).unwrap();
+                assert_eq!(THIRD_COLUMN_NAME, third_column_name);
+            }
+
+            // cass_result_column_type test
+            {
+                let first_col_type = cass_result_column_type(result_ptr, 0);
+                assert_eq!(CassValueType::CASS_VALUE_TYPE_BIGINT, first_col_type);
+                let second_col_type = cass_result_column_type(result_ptr, 1);
+                assert_eq!(CassValueType::CASS_VALUE_TYPE_VARINT, second_col_type);
+                let third_col_type = cass_result_column_type(result_ptr, 2);
+                assert_eq!(CassValueType::CASS_VALUE_TYPE_LIST, third_col_type);
+                let out_of_bound_col_type = cass_result_column_type(result_ptr, 555);
+                assert_eq!(
+                    CassValueType::CASS_VALUE_TYPE_UNKNOWN,
+                    out_of_bound_col_type
+                );
+            }
+
+            // cass_result_column_data_type test
+            {
+                let first_col_data_type = ptr_to_ref(cass_result_column_data_type(result_ptr, 0));
+                assert_eq!(
+                    &CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BIGINT),
+                    first_col_data_type
+                );
+                let second_col_data_type = ptr_to_ref(cass_result_column_data_type(result_ptr, 1));
+                assert_eq!(
+                    &CassDataType::Value(CassValueType::CASS_VALUE_TYPE_VARINT),
+                    second_col_data_type
+                );
+                let third_col_data_type = ptr_to_ref(cass_result_column_data_type(result_ptr, 2));
+                assert_eq!(
+                    &CassDataType::List {
+                        typ: Some(Arc::new(CassDataType::Value(
+                            CassValueType::CASS_VALUE_TYPE_DOUBLE
+                        ))),
+                        frozen: false
+                    },
+                    third_col_data_type
+                );
+                let out_of_bound_col_data_type = cass_result_column_data_type(result_ptr, 555);
+                assert!(out_of_bound_col_data_type.is_null());
+            }
+        }
+    }
+
+    fn create_non_rows_cass_result() -> CassResult {
+        let metadata = Arc::new(CassResultData::from_result_payload(
+            PagingStateResponse::NoMorePages,
+            vec![],
+            None,
+            None,
+        ));
+        CassResult {
+            rows: None,
+            metadata,
+        }
+    }
+
+    #[test]
+    fn non_rows_cass_result_api_test() {
+        let result = create_non_rows_cass_result();
+
+        // Check that API functions do not panic when rows are empty - e.g. for INSERT queries.
+        unsafe {
+            let result_ptr = std::ptr::addr_of!(result);
+
+            assert_eq!(0, cass_result_column_count(result_ptr));
+            assert_eq!(
+                CassValueType::CASS_VALUE_TYPE_UNKNOWN,
+                cass_result_column_type(result_ptr, 0)
+            );
+            assert!(cass_result_column_data_type(result_ptr, 0).is_null());
+            assert!(cass_result_first_row(result_ptr).is_null());
+
+            {
+                let mut name_ptr: *const c_char = std::ptr::null();
+                let mut name_length: size_t = 0;
+                let cass_err = cass_result_column_name(
+                    result_ptr,
+                    0,
+                    addr_of_mut!(name_ptr),
+                    addr_of_mut!(name_length),
+                );
+                assert_eq!(CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS, cass_err);
+            }
+        }
+    }
+}
+
 // CassResult functions:
 /*
 extern "C" {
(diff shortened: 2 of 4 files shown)


mykaul:
Do we have tests to validate this?

muzarski:
> Do we have tests to validate this?

At first, I wasn't sure if there are integration tests checking this - unfortunately, the answer is no (previous implementation of `cass_result_first_row` would panic in some cases - this bug was never detected). Tomorrow I'll try to implement some cpp integration tests for the `cass_result_*` API.

muzarski:
v2: Fixed `cass_result_first_row`

wprzytula:
> > Do we have tests to validate this?
> 
> At first, I wasn't sure if there are integration tests checking this - unfortunately, the answer is no (previous implementation of `cass_result_first_row` would panic in some cases - this bug was never detected). Tomorrow I'll try to implement some cpp integration tests for the `cass_result_*` API.

WDYT about adding Rust unit tests first? They should be the first layer to detect errors, they have smaller overhead and, probably the most important for convenience of us maintainers, they are written in Rust, not in Datastax C++ test framework.

muzarski:
v3: implemented unit tests for basic `cass_result_*` API.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1298,6 +1298,188 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+        result_ptr: *const CassResult,
+        column_index: u64,
+    ) -> Option<&'static str> {
+        let mut name_ptr: *const c_char = std::ptr::null();
+        let mut name_length: size_t = 0;
+        let cass_err = cass_result_column_name(
+            result_ptr,
+            column_index,
+            std::ptr::addr_of_mut!(name_ptr),
+            std::ptr::addr_of_mut!(name_length),
```
wprzytula: As this is going to be prevalent in cpp-rust-driver code, let's `use std::ptr::addr_of[_mut]`.

wprzytula (approved):

wprzytula:
In this PR, you should instantly enable related integration tests and/or examples.

scylla-rust-wrapper/src/session.rs
```diff
@@ -289,11 +285,11 @@ pub unsafe extern "C" fn cass_session_execute(
...
-                let metadata = Arc::new(CassResultData {
-                    paging_state: result.paging_state,
-                    col_specs: result.col_specs,
-                    tracing_id: result.tracing_id,
-                });
+                let metadata = Arc::new(CassResultData::from_result_payload(
+                    result.paging_state,
+                    result.col_specs,
+                    result.tracing_id,
+                ));
```
Lorak-mmk: Allocating this on each execute seems very wasteful.
Can we store it in prepared statement and share an Arc?
What is cpp-driver doing? Does it allocate on each execute or does it share with prepared?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -922,6 +922,38 @@ pub unsafe extern "C" fn cass_result_column_name(
...
+) -> CassValueType {
+    let result_from_raw: &CassResult = ptr_to_ref(result);
+    let index_usize: usize = index.try_into().unwrap();
+
+    result_from_raw
+        .metadata
+        .col_data_types
+        .get(index_usize)
+        .map(|arc_data_type| cass_data_type_type(Arc::as_ptr(arc_data_type)))
+        .unwrap_or(CassValueType::CASS_VALUE_TYPE_UNKNOWN)
```
Lorak-mmk: This could call `cass_result_column_data_type` and map non-null using `cass_data_type_type`.
Not sure if it would be better in any way (but it would avoid repetition of argument conversion and accessing col_data_types.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/session.rs
muzarski: Current definition of `Statement` is:
```rust
#[derive(Clone)]
pub enum Statement {
    Simple(SimpleQuery),
    // Arc is needed, because PreparedStatement is passed by reference to session.execute
    Prepared(Arc<PreparedStatement>),
}
```

So it can be either a simple query or prepared statement. In case of simple query, we don't have information about the result column specs.

When it comes to prepared statement, we could use columns specs (but why, when they are provided in `QueryResult`). What about `paging_state` and `tracing_id`? They are not part of prepared statement, but rather a part of result.

The only optimization I can think of is to extend the `Statement::Prepared` variant by an additional field of type `Arc<Vec<Arc<CassDataType>>>` (`CassDataType` wrapped with Arc since it needs to be shared with `CassValue`, which resides in `CassRow`). It could be then computed during preparation, and passed to `CassResult::col_data_types` upon exeuction (simply clone an `Arc`).

Lorak-mmk: > The only optimization I can think of is to extend the `Statement::Prepared` variant by an additional field of type `Arc<Vec<Arc<CassDataType>>>` (`CassDataType` wrapped with Arc since it needs to be shared with `CassValue`, which resides in `CassRow`). It could be then computed during preparation, and passed to `CassResult::col_data_types` upon exeuction (simply clone an `Arc`).

This is what I had in mind by storing it in prepared statement and sharing an Arc.
Nit: I think it makes more sense to make `CassPrepared` a struct instead of an alias and add the data type there.

I looked briefly at cpp-driver code and I'm pretty sure that this is what it does too.

muzarski:
This PR is waiting for: https://github.com/scylladb/cpp-rust-driver/pull/143

dkropachev:
> This PR is waiting for: #143

#143 is in, can you continue working on it ?

muzarski:
> > This PR is waiting for: #143
> 
> 
> 
> #143 is in, can you continue working on it ? 

Yes

muzarski:
Rebased on master

muzarski:
v4:
Now, the result metadata (`Vec<Arc<CassDataType>>`) is stored in `CassPrepared`. It allows us to omit the allocations each time we execute the prepared statement. We don't have to construct each `CassDataType` from scratch (it can be a really costly operation) based on the column specs.

Now, `CassResultData::from_result_payload` accepts an additional argument, which is an optional vector of result column data types. 
- For simple queries, we provide `None` - then `CassDataTypes` are created based on column specs received from the server. 
- For prepared statements, we provide a `Some(<metadata cached in CassPrepared>)` to avoid additional allocations

scylla-rust-wrapper/src/session.rs
muzarski: Done

scylla-rust-wrapper/src/prepared.rs
```diff
@@ -12,6 +12,8 @@ use scylla::prepared_statement::PreparedStatement;
 pub struct CassPrepared {
     // Data types of columns from PreparedMetadata.
     pub variable_col_data_types: Vec<Arc<CassDataType>>,
+    // Data types of columns from ResultMetadata.
+    pub result_col_data_types: Vec<Arc<CassDataType>>,
     pub statement: PreparedStatement,
 }
```
Lorak-mmk: Do I understand correctly: first field is for types of bind variables, second for types of columns returned from query?

scylla-rust-wrapper/src/session.rs
```diff
@@ -285,40 +286,77 @@ pub unsafe extern "C" fn cass_session_execute(
...
+                                PagingStateResponse::NoMorePages,
+                                maybe_result_col_data_types,
+                            )
+                        })
                 }
             }
             Statement::Prepared(prepared) => {
+                // Clone vector of the Arc<CassDataType>, so we don't do additional allocations when constructing
+                // CassDataTypes in `CassResultData::from_result_payload`.
+                let maybe_result_col_data_types = Some(prepared.result_col_data_types.clone());
```
Lorak-mmk, line 330: Why do you store `Vec<Arc<...>>` and clone the vector (which needs an allocation) instead of putting Vec in another Arc?

scylla-rust-wrapper/src/prepared.rs
muzarski: Yes

scylla-rust-wrapper/src/session.rs
muzarski: True, I'll do it

Lorak-mmk: And if you put Vec in Arc, is there a need to have individual data types in Arcs too?

muzarski: Yes, I even decided to put a comment in the code to justify both Arcs.

- `Vec<Arc<...>>` -> to share vector with `CassResultData`
- `Arc<CassDataType>` - to share each data type with other structs. such as `CassValue`.

muzarski: BTW, wrapping this vector with `Arc` implies additional needless allocation for simple queries (for an `Arc`), but I think performance for prepared statements is more important

Lorak-mmk: There will already be 1 allocation for Vec and at least 1 for each element. Another one for this new Arc doesn't seem like much.

muzarski:
v4.1: wrapped the result data type metadata vector with an `Arc`.

The reason for this is that `CassPrepared` and `CassResultData` should share this vector, and this way we can omit an additional allocation each execution.

Note: Each individual `CassDataType` in the vector still needs to be wrapped with an `Arc`. This is because each `CassValue` (contained in `CassResult` in each `CassRow`) needs to hold information about the type.

Lorak-mmk (approved):

scylla-rust-wrapper/src/prepared.rs
```diff
@@ -12,6 +12,11 @@ use scylla::prepared_statement::PreparedStatement;
 pub struct CassPrepared {
     // Data types of columns from PreparedMetadata.
     pub variable_col_data_types: Vec<Arc<CassDataType>>,
+    // Data types of columns from ResultMetadata.
+    //
+    // Arc<CassDataType> -> to share each data type with other structs such as `CassValue`
+    // Arc<Vec<...>> -> to share the whole vector with `CassResultData``.
```
wprzytula: typo: repeated ``

scylla-rust-wrapper/src/session.rs
```diff
@@ -285,40 +286,77 @@ pub unsafe extern "C" fn cass_session_execute(
...
-        let query_res: Result<(QueryResult, PagingStateResponse), QueryError> = match statement {
+        // Creating a type alias here to fix clippy lints.
+        // I want this type to be explicit, so future developers can understand
+        // what's going on here (and why we include some weird Option of data types).
+        type QueryRes = Result<
+            (
+                QueryResult,
+                PagingStateResponse,
+                // We unfortunately have to retrieve the metadata here.
+                // Since `query.qery` is consumed, we cannot match the statement
```
wprzytula: typo: `qery`

scylla-rust-wrapper/src/session.rs
```diff
@@ -285,40 +286,77 @@ pub unsafe extern "C" fn cass_session_execute(
...
+                // after execution, to retrieve the cached metadata in case
+                // of prepared statements.
+                Option<Arc<Vec<Arc<CassDataType>>>>,
+            ),
+            QueryError,
+        >;
+        let query_res: QueryRes = match statement {
             Statement::Simple(query) => {
+                // We don't store result metadta for Queries - return None.
```
wprzytula: typo: `metadta`

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -22,9 +24,36 @@ pub struct CassResult {
...
 
+impl CassResultData {
+    pub fn from_result_payload(
+        paging_state_response: PagingStateResponse,
+        col_specs: Vec<ColumnSpec>,
+        maybe_col_data_types: Option<Arc<Vec<Arc<CassDataType>>>>,
+        tracing_id: Option<Uuid>,
+    ) -> CassResultData {
+        let col_data_types = maybe_col_data_types.unwrap_or_else(|| {
+            // This allocation is unfortunately necessary, because of the type of CassResultData::col_data_types.
```
wprzytula: Could you add a comment that `maybe_col_data_types` is going to be `Some` for prepared statement execution and `None` for unprepared?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -934,6 +934,34 @@ pub unsafe extern "C" fn cass_result_column_name(
...
+    cass_data_type_type(data_type_ptr)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_result_column_data_type(
+    result: *const CassResult,
+    index: size_t,
+) -> *const CassDataType {
+    let result_from_raw: &CassResult = ptr_to_ref(result);
+    let index_usize: usize = index.try_into().unwrap();
```
wprzytula: Justify `unwrap`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1340,11 +1340,12 @@ pub unsafe extern "C" fn cass_result_column_count(result_raw: *const CassResult)
...
-        return result.rows.as_ref().unwrap().first().unwrap();
-    }
-
-    std::ptr::null()
+    result
+        .rows
+        .as_ref()
+        .and_then(|rows| rows.first())
+        .map(|row| row as *const CassRow)
+        .unwrap_or(std::ptr::null())
```
wprzytula, line 1353: Thank Ferris, the monster is gone...

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1380,6 +1380,197 @@ pub unsafe extern "C" fn cass_result_paging_state_token(
...
+
+    use super::{cass_result_column_count, cass_result_column_type, CassResult, CassResultData};
+
+    fn col_spec(name: &str, typ: ColumnType) -> ColumnSpec {
+        ColumnSpec {
+            table_spec: TableSpec::borrowed("ks", "tbl"),
+            name: name.to_string(),
+            typ,
+        }
+    }
```
wprzytula: nit: I would prefer to use `to_owned()` than `to_string()` for `&str` -> `String` conversion, because `to_string()` is very generic (over anything implementing `Display`), whereas `to_owned()` hints that the struct is some borrowed form of string.

scylla-rust-wrapper/src/session.rs
```diff
@@ -377,7 +377,7 @@ pub unsafe extern "C" fn cass_session_execute(
     }
 }
 
-fn create_cass_rows_from_rows(
+pub(crate) fn create_cass_rows_from_rows(
     rows: Option<Vec<Row>>,
     metadata: &Arc<CassResultData>,
 ) -> Option<Vec<CassRow>> {
```
wprzytula: This function unnecessarily accepts and returns `Option`. Let's remove it from signature altogether, to decrease complexity.

wprzytula (changes requested):

muzarski:
Rebased on master

scylla-rust-wrapper/src/query_result.rs
muzarski: Actually, I have a question. Here is the definition of `size_t` in `types.rs`:
```rust
pub type size_t = cass_uint64_t;
pub type cass_uint64_t = u64;
```
Why? There is a `libc::size_t` which is an alias for `usize`. Couldn't we make use of it? Then, we wouldn't need to do a possibly lossy conversion each time (from `u64` to `usize`).

cc: @Lorak-mmk

muzarski: I believe that sizes of both `size_t` and `usize` depend on the architecture. I don't understand why we assume that `size_t` is always `u64` (8 bytes long)

Lorak-mmk: You are right of course. This alias was added probably by accident by @wprzytula in https://github.com/scylladb/cpp-rust-driver/commit/de6143af54e9140f6c33715653472bbe7b3d2644#diff-fffb385b3ac71791e0a2ddee4b04afa8c2d889e2713f4b1e4f29d216ec6045fd
Before that we already had definition for size_t generated by bindgen in build.rs:
https://github.com/scylladb/cpp-rust-driver/blob/4d056edf7b2180f06c7b4162efa805da14255cd7/scylla-rust-wrapper/build.rs#L24-L37

This generated header is included in types.rs, so it should be safe to remove this wrong alias - even more, they should crash and not compile. However for some reason I can't remove it, it causes errors. This needs investigation and proper fix.

Lorak-mmk: Generated `basic_types.rs` file is empty for some reason...

Lorak-mmk: Adding `.size_t_is_usize(false)` call to the builder makes the definition generate again:
```
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: ./target/debug/build/scylla-cpp-driver-rust-e897748878d67d0e/out/basic_types.rs
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ /* automatically generated by rust-bindgen 0.65.1 */
   2   │ 
   3   │ pub type size_t = ::std::os::raw::c_ulong;
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

muzarski: > Generated `basic_types.rs` file is empty for some reason...

Same happened for me when I was playing with `build.rs` and `types.rs`

scylla-rust-wrapper/src/query_result.rs
wprzytula: Have you done that? I can't see it.

muzarski: Sorry, I addressed it locally.

muzarski:
v4.2: addressed @wprzytula comments

I'd like to introduce the `size_t` fix (maybe next to other binding fixes) in other PRs. I'm currently investigating it.

muzarski:
v4.2.1: fixed clippy lint

muzarski:
v4.2.2: replaced `unwrap()` with `expect()` for `u64` to `usize` conversion.

wprzytula (approved):

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk: Could you apply this fix and remove the redundant `size_t` alias?

muzarski: I opened a separate PR that solves issues regarding bindings in `basic_types.rs`: https://github.com/scylladb/cpp-rust-driver/pull/188

Lorak-mmk (approved):

PR #138, opened 2024-07-08
cargo: update rust driver's version to 0.13.1

### LINKED ISSUE #126: Update Rust Driver version used to 0.13

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/126

This PR updates the rust driver's version to 0.13.1

## API breaking changes
First commit fixes some minor issues caused by API breaking changes in rust driver.

## Serialization
Following commits focus on serialization. This version of rust driver already uses a new serialization framework which validates the types. In result, some integration tests are failing, since cpp-driver uses a bit different rust-to-CQL type mapping. A simple example is a `i64` value which can be mapped to 4 CQL types: `bigint/counter/time/timestamp`.

In this PR we address and fix this issue. We introduce a `CassCqlValue` enum which is a narrower version of rust driver's `CqlValue` with a custom `SerializeValue` (still called `SerializeCql` in 0.13) implementation. The serialization DOES NOT contain any typechecks. The typechecks will be introduced in a follow-up PR. We want to follow the same approach as cpp-driver, and do the typechecks during bindings (i.e. binding value to collection, tuple, UDT or statement).

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/tuple.rs
+++ b/scylla-rust-wrapper/src/tuple.rs
@@ -3,8 +3,7 @@ use crate::binding;
 use crate::cass_error::CassError;
 use crate::cass_types::CassDataType;
 use crate::types::*;
-use scylla::frame::response::result::CqlValue;
-use std::convert::TryFrom;
+use crate::value::CassCqlValue;
 use std::sync::Arc;
 
 static EMPTY_TUPLE_TYPE: CassDataType = CassDataType::Tuple(Vec::new());
@@ -12,7 +11,7 @@ static EMPTY_TUPLE_TYPE: CassDataType = CassDataType::Tuple(Vec::new());
 #[derive(Clone)]
 pub struct CassTuple {
     pub data_type: Option<Arc<CassDataType>>,
-    pub items: Vec<Option<CqlValue>>,
+    pub items: Vec<Option<CassCqlValue>>,
 }
 
 impl CassTuple {
@@ -32,7 +31,7 @@ impl CassTuple {
     // not bind items with too high index.
     // If it was created using `cass_tuple_new_from_data_type` we additionally check if
     // value has correct type.
-    fn bind_value(&mut self, index: usize, v: Option<CqlValue>) -> CassError {
+    fn bind_value(&mut self, index: usize, v: Option<CassCqlValue>) -> CassError {
         if index >= self.items.len() {
             return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
         }
@@ -49,10 +48,9 @@ impl CassTuple {
     }
 }
 
-impl TryFrom<&CassTuple> for CqlValue {
-    type Error = CassError;
-    fn try_from(tuple: &CassTuple) -> Result<Self, Self::Error> {
-        Ok(CqlValue::Tuple(tuple.items.clone()))
+impl From<&CassTuple> for CassCqlValue {
+    fn from(tuple: &CassTuple) -> Self {
+        CassCqlValue::Tuple(tuple.items.clone())
     }
 }
 
--- a/scylla-rust-wrapper/src/user_type.rs
+++ b/scylla-rust-wrapper/src/user_type.rs
@@ -16,7 +16,7 @@ pub struct CassUserType {
 }
 
 impl CassUserType {
-    fn set_option_by_index(&mut self, index: usize, value: Option<CassCqlValue>) -> CassError {
+    fn set_field_by_index(&mut self, index: usize, value: Option<CassCqlValue>) -> CassError {
         if index >= self.field_values.len() {
             return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
         }
@@ -27,7 +27,7 @@ impl CassUserType {
         CassError::CASS_OK
     }
 
-    fn set_option_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
+    fn set_field_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
         let mut found_field: bool = false;
         for (index, (field_name, field_type)) in
             self.data_type.get_udt_type().field_types.iter().enumerate()
@@ -97,8 +97,8 @@ pub unsafe extern "C" fn cass_user_type_data_type(
 }
 
 prepare_binders_macro!(@index_and_name CassUserType,
-    |udt: &mut CassUserType, index, v| udt.set_option_by_index(index, v),
-    |udt: &mut CassUserType, name, v| udt.set_option_by_name(name, v));
+    |udt: &mut CassUserType, index, v| udt.set_field_by_index(index, v),
+    |udt: &mut CassUserType, name, v| udt.set_field_by_name(name, v));
 
 make_binders!(
     null,
--- a/scylla-rust-wrapper/src/user_type.rs
+++ b/scylla-rust-wrapper/src/user_type.rs
@@ -3,7 +3,7 @@ use crate::binding::is_compatible_type;
 use crate::cass_error::CassError;
 use crate::cass_types::CassDataType;
 use crate::types::*;
-use scylla::frame::response::result::CqlValue;
+use crate::value::CassCqlValue;
 use std::os::raw::c_char;
 use std::sync::Arc;
 
@@ -12,11 +12,11 @@ pub struct CassUserType {
     pub data_type: Arc<CassDataType>,
 
     // Vec to preserve the order of fields
-    pub field_values: Vec<Option<CqlValue>>,
+    pub field_values: Vec<Option<CassCqlValue>>,
 }
 
 impl CassUserType {
-    fn set_option_by_index(&mut self, index: usize, value: Option<CqlValue>) -> CassError {
+    fn set_option_by_index(&mut self, index: usize, value: Option<CassCqlValue>) -> CassError {
         if index >= self.field_values.len() {
             return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
         }
@@ -27,8 +27,7 @@ impl CassUserType {
         CassError::CASS_OK
     }
 
-    //TODO: some hashtable for name lookup?
-    fn set_option_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
+    fn set_option_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
         let mut found_field: bool = false;
         for (index, (field_name, field_type)) in
             self.data_type.get_udt_type().field_types.iter().enumerate()
@@ -53,9 +52,9 @@ impl CassUserType {
     }
 }
 
-impl From<&CassUserType> for CqlValue {
+impl From<&CassUserType> for CassCqlValue {
     fn from(user_type: &CassUserType) -> Self {
-        CqlValue::UserDefinedType {
+        CassCqlValue::UserDefinedType {
             keyspace: user_type.data_type.get_udt_type().keyspace.clone(),
             type_name: user_type.data_type.get_udt_type().name.clone(),
             fields: user_type
--- a/scylla-rust-wrapper/src/value.rs
+++ b/scylla-rust-wrapper/src/value.rs
@@ -4,8 +4,8 @@ use scylla::{
     frame::{response::result::ColumnType, value::CqlDate},
     serialize::{
         value::{
-            BuiltinSerializationErrorKind, SerializeCql, SetOrListSerializationErrorKind,
-            TupleSerializationErrorKind,
+            BuiltinSerializationErrorKind, MapSerializationErrorKind, SerializeCql,
+            SetOrListSerializationErrorKind, TupleSerializationErrorKind,
         },
         writers::WrittenCellProof,
         CellWriter, SerializationError,
@@ -85,7 +85,9 @@ impl CassCqlValue {
             CassCqlValue::Inet(v) => serialize_inet(v, writer),
             CassCqlValue::Tuple(fields) => serialize_tuple_like(fields.iter(), writer),
             CassCqlValue::List(l) => serialize_sequence(l.len(), l.iter(), writer),
-            CassCqlValue::Map(_) => todo!(),
+            CassCqlValue::Map(m) => {
+                serialize_mapping(m.len(), m.iter().map(|p| (&p.0, &p.1)), writer)
+            }
             CassCqlValue::Set(s) => serialize_sequence(s.len(), s.iter(), writer),
             CassCqlValue::UserDefinedType {
                 keyspace,
@@ -235,6 +237,40 @@ fn serialize_sequence<'t, 'b>(
             mk_ser_err_named(
                 rust_name,
                 SetOrListSerializationErrorKind::ElementSerializationFailed(err),
+            )
+        })?;
+    }
+
+    builder
+        .finish()
+        .map_err(|_| mk_ser_err_named(rust_name, BuiltinSerializationErrorKind::SizeOverflow))
+}
+
+fn serialize_mapping<'t, 'b>(
+    len: usize,
+    iter: impl Iterator<Item = (&'t CassCqlValue, &'t CassCqlValue)>,
+    writer: CellWriter<'b>,
+) -> Result<WrittenCellProof<'b>, SerializationError> {
+    let rust_name = std::any::type_name::<CassCqlValue>();
+
+    let mut builder = writer.into_value_builder();
+
+    let element_count: i32 = len
+        .try_into()
+        .map_err(|_| mk_ser_err_named(rust_name, MapSerializationErrorKind::TooManyElements))?;
+    builder.append_bytes(&element_count.to_be_bytes());
+
+    for (k, v) in iter {
+        k.do_serialize(builder.make_sub_writer()).map_err(|err| {
+            mk_ser_err_named(
+                rust_name,
+                MapSerializationErrorKind::KeySerializationFailed(err),
+            )
+        })?;
+        v.do_serialize(builder.make_sub_writer()).map_err(|err| {
+            mk_ser_err_named(
+                rust_name,
+                MapSerializationErrorKind::ValueSerializationFailed(err),
             )
         })?;
     }
--- a/scylla-rust-wrapper/src/value.rs
+++ b/scylla-rust-wrapper/src/value.rs
@@ -1,0 +1,57 @@
+use std::net::IpAddr;
+
+use scylla::{
+    frame::{response::result::ColumnType, value::CqlDate},
+    serialize::{value::SerializeCql, writers::WrittenCellProof, CellWriter, SerializationError},
+};
+use uuid::Uuid;
+
+/// A narrower version of rust driver's CqlValue.
+///
+/// cpp-driver's API allows to map single rust type to
+/// multiple CQL types. For example `cass_statement_bind_int64`
+/// can be used to bind a value to the column of following CQL types:
+/// - bigint
+/// - time
+/// - counter
+/// - timestamp
+/// There is no such method as `cass_statement_bind_counter`, and so
+/// we need to serialize the counter value using `CassCqlValue::BigInt`.
+#[derive(Clone, Debug, PartialEq)]
+pub enum CassCqlValue {
+    TinyInt(i8),
+    SmallInt(i16),
+    Int(i32),
+    BigInt(i64),
+    Float(f32),
+    Double(f64),
+    Boolean(bool),
+    Text(String),
+    Blob(Vec<u8>),
+    Uuid(Uuid),
+    Date(CqlDate),
+    Inet(IpAddr),
+    Tuple(Vec<Option<CassCqlValue>>),
+    List(Vec<CassCqlValue>),
+    Map(Vec<(CassCqlValue, CassCqlValue)>),
+    Set(Vec<CassCqlValue>),
+    UserDefinedType {
+        keyspace: String,
+        type_name: String,
+        /// Order of `fields` vector must match the order of fields as defined in the UDT. The
+        /// driver does not check it by itself, so incorrect data will be written if the order is
+        /// wrong.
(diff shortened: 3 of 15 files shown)


muzarski:
@Lorak-mmk @wprzytula Any ideas why the unit tests using scylla-proxy fail? In both tests, `cass_session_execute_batch` seems to return `CASS_ERROR_LIB_UNABLE_TO_CONNECT` for some weird reason.

wprzytula:
> For some reason, I could not add a driver's dependency with a standard version = "0.13" notation. I had to use a VCS path to a specific commit. The reason is that, otherwise, rust compiler would complain about scylla-rust-wrapper and scylla-proxy using different versions of DbError enum - even though, I changed the version of scylla-proxy, and it points to the commit of 0.13 release (https://github.com/scylladb/scylla-rust-driver/commit/e68adf5dcf4fb4667d37a42a628b106a028240d4).

My idea why this happens is: if you choose a specific commit for scylla-proxy, then `scylla-proxy` (together with `scylla-cql` it depends on) is fetched from GH. Choosing `version` for `scylla` instead of a commit, makes `scylla` (together with `scylla-cql` it depends on) come from crates.io, this way making those DB errors different (coming from two different source code definitions - one in GH `scylla-cql` and another in crates.io `scylla-cql`).

wprzytula:
> @Lorak-mmk @wprzytula Any ideas why the unit tests using scylla-proxy fail? In both tests, `cass_session_execute_batch` seems to return `CASS_ERROR_LIB_UNABLE_TO_CONNECT` for some weird reason.

The culprit is this: https://github.com/scylladb/scylla-rust-driver/pull/770.
Logs show (I recommend running one of those failing tests in two terminals side by side, one before the commit that updates the driver and one after) that the updated driver, having failed the initial metadata fetch, starts the re-resolution procedure. This leads to the problem. Further investigation TODO.

wprzytula:
OK, a solution is to alter the proxy rules in the test code not to drop the control connection, but instead return a DB error on attempt to fetch metadata. This way the driver does not suffer from lack of connections in the pool.

Just change this function in a following way:
```rust
    fn generic_drop_queries_rules() -> impl IntoIterator<Item = RequestRule> {
        [RequestRule(
            Condition::RequestOpcode(RequestOpcode::Query),
            // We won't respond to any queries (including metadata fetch),
            // but the driver will manage to continue with dummy metadata.
            // RequestReaction::drop_connection(), <--- this is removed
            RequestReaction::forge().server_error(), // <---- this is added instead
        )]
    }
```

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -12,11 +12,11 @@ pub struct CassUserType {
...
-    pub field_values: Vec<Option<CqlValue>>,
+    pub field_values: Vec<Option<CassCqlValue>>,
 }
 
 impl CassUserType {
-    fn set_option_by_index(&mut self, index: usize, value: Option<CqlValue>) -> CassError {
+    fn set_option_by_index(&mut self, index: usize, value: Option<CassCqlValue>) -> CassError {
         if index >= self.field_values.len() {
             return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
         }
```
wprzytula: Shouldn't it be named `set_field_by_index` ?

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -28,7 +28,7 @@ impl CassUserType {
     }
 
     //TODO: some hashtable for name lookup?
-    fn set_option_by_name(&mut self, name: &str, value: Option<CqlValue>) -> CassError {
+    fn set_option_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
         let mut found_field: bool = false;
         for (index, (field_name, field_type)) in
```
wprzytula: ditto: `set_field_by_name` ?

scylla-rust-wrapper/src/tuple.rs
```diff
@@ -49,10 +49,10 @@ impl CassTuple {
...
 }
 
-impl TryFrom<&CassTuple> for CqlValue {
+impl TryFrom<&CassTuple> for CassCqlValue {
     type Error = CassError;
     fn try_from(tuple: &CassTuple) -> Result<Self, Self::Error> {
-        Ok(CqlValue::Tuple(tuple.items.clone()))
+        Ok(CassCqlValue::Tuple(tuple.items.clone()))
     }
 }
```
wprzytula: Why is this `TryFrom` if the conversion is infallible? Let's make if `From`.

scylla-rust-wrapper/src/user_type.rs
wprzytula, line 22: This function doesn't seem to actually set anything...

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -28,7 +28,7 @@ impl CassUserType {
     }
 
     //TODO: some hashtable for name lookup?
```
wprzytula: Hashtable makes no sense here. Up to 10-100 values, iterating over an array is faster than hash lookup. IMO we can remove the TODO.

scylla-rust-wrapper/src/value.rs
```diff
@@ -2,10 +2,28 @@ use std::net::IpAddr;
...
+/// multiple CQL types. For example `cass_statement_bind_int64`
+/// can be used to bind a value to the column of following CQL types:
+/// - bigint
+/// - time
+/// - counter
+/// - timestamp
+/// There is no such method as `cass_statement_bind_counter`, and so
+/// we need to serialize the counter value using `CassCqlValue::BigInt`.
 #[derive(Clone, Debug, PartialEq)]
 pub enum CassCqlValue {
```
wprzytula: This should belong to the previous commit.

scylla-rust-wrapper/src/value.rs
```diff
@@ -72,7 +73,31 @@ impl SerializeCql for CassCqlValue {
             CassCqlValue::Uuid(v) => serialize_uuid(v, typ, writer),
             CassCqlValue::Date(v) => serialize_date(v, typ, writer),
             CassCqlValue::Inet(v) => serialize_inet(v, typ, writer),
-            CassCqlValue::Tuple(_) => todo!(),
+            CassCqlValue::Tuple(t) => {
+                // We allow serializing tuples that have less fields
+                // than the database tuple, but not the other way around.
```
wprzytula: Why so? Source? cpp-driver or Rust driver?

scylla-rust-wrapper/src/value.rs
```diff
@@ -261,3 +261,48 @@ fn serialize_tuple_like<'t, 'b>(
...
+    let elt = match typ {
+        ColumnType::List(elt) | ColumnType::Set(elt) => elt,
+        _ => {
+            return Err(mk_typck_err_named(
+                rust_name,
+                typ,
+                SetOrListTypeCheckErrorKind::NotSetOrList,
+            ));
+        }
+    };
```
wprzytula: It's moderately disappointing that we have to duplicate this logic from the Rust driver.

scylla-rust-wrapper/src/value.rs
```diff
@@ -357,3 +358,79 @@ fn serialize_mapping<'t, 'b, K: SerializeCql + 't, V: SerializeCql + 't>(
...
+            typ,
+            UdtTypeCheckErrorKind::NameMismatch {
+                keyspace: dst_keyspace.clone(),
+                type_name: dst_type_name.clone(),
+            },
+        ));
+    }
+
+    // Allow columns present in the CQL type which are not present in CqlValue,
+    // but not the other way around
```
wprzytula: CassCqlValue

scylla-rust-wrapper/src/value.rs
muzarski: Do you have a better idea on how to do it?

wprzytula: Perhaps we could somehow extract and expose the common logic in the Rust driver?

muzarski: True. Anyway, it's not something that we could fix now, as it would require a new release of rust driver. Maybe I'll leave it as TODO for now.

wprzytula: Fortunately, in cpp-rust-driver we are free to reorganize anything without fear of breaking changes, because the API is so superficial and does not limit our design choices in any way.

muzarski:
One integration test failed:
```
/home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/objects/future.hpp:152: Failure
      Expected: CASS_OK
To be equal to: wait_code
      Which is: CASS_ERROR_SERVER_INVALID_QUERY [Invalid query]
Invalid query: Database returned an error: The query is syntactically correct but invalid, Error message: Cannot include a counter statement in a logged batch
/home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/tests/test_batch.cpp:316: Failure
      Expected: number_of_rows
      Which is: 100
To be equal to: result.row_count()
      Which is: 0
[  FAILED  ] BatchCounterThreeNodeClusterTests.Integration_Cassandra_MixedPreparedAndSimple (13988 ms)
```

Isn't it weird that we get a server error that never occured before, after bumping the driver's version?
I'll try to investigate it further.

muzarski:
> One integration test failed:
> 
> ```
> /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/objects/future.hpp:152: Failure
>       Expected: CASS_OK
> To be equal to: wait_code
>       Which is: CASS_ERROR_SERVER_INVALID_QUERY [Invalid query]
> Invalid query: Database returned an error: The query is syntactically correct but invalid, Error message: Cannot include a counter statement in a logged batch
> /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/tests/test_batch.cpp:316: Failure
>       Expected: number_of_rows
>       Which is: 100
> To be equal to: result.row_count()
>       Which is: 0
> [  FAILED  ] BatchCounterThreeNodeClusterTests.Integration_Cassandra_MixedPreparedAndSimple (13988 ms)
> ```
> 
> Isn't it weird that we get a server error that never occured before, after bumping the driver's version? I'll try to investigate it further.

The issue can be reproduced in rust driver:
```rust
#[tokio::test]
async fn test_counter_batch() {
    setup_tracing();
    let session = Arc::new(create_new_session_builder().build().await.unwrap());
    let ks = unique_keyspace_name();

    session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
    session
        .query(
            format!(
                "CREATE TABLE IF NOT EXISTS {}.t_batch (key int PRIMARY KEY, value counter)",
                ks
            ),
            &[],
        )
        .await
        .unwrap();

    let q = format!("UPDATE {}.t_batch SET value = value + ? WHERE key = ?", ks);

    let mut my_batch = Batch::new(BatchType::Counter);
    assert_eq!(BatchType::Counter, my_batch.get_type());
    my_batch.append_statement(&q[..]);
    my_batch.append_statement(&q[..]);
    my_batch.append_statement(&q[..]);

    session
        .batch(
            &my_batch,
            ((Counter(1), 2), (Counter(1), 2), (Counter(1), 2)),
        )
        .await
        .unwrap();
}
```

The error:
```
---- transport::session_test::test_my_batch stdout ----
Unique name: test_rust_1720524333_0
thread 'transport::session_test::test_counter_batch' panicked at scylla/src/transport/session_test.rs:404:10:
called `Result::unwrap()` on an `Err` value: DbError(Invalid, "Cannot include a counter statement in a logged batch")
```

This seems like a rust-driver issue, since you can clearly see that I've set the `BatchType` to `BatchType::Counter`. Previous version of the driver does not have this issue.

cc: @Lorak-mmk @wprzytula

wprzytula:
> > One integration test failed:
> > ```
> > /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/objects/future.hpp:152: Failure
> >       Expected: CASS_OK
> > To be equal to: wait_code
> >       Which is: CASS_ERROR_SERVER_INVALID_QUERY [Invalid query]
> > Invalid query: Database returned an error: The query is syntactically correct but invalid, Error message: Cannot include a counter statement in a logged batch
> > /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/tests/test_batch.cpp:316: Failure
> >       Expected: number_of_rows
> >       Which is: 100
> > To be equal to: result.row_count()
> >       Which is: 0
> > [  FAILED  ] BatchCounterThreeNodeClusterTests.Integration_Cassandra_MixedPreparedAndSimple (13988 ms)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Isn't it weird that we get a server error that never occured before, after bumping the driver's version? I'll try to investigate it further.
> 
> The issue can be reproduced in rust driver:
> 
> ```rust
> #[tokio::test]
> async fn test_counter_batch() {
>     setup_tracing();
>     let session = Arc::new(create_new_session_builder().build().await.unwrap());
>     let ks = unique_keyspace_name();
> 
>     session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
>     session
>         .query(
>             format!(
>                 "CREATE TABLE IF NOT EXISTS {}.t_batch (key int PRIMARY KEY, value counter)",
>                 ks
>             ),
>             &[],
>         )
>         .await
>         .unwrap();
> 
>     let q = format!("UPDATE {}.t_batch SET value = value + ? WHERE key = ?", ks);
> 
>     let mut my_batch = Batch::new(BatchType::Counter);
>     assert_eq!(BatchType::Counter, my_batch.get_type());
>     my_batch.append_statement(&q[..]);
>     my_batch.append_statement(&q[..]);
>     my_batch.append_statement(&q[..]);
> 
>     session
>         .batch(
>             &my_batch,
>             ((Counter(1), 2), (Counter(1), 2), (Counter(1), 2)),
>         )
>         .await
>         .unwrap();
> }
> ```
> 
> The error:
> 
> ```
> ---- transport::session_test::test_my_batch stdout ----
> Unique name: test_rust_1720524333_0
> thread 'transport::session_test::test_counter_batch' panicked at scylla/src/transport/session_test.rs:404:10:
> called `Result::unwrap()` on an `Err` value: DbError(Invalid, "Cannot include a counter statement in a logged batch")
> ```
> 
> This seems like a rust-driver issue, since you can clearly see that I've set the `BatchType` to `BatchType::Counter`. Previous version of the driver does not have this issue.
> 
> cc: @Lorak-mmk @wprzytula

Please open an issue in Rust driver repo. Also, specify what version of the driver behaves correctly and what version misbehaves. If you feel like, perform `git bisect` to find the culprit.

muzarski:
@wprzytula I opened a PR in rust-driver that will fix the failing test: https://github.com/scylladb/scylla-rust-driver/pull/1038. I think that we should filter out the failing test for now, and include it again when new release of rust-driver with a fix is out (unless we can wait for new release if it happens anytime soon, but I don't think so).

scylla-rust-wrapper/src/value.rs
muzarski: This is a logic from rust-driver. I'll check how cpp-driver behaves in this case.

scylla-rust-wrapper/src/value.rs
```diff
@@ -44,3 +62,73 @@ impl SerializeCql for CassCqlValue {
...
+                $typ,
+                BuiltinTypeCheckErrorKind::MismatchedType {
+                    expected: &[$(ColumnType::$cql),*],
+                }
+            ))
+        }
+    };
+}
+
+macro_rules! impl_serialize_via_writer {
```
Lorak-mmk: The name of this macro suggests that it creates some `impl` but it creates a function.

Lorak-mmk (changes requested):
> There is a is_compatible_type function in binding module. Is it really necessary, when we already do a type checking in our SerializeCql implementation? Can this be removed?

Rust Driver's type checking happens when we serialize the values, which in cpp-rust-driver happens during a call to e.g. `cass_session_execute`.
`is_compatible_type` is called in binding functions like `cass_statement_bind_*`.

We have to either implement it at some point, or wait for Rust Driver to introduce BoundStatement and use it in cpp-rust-driver.

> For some reason, I could not add a driver's dependency with a standard version = "0.13" notation. I had to use a VCS path to a specific commit. The reason is that, otherwise, rust compiler would complain about scylla-rust-wrapper and scylla-proxy using different versions of DbError enum - even though, I changed the version of scylla-proxy, and it points to the commit of 0.13 release (https://github.com/scylladb/scylla-rust-driver/commit/e68adf5dcf4fb4667d37a42a628b106a028240d4).

Can you use a tag when specifying dependency, instead of commit hash? Main disadvantage of specifying commit hash is that it's not obvious which version it follows, but tag would solve it.

muzarski:
v2: addressed review comments

scylla-rust-wrapper/src/value.rs
muzarski: Changed it to `fn_serialize_via_writer`

scylla-rust-wrapper/src/value.rs
muzarski: `equals` implementation for Tuples in cpp-driver:
```c++
  virtual bool equals(const DataType::ConstPtr& data_type) const {
    assert(value_type() == CASS_VALUE_TYPE_TUPLE);

    if (value_type() != data_type->value_type()) {
      return false;
    }

    const ConstPtr& tuple_type(data_type);

    // Only compare sub-types if both have sub-types
    if (!types_.empty() && !tuple_type->types_.empty()) {
      if (types_.size() != tuple_type->types_.size()) {
        return false;
      }
      for (size_t i = 0; i < types_.size(); ++i) {
        if (!types_[i]->equals(tuple_type->types_[i])) {
          return false;
        }
      }
    }

    return true;
  }
```

So we should not allow serializing tuples that have less fields than the DB tuple. However, notice that this check is performed during `cass_statement_bind_tuple*`. It is something that Karol mentioned in other comment. There are no checks during serialization itself.

muzarski: As a side note: `equals` implementation for UDTs is a bit more flexible in cpp-driver:
```c++
  virtual bool equals(const DataType::ConstPtr& data_type) const {
    assert(value_type() == CASS_VALUE_TYPE_UDT);
    if (data_type->value_type() != CASS_VALUE_TYPE_UDT) {
      return false;
    }

    const UserType::ConstPtr& user_type(data_type);

    if (!equals_both_not_empty(keyspace_, user_type->keyspace_)) {
      return false;
    }

    if (!equals_both_not_empty(type_name_, user_type->type_name_)) {
      return false;
    }


    // UDT's can be considered equal as long as the mutual first fields shared
    // between them are equal. UDT's are append only as far as fields go, so a
    // newer 'version' of the UDT data type after a schema change event should be
    // treated as equivalent in this scenario, by simply looking at the first N
    // mutual fields they should share.
    size_t min_fields_size = std::min(fields_.size(), user_type->fields_.size());

    for (size_t i = 0; i < min_fields_size; ++i) {
      if (fields_[i].name != user_type->fields_[i].name ||
          !fields_[i].type->equals(user_type->fields_[i].type)) {
        return false;
      }
    }

    return true;
  }
```

muzarski:
v2.1: bumped rust driver's version to 0.13.1

scylla-rust-wrapper/src/value.rs
Lorak-mmk: Regarding the tuples: as we discussed yesterday @muzarski types and sizes are only checked for typed tuple (`cass_tuple_new_from_data_type`), not for untyped tuple (`cass_tuple_new`)

scylla-rust-wrapper/src/value.rs
```diff
@@ -55,3 +62,73 @@ impl SerializeCql for CassCqlValue {
...
     }
 }
+
+fn mk_typck_err<T>(
+    got: &ColumnType,
+    kind: impl Into<BuiltinTypeCheckErrorKind>,
+) -> SerializationError {
+    mk_typck_err_named(std::any::type_name::<T>(), got, kind)
+}
+
```
Lorak-mmk: Please add a TODO (or even better: issue here and in rust driver) to export relevant functions from scylla-cql so there is not need for coppy-paste here.

scylla-rust-wrapper/src/value.rs
Lorak-mmk: Please make sure that semantics of tuples and UDTs match those in cpp-driver

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/value.rs
wprzytula: I suggest exporting such internal utility functions shared between Rust driver and cpp-rust-driver using `#[doc(hidden)]` attribute, as this gives us freedom to modify this as non-public API.

Lorak-mmk: Or we could add a feature `cpp-rust-wrapper` and only export if it is enabled.

muzarski:
v3: as discussed with @Lorak-mmk offline, we decided to follow the same approach as cpp-driver - we don't do any typechecks during serialization. The type checks are supposed to take place during binding (e.g. binding value to tuple or statement). This approach is better, since we can separate typecheck and serialization logic - this is really handful for untyped tuples and collections (cpp-driver allows users to define such objects).

In this PR, we only introduce a naive serialization logic (custom `SerializeValue` implementation for `CassCqlValue`). During serialization step, we assume that type checks were already done earlier (as of now, this is obviously not true - typecheck logic will be implemented in a follow-up PR).

It's worth noting that we do not lose anything, since type checks were not performed by previous version of rust driver anyway.

muzarski:
Opened an issue on what should be implemented in follow-up PR with typechecks: https://github.com/scylladb/cpp-rust-driver/issues/142.

Lorak-mmk (approved):

PR #143, opened 2024-07-18
binding: typechecks

### LINKED ISSUE #142: binding: implement typechecks logic
Cpp-driver performs typechecks during binding. To be more precise, it performs typechecks when user executes:
- `cass_statement_bind_*`
- `cass_collection_append_*`
- `cass_user_type_set_*`
- `cass_tuple_set_*`

The driver also allows users to define untyped (`cass_{tuple/collection}_new`) and typed (`cass_{tuple/collection}_new_from_data_type`) versions of tuples and collections. In case of binding to untyped tuples/collections, no type checks are performed.

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/142

## Motivation

Right now, cpp-rust-driver does not perform any typechecks during binding. This PR changes that and implements a typecheck logic based on cpp-driver's logic.

### Why not just use a builtin rust-driver's typechecks mechanism?
The reason for this is that cpp-driver has a bit different language (cpp/rust) to CQL type mapping. A simple example is an int64_t/i64 value which in cpp-driver can be mapped to 4 different CQL types.

Other reason is that cpp-driver allows users to define untyped tuples/collections. For such objects, the typechecks are skipped.

Cpp-driver performs typechecks during binding - we will follow the same approach. It is much convenient and cleaner to handle untyped objects this way.

## Implemented API functions
- `cass_collection_new_from_data_type` -> allows user to create a TYPED collection. We also adjusted the `cass_collection_new` logic - this function is responsible for creating an UNTYPED version of a collection.
- `cass_collection_data_type` -> returns a data type representing a collection. Since collections now hold info about the data types, this function was very trivial to implement.

## Unit tests
I implemented unit tests for simple and complex (udt, tuples, collections) types. Please make sure that all edge cases are covered.

## Pre-review checklist

- ~[ ] I have enabled appropriate tests in `.github/workflows/cassandra.yml` in `gtest_filter`~

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,13 +173,6 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
             <td colspan=2 align="center" style="font-weight:bold">Collection</td>
         </tr>
         <tr>
-            <td>cass_collection_new_from_data_type</td>
-            <td rowspan="2">Unimplemented</td>
-        </tr>
-        <tr>
-            <td>cass_collection_data_type</td>
-        </tr>
-        <tr>
             <td>cass_collection_append_custom[_n]</td>
             <td>Unimplemented because of the same reasons as binding for statements.<br> <b>Note</b>: The driver does not check whether the type of the appended value is compatible with the type of the collection items.</td>
         </tr>
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -87,6 +87,42 @@ impl UDTDataType {
     pub fn get_field_by_index(&self, index: usize) -> Option<&Arc<CassDataType>> {
         self.field_types.get(index).map(|(_, b)| b)
     }
+
+    fn typecheck_equals(&self, other: &UDTDataType) -> bool {
+        // See: https://github.com/scylladb/cpp-driver/blob/master/src/data_type.hpp#L354-L386
+
+        if !any_string_empty_or_both_equal(&self.keyspace, &other.keyspace) {
+            return false;
+        }
+        if !any_string_empty_or_both_equal(&self.name, &other.name) {
+            return false;
+        }
+
+        // A comment from cpp-driver:
+        //// UDT's can be considered equal as long as the mutual first fields shared
+        //// between them are equal. UDT's are append only as far as fields go, so a
+        //// newer 'version' of the UDT data type after a schema change event should be
+        //// treated as equivalent in this scenario, by simply looking at the first N
+        //// mutual fields they should share.
+        //
+        // Iterator returned from zip() is perfect for checking the first mutual fields.
+        for (field, other_field) in self.field_types.iter().zip(other.field_types.iter()) {
+            // Compare field names.
+            if field.0 != other_field.0 {
+                return false;
+            }
+            // Compare field types.
+            if !field.1.typecheck_equals(&other_field.1) {
+                return false;
+            }
+        }
+
+        true
+    }
+}
+
+fn any_string_empty_or_both_equal(s1: &str, s2: &str) -> bool {
+    s1.is_empty() || s2.is_empty() || s1 == s2
 }
 
 impl Default for UDTDataType {
@@ -124,7 +160,10 @@ impl CassDataType {
     pub fn typecheck_equals(&self, other: &CassDataType) -> bool {
         match self {
             CassDataType::Value(t) => *t == other.get_value_type(),
-            CassDataType::UDT(_) => todo!(),
+            CassDataType::UDT(udt) => match other {
+                CassDataType::UDT(other_udt) => udt.typecheck_equals(other_udt),
+                _ => false,
+            },
             CassDataType::List { .. } => todo!(),
             CassDataType::Set { .. } => todo!(),
             CassDataType::Map { .. } => todo!(),
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -116,6 +116,25 @@ pub enum CassDataType {
     Custom(String),
 }
 
+impl CassDataType {
+    /// Checks for equality during typechecks.
+    ///
+    /// This takes into account the fact that tuples/collections may be untyped.
+    pub fn typecheck_equals(&self, other: &CassDataType) -> bool {
+        match self {
+            CassDataType::Value(t) => *t == other.get_value_type(),
+            CassDataType::UDT(_) => todo!(),
+            CassDataType::List { .. } => todo!(),
+            CassDataType::Set { .. } => todo!(),
+            CassDataType::Map { .. } => todo!(),
+            CassDataType::Tuple(_) => todo!(),
+            CassDataType::Custom(_) => {
+                unimplemented!("Cpp-rust-driver does not support custom types!")
+            }
+        }
+    }
+}
+
 impl From<NativeType> for CassValueType {
     fn from(native_type: NativeType) -> CassValueType {
         match native_type {
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -15,7 +15,7 @@ include!(concat!(env!("OUT_DIR"), "/cppdriver_data_types.rs"));
 include!(concat!(env!("OUT_DIR"), "/cppdriver_data_query_error.rs"));
 include!(concat!(env!("OUT_DIR"), "/cppdriver_batch_types.rs"));
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct UDTDataType {
     // Vec to preserve the order of types
     pub field_types: Vec<(String, Arc<CassDataType>)>,
@@ -95,7 +95,7 @@ impl Default for UDTDataType {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum CassDataType {
     Value(CassValueType),
     UDT(UDTDataType),
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -136,10 +136,12 @@ pub enum CassDataType {
     Value(CassValueType),
     UDT(UDTDataType),
     List {
+        // None stands for untyped list.
         typ: Option<Arc<CassDataType>>,
         frozen: bool,
     },
     Set {
+        // None stands for untyped set.
         typ: Option<Arc<CassDataType>>,
         frozen: bool,
     },
@@ -164,8 +166,21 @@ impl CassDataType {
                 CassDataType::UDT(other_udt) => udt.typecheck_equals(other_udt),
                 _ => false,
             },
-            CassDataType::List { .. } => todo!(),
-            CassDataType::Set { .. } => todo!(),
+            CassDataType::List { typ, .. } | CassDataType::Set { typ, .. } => match other {
+                CassDataType::List { typ: other_typ, .. }
+                | CassDataType::Set { typ: other_typ, .. } => {
+                    // If one of them is list, and the other is set, fail the typecheck.
+                    if self.get_value_type() != other.get_value_type() {
+                        return false;
+                    }
+                    match (typ, other_typ) {
+                        // One of them is untyped, skip the typecheck for subtype.
+                        (None, _) | (_, None) => true,
+                        (Some(typ), Some(other_typ)) => typ.typecheck_equals(other_typ),
+                    }
+                }
+                _ => false,
+            },
             CassDataType::Map { .. } => todo!(),
             CassDataType::Tuple(sub) => match other {
                 CassDataType::Tuple(other_sub) => {
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -112,6 +112,7 @@ pub enum CassDataType {
         val_type: Option<Arc<CassDataType>>,
         frozen: bool,
     },
+    // Empty vector stands for untyped tuple.
     Tuple(Vec<Arc<CassDataType>>),
     Custom(String),
 }
@@ -127,7 +128,23 @@ impl CassDataType {
             CassDataType::List { .. } => todo!(),
             CassDataType::Set { .. } => todo!(),
             CassDataType::Map { .. } => todo!(),
-            CassDataType::Tuple(_) => todo!(),
+            CassDataType::Tuple(sub) => match other {
+                CassDataType::Tuple(other_sub) => {
+                    // If either of tuples is untyped, skip the typecheck for subtypes.
+                    if sub.is_empty() || other_sub.is_empty() {
+                        return true;
+                    }
+
+                    // If both are non-empty, check for subtypes equality.
+                    if sub.len() != other_sub.len() {
+                        return false;
+                    }
+                    sub.iter()
+                        .zip(other_sub.iter())
+                        .all(|(typ, other_typ)| typ.typecheck_equals(other_typ))
+                }
+                _ => false,
+            },
             CassDataType::Custom(_) => {
                 unimplemented!("Cpp-rust-driver does not support custom types!")
             }
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -132,6 +132,13 @@ impl Default for UDTDataType {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+pub enum MapDataType {
+    Untyped,
+    Key(Arc<CassDataType>),
+    KeyAndValue(Arc<CassDataType>, Arc<CassDataType>),
+}
+
+#[derive(Clone, Debug, PartialEq)]
 pub enum CassDataType {
     Value(CassValueType),
     UDT(UDTDataType),
@@ -146,10 +153,7 @@ pub enum CassDataType {
         frozen: bool,
     },
     Map {
-        // None, None stands for untyped map.
-        // Some, None stands for a map with an untyped value type.
-        key_type: Option<Arc<CassDataType>>,
-        val_type: Option<Arc<CassDataType>>,
+        typ: MapDataType,
         frozen: bool,
     },
     // Empty vector stands for untyped tuple.
@@ -183,29 +187,22 @@ impl CassDataType {
                 }
                 _ => false,
             },
-            CassDataType::Map {
-                key_type: k,
-                val_type: v,
-                ..
-            } => match other {
-                CassDataType::Map {
-                    key_type: k_other,
-                    val_type: v_other,
-                    ..
-                } => match ((k, v), (k_other, v_other)) {
+            CassDataType::Map { typ: t, .. } => match other {
+                CassDataType::Map { typ: t_other, .. } => match (t, t_other) {
                     // See https://github.com/scylladb/cpp-driver/blob/master/src/data_type.hpp#L218
                     // In cpp-driver the types are held in a vector.
                     // The logic is following:
 
                     // If either of vectors is empty, skip the typecheck.
-                    ((None, None), _) => true,
-                    (_, (None, None)) => true,
+                    (MapDataType::Untyped, _) => true,
+                    (_, MapDataType::Untyped) => true,
 
                     // Otherwise, the vectors should have equal length and we perform the typecheck for subtypes.
-                    ((Some(k), None), (Some(k_other), None)) => k.typecheck_equals(k_other),
-                    ((Some(k), Some(v)), (Some(k_other), Some(v_other))) => {
-                        k.typecheck_equals(k_other) && v.typecheck_equals(v_other)
-                    }
+                    (MapDataType::Key(k), MapDataType::Key(k_other)) => k.typecheck_equals(k_other),
+                    (
+                        MapDataType::KeyAndValue(k, v),
+                        MapDataType::KeyAndValue(k_other, v_other),
+                    ) => k.typecheck_equals(k_other) && v.typecheck_equals(v_other),
                     _ => false,
                 },
                 _ => false,
@@ -278,16 +275,18 @@ pub fn get_column_type_from_cql_type(
                 frozen: *frozen,
             },
             CollectionType::Map(key, value) => CassDataType::Map {
-                key_type: Some(Arc::new(get_column_type_from_cql_type(
-                    key,
-                    user_defined_types,
-                    keyspace_name,
-                ))),
-                val_type: Some(Arc::new(get_column_type_from_cql_type(
-                    value,
-                    user_defined_types,
-                    keyspace_name,
-                ))),
+                typ: MapDataType::KeyAndValue(
+                    Arc::new(get_column_type_from_cql_type(
+                        key,
+                        user_defined_types,
+                        keyspace_name,
+                    )),
+                    Arc::new(get_column_type_from_cql_type(
+                        value,
+                        user_defined_types,
+                        keyspace_name,
+                    )),
+                ),
                 frozen: *frozen,
             },
             CollectionType::Set(set) => CassDataType::Set {
@@ -340,10 +339,19 @@ impl CassDataType {
                 }
             }
             CassDataType::Map {
-                key_type, val_type, ..
+                typ: MapDataType::Untyped,
+                ..
+            } => None,
+            CassDataType::Map {
+                typ: MapDataType::Key(k),
+                ..
+            } => (index == 0).then_some(k),
+            CassDataType::Map {
+                typ: MapDataType::KeyAndValue(k, v),
+                ..
             } => match index {
-                0 => key_type.as_ref(),
-                1 => val_type.as_ref(),
+                0 => Some(k),
+                1 => Some(v),
                 _ => None,
             },
             CassDataType::Tuple(v) => v.get(index),
@@ -361,17 +369,28 @@ impl CassDataType {
                 }
             },
             CassDataType::Map {
-                key_type, val_type, ..
+                typ: MapDataType::KeyAndValue(_, _),
+                ..
+            } => Err(CassError::CASS_ERROR_LIB_BAD_PARAMS),
+            CassDataType::Map {
+                typ: MapDataType::Key(k),
+                frozen,
             } => {
-                if key_type.is_some() && val_type.is_some() {
-                    Err(CassError::CASS_ERROR_LIB_BAD_PARAMS)
-                } else if key_type.is_none() {
-                    *key_type = Some(sub_type);
-                    Ok(())
-                } else {
-                    *val_type = Some(sub_type);
-                    Ok(())
-                }
+                *self = CassDataType::Map {
+                    typ: MapDataType::KeyAndValue(k.clone(), sub_type),
+                    frozen: *frozen,
+                };
+                Ok(())
+            }
+            CassDataType::Map {
+                typ: MapDataType::Untyped,
+                frozen,
+            } => {
+                *self = CassDataType::Map {
+                    typ: MapDataType::Key(sub_type),
+                    frozen: *frozen,
+                };
+                Ok(())
             }
             CassDataType::Tuple(types) => {
                 types.push(sub_type);
@@ -423,8 +442,10 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
             frozen: false,
         },
         ColumnType::Map(key, value) => CassDataType::Map {
-            key_type: Some(Arc::new(get_column_type(key.as_ref()))),
-            val_type: Some(Arc::new(get_column_type(value.as_ref()))),
+            typ: MapDataType::KeyAndValue(
+                Arc::new(get_column_type(key.as_ref())),
+                Arc::new(get_column_type(value.as_ref())),
+            ),
             frozen: false,
         },
         ColumnType::Set(boxed_type) => CassDataType::Set {
@@ -475,8 +496,7 @@ pub unsafe extern "C" fn cass_data_type_new(value_type: CassValueType) -> *const
         },
         CassValueType::CASS_VALUE_TYPE_TUPLE => CassDataType::Tuple(Vec::new()),
         CassValueType::CASS_VALUE_TYPE_MAP => CassDataType::Map {
-            key_type: None,
-            val_type: None,
+            typ: MapDataType::Untyped,
             frozen: false,
         },
         CassValueType::CASS_VALUE_TYPE_UDT => CassDataType::UDT(UDTDataType::new()),
@@ -673,9 +693,11 @@ pub unsafe extern "C" fn cass_data_type_sub_type_count(data_type: *const CassDat
         CassDataType::Value(..) => 0,
         CassDataType::UDT(udt_data_type) => udt_data_type.field_types.len() as size_t,
         CassDataType::List { typ, .. } | CassDataType::Set { typ, .. } => typ.is_some() as size_t,
-        CassDataType::Map {
-            key_type, val_type, ..
-        } => key_type.is_some() as size_t + val_type.is_some() as size_t,
+        CassDataType::Map { typ, .. } => match typ {
+            MapDataType::Untyped => 0,
+            MapDataType::Key(_) => 1,
+            MapDataType::KeyAndValue(_, _) => 2,
+        },
         CassDataType::Tuple(v) => v.len() as size_t,
         CassDataType::Custom(..) => 0,
     }
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -146,6 +146,8 @@ pub enum CassDataType {
         frozen: bool,
     },
     Map {
+        // None, None stands for untyped map.
+        // Some, None stands for a map with an untyped value type.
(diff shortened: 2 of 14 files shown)


muzarski:
v1.1: Added a link to cpp-driver's UDT typecheck logic in a comment.

roydahan (approved):

scylla-rust-wrapper/src/value.rs
```diff
@@ -89,7 +89,16 @@ impl CassCqlValue {
...
+            CassCqlValue::Text(_) => {
+                matches!(
+                    typ.get_value_type(),
+                    CassValueType::CASS_VALUE_TYPE_TEXT
+                        | CassValueType::CASS_VALUE_TYPE_VARCHAR
+                        | CassValueType::CASS_VALUE_TYPE_ASCII
+                        | CassValueType::CASS_VALUE_TYPE_BLOB
+                        | CassValueType::CASS_VALUE_TYPE_VARINT
+                )
+            }
```
Lorak-mmk, line 118: cpp-driver also includes `CASS_VALUE_TYPE_CUSTOM`. I assume we are skipping support for this. Plase not this in the commit message here and in the commit about blobs.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -116,6 +116,26 @@ pub enum CassDataType {
...
+            CassDataType::Value(t) => *t == other.get_value_type(),
+            CassDataType::UDT(_) => todo!(),
+            CassDataType::List { .. } => todo!(),
+            CassDataType::Set { .. } => todo!(),
+            CassDataType::Map { .. } => todo!(),
+            CassDataType::Tuple(_) => todo!(),
+            CassDataType::Custom(name) => match other {
+                CassDataType::Custom(other_name) => name == other_name,
+                _ => false,
+            },
```
Lorak-mmk: Here I see that there is some handling for custom types, so my previous comment may be incorrect.
Do we want to support them?

Lorak-mmk: It will probably become clear in the next commit, but this one in isolation is not clear to me. Could you explain in comment or commit message (or both) why do we need `typecheck_equals`? Previous commits already implemented type checking, so it's not clear to me as a reviewer where this is going to be used.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -112,6 +112,7 @@ pub enum CassDataType {
         val_type: Option<Arc<CassDataType>>,
         frozen: bool,
     },
+    // Empty vector stands for untyped tuple.
```
Lorak-mmk, line 159: I know that this PR didn't introduce this, but I wonder if this is correct.
Is it possible to have an empty tuple in Scylla? If so, it could be a problem because this code won't differentiate between those 2 cases.

README.md
Lorak-mmk, line 1: Regarding commit `collection: include info about data type `: there is typo in the message.
`cass_collection_new_from_datat_type` should be `cass_collection_new_from_data_type`

scylla-rust-wrapper/src/collection.rs
```diff
@@ -32,8 +32,49 @@ pub struct CassCollection {
...
+                    ..
+                } => {
+                    // Only do the typecheck if both map types are present.
+                    if let (Some(k_typ), Some(v_typ)) = (k_typ, v_typ) {
+                        if index % 2 == 0 && !value::is_type_compatible(value, k_typ) {
+                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                        }
+                        if !value::is_type_compatible(value, v_typ) {
+                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                        }
```
Lorak-mmk: This condition looks wrong. If `index % 2 == 0` and `value` is compatible with `k_typ` then we move to second if and require that it is also compatible with `v_typ`.

Didn't tests catch this? Please make sure they do.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cass_types.rs
muzarski: I was surprised that there is a CassDataType::Custom variant as well. If we don't plan on supporting it, we should remove it I think. AFAIR, custom types are not planned to be supported in the Rust driver.

scylla-rust-wrapper/src/collection.rs
muzarski: Good catch, I'll make sure to implement the corresponding test case.

scylla-rust-wrapper/src/cass_types.rs
muzarski: This is used for complex types (udts, tuples, collections) in `value::is_type_compatible`. I'll include this information in the commit message.

muzarski:
v1.2: rebased on master

scylla-rust-wrapper/src/cass_types.rs
muzarski: > Is it possible to have an empty tuple in Scylla?

It doesn't seem like it:
image

I also checked creating an empty tuple type (`tuple<>`) using cqlsh. Scylla rejects such type.

muzarski:
v2: addressed some review comments

What should we do with the custom types? @Lorak-mmk

Lorak-mmk:
I think we should either support them fully (which would require some changes in Rust Driver, probably not big ones) or not support them at all - remove them from enums, return errors as early as possible when encountering them somewhere. WDYT @wprzytula ?

muzarski:
Rebased on master (duration)

muzarski:
Rebased on master (decimal)

scylla-rust-wrapper/src/cass_types.rs
muzarski: I opened an issue: https://github.com/scylladb/cpp-rust-driver/issues/158

muzarski: I'll remove mentions of custom types in a separate PR.

scylla-rust-wrapper/src/value.rs
muzarski: Done

muzarski:
v1.3: added mention about not supporting custom types in Text and Blob commits.

Lorak-mmk (approved):

scylla-rust-wrapper/src/value.rs
```diff
@@ -297,3 +297,73 @@ fn serialize_udt<'b>(
...
+            from(CassValueType::CASS_VALUE_TYPE_UDT),
+        ]
+    }
+
+    #[test]
+    fn typecheck_simple_test() {
+        struct TestCase {
+            pub value: Option<CassCqlValue>,
+            pub compatible_types: Vec<CassDataType>,
+        }
```
wprzytula: `pub` fields in non-`pub` struct make no sense.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -116,6 +116,26 @@ pub enum CassDataType {
     Custom(String),
 }
```
wprzytula, line 162: Typo in commit message of "types: introduce typecheck_equals method":
`value::is_type_compatbile` -> `value::is_type_compatible`

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -87,6 +87,42 @@ impl UDTDataType {
...
+            }
+        }
+
+        true
+    }
+}
+
+fn equals_both_non_empty_strings(s1: &str, s2: &str) -> bool {
+    s1.is_empty() || s2.is_empty() || s1 == s2
 }
```
wprzytula: The name is counterintuitive. Maybe `any_string_empty_or_both_equal`?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -146,6 +146,8 @@ pub enum CassDataType {
         frozen: bool,
     },
     Map {
+        // None, None stands for untyped map.
+        // Some, None stands for a map with an untyped value type.
         key_type: Option<Arc<CassDataType>>,
         val_type: Option<Arc<CassDataType>>,
```
wprzytula: What about `None, Some` ? Is it valid state for a Map? If not, perhaps we should make it irrepresentable in our Map type?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -181,7 +183,33 @@ impl CassDataType {
...
+                    // If either of vectors is empty, skip the typecheck.
+                    ((None, None), _) => true,
+                    (_, (None, None)) => true,
+
+                    // Otherwise, the vectors should have equal length and we perform the typecheck for subtypes.
+                    ((Some(k), None), (Some(k_other), None)) => k.typecheck_equals(k_other),
+                    ((Some(k), Some(v)), (Some(k_other), Some(v_other))) => {
+                        k.typecheck_equals(k_other) && v.typecheck_equals(v_other)
+                    }
+                    _ => false,
```
wprzytula: It seems that we indeed treat `(None, Some)` state as invalid. I suggest storing the key and value types in a specific enum:

```rust
enum MapType {
  Untyped,
  Key(CqlDataType),
  KeyAndValue(CqlDataType),
}
```

scylla-rust-wrapper/src/value.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
         }
     }
+
+    #[test]
+    fn typecheck_complex_test() {
+        struct TestCase {
+            pub value: CassCqlValue,
+            pub compatible_types: Vec<CassDataType>,
+            pub incompatible_types: Vec<CassDataType>,
+        }
```
wprzytula: Again, `pub` fields of non-`pub` struct.

scylla-rust-wrapper/src/tuple.rs
```diff
@@ -50,7 +50,10 @@ impl CassTuple {
 
 impl From<&CassTuple> for CassCqlValue {
     fn from(tuple: &CassTuple) -> Self {
-        CassCqlValue::Tuple(tuple.items.clone())
+        CassCqlValue::Tuple {
+            data_type: tuple.data_type.clone(),
+            fields: tuple.items.clone(),
+        }
     }
 }
```
wprzytula, line 58: Not related to this PR:
Is it customary to implement `From` on references? Especially if this involves cloning.
What I mean is that `From` is the opposite side of `Into`, and `Into` is for _owned -> owned_ conversions. See [Rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

scylla-rust-wrapper/src/value.rs
```diff
@@ -525,8 +534,8 @@ mod tests {
     fn typecheck_complex_test() {
         struct TestCase {
             pub value: CassCqlValue,
-            pub compatible_types: Vec<CassDataType>,
-            pub incompatible_types: Vec<CassDataType>,
+            pub compatible_types: Vec<Arc<CassDataType>>,
+            pub incompatible_types: Vec<Arc<CassDataType>>,
         }
```
wprzytula: This should belong to the previous commit.

scylla-rust-wrapper/src/value.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
+    #[test]
+    fn typecheck_complex_test() {
+        struct TestCase {
+            pub value: CassCqlValue,
+            pub compatible_types: Vec<CassDataType>,
+            pub incompatible_types: Vec<CassDataType>,
+        }
+
+        let run_test_cases = |test_cases: Vec<TestCase>| {
+            for case in test_cases {
```
wprzytula: Nit: you can use statically allocated slices instead of `Vec` for storing test cases.

scylla-rust-wrapper/src/collection.rs
```diff
@@ -57,18 +75,61 @@ pub unsafe extern "C" fn cass_collection_new(
 ) -> *mut CassCollection {
     let capacity = match collection_type {
         // Maps consist of a key and a value, so twice
-        // the number of CqlValue will be stored.
+        // the number of CassCqlValue will be stored.
```
wprzytula, line 131: I think the previous formulation was correct. If `CqlValue::Map` has capacity `n`, then `CassCollection` functioning as a Map has capacity `2n`.

scylla-rust-wrapper/src/collection.rs
```diff
@@ -57,18 +75,61 @@ pub unsafe extern "C" fn cass_collection_new(
...
+    let collection_ref = ptr_to_ref(collection);
+
+    match &collection_ref.data_type {
+        Some(dt) => Arc::as_ptr(dt),
+        None => match collection_ref.collection_type {
+            CassCollectionType::CASS_COLLECTION_TYPE_LIST => &UNTYPED_LIST_TYPE,
+            CassCollectionType::CASS_COLLECTION_TYPE_SET => &UNTYPED_SET_TYPE,
+            CassCollectionType::CASS_COLLECTION_TYPE_MAP => &UNTYPED_MAP_TYPE,
+            // CassCollectionType is a C enum. Can user provide a value out of range?
+            _ => unreachable!(),
```
wprzytula: User can provide a value out of range, but then it's going to be their fault. That's how C contract on UB works...

scylla-rust-wrapper/src/collection.rs
```diff
@@ -1,14 +1,32 @@
...
+};
+static UNTYPED_SET_TYPE: CassDataType = CassDataType::Set {
+    typ: None,
+    frozen: false,
+};
+static UNTYPED_MAP_TYPE: CassDataType = CassDataType::Map {
+    key_type: None,
+    val_type: None,
+    frozen: false,
+};
```
wprzytula: Great idea! Love it!

scylla-rust-wrapper/src/collection.rs
```diff
@@ -1,14 +1,32 @@
...
 
+/// These constants help us to save an allocation in case user calls `cass_collection_new` (untyped collection).
+static UNTYPED_LIST_TYPE: CassDataType = CassDataType::List {
+    typ: None,
+    frozen: false,
+};
+static UNTYPED_SET_TYPE: CassDataType = CassDataType::Set {
+    typ: None,
+    frozen: false,
+};
```
wprzytula: Nit: this probably shouldn't be a docstring, because it would only document the first of the constants.

scylla-rust-wrapper/src/collection.rs
```diff
@@ -57,18 +75,61 @@ pub unsafe extern "C" fn cass_collection_new(
...
+    item_count: size_t,
+) -> *mut CassCollection {
+    let data_type = clone_arced(data_type);
+    let (capacity, collection_type) = match data_type.as_ref() {
+        CassDataType::List { .. } => (item_count, CassCollectionType::CASS_COLLECTION_TYPE_LIST),
+        CassDataType::Set { .. } => (item_count, CassCollectionType::CASS_COLLECTION_TYPE_SET),
+        // Maps consist of a key and a value, so twice
+        // the number of CassCqlValue will be stored.
+        CassDataType::Map { .. } => (item_count * 2, CassCollectionType::CASS_COLLECTION_TYPE_MAP),
+        _ => return std::ptr::null_mut(),
```
wprzytula, line 156: Does cpp driver silently return a NULL pointer on bad data type passed? If so, that's sad...

scylla-rust-wrapper/src/collection.rs
```diff
@@ -32,8 +32,49 @@ pub struct CassCollection {
...
+            match data_type.as_ref() {
+                CassDataType::List { typ: subtype, .. }
+                | CassDataType::Set { typ: subtype, .. } => match subtype {
+                    Some(subtype) => {
+                        if !value::is_type_compatible(value, subtype) {
+                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                        }
+                    }
+                    None => {}
+                },
```
wprzytula: `if let` ?

wprzytula (changes requested):

muzarski:
Rebased on master

scylla-rust-wrapper/src/collection.rs
muzarski: I only changed `CqlValue` -> `CassCqlValue`.

scylla-rust-wrapper/src/collection.rs
muzarski: I think we should panic here then (with some insightful error message). I wouldn't silently return a null pointer.

scylla-rust-wrapper/src/tuple.rs
muzarski: Sure, I can change it to `to_cass_cql_value` method. But I'd do it in a follow-up PR since such `From` implementation appears in multiple places AFAIR.

scylla-rust-wrapper/src/cass_types.rs
muzarski: Introduced `MapDataType` enum (last commit) which represents valid states of map data type.

muzarski:
v3: addressed review comments.

scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk: If we are not going to support Custom types then I see no point in introducing code related to it. Please replace this check with some panic.

scylla-rust-wrapper/src/tuple.rs
Lorak-mmk: And of course this implementation is my fault :( https://github.com/scylladb/cpp-rust-driver/commit/db1a9a7f610f070ad60a3175cec4a9b11f579542

Lorak-mmk (changes requested):
One small comment, otherwise looks good to me

scylla-rust-wrapper/src/tuple.rs
muzarski: I opened an issue to remember about it: https://github.com/scylladb/cpp-rust-driver/issues/177

scylla-rust-wrapper/src/cass_types.rs
muzarski: Done

Lorak-mmk (approved):

scylla-rust-wrapper/src/user_type.rs
```diff
@@ -20,7 +19,7 @@ impl CassUserType {
         if index >= self.field_values.len() {
             return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
         }
-        if !is_compatible_type(&self.data_type.get_udt_type().field_types[index].1, &value) {
+        if !value::is_type_compatible(&value, &self.data_type.get_udt_type().field_types[index].1) {
```
wprzytula, line 22: Why did you decide to swap the parameters?

scylla-rust-wrapper/src/value.rs
```diff
@@ -297,3 +297,73 @@ fn serialize_udt<'b>(
...
+            for typ in all_simple_types.iter() {
+                let result = is_type_compatible(&case.value, typ);
+                let expected = case.compatible_types.iter().any(|t| t == typ);
+                assert_eq!(
+                    expected, result,
+                    "Typecheck test for value {:?} and type {:?} failed.",
+                    case.value, typ
+                );
+            }
+        }
```
wprzytula: Let's enrich the assertion so that it includes info whether the value was deemed type compatible when it should not or the opposite.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -127,7 +128,23 @@ impl CassDataType {
             CassDataType::List { .. } => todo!(),
             CassDataType::Set { .. } => todo!(),
             CassDataType::Map { .. } => todo!(),
-            CassDataType::Tuple(_) => todo!(),
+            CassDataType::Tuple(sub) => match other {
+                CassDataType::Tuple(other_sub) => {
+                    // If either of tuples is untyped, skip the typecheck for subtypes.
+                    if sub.is_empty() || other_sub.is_empty() {
+                        return true;
+                    }
```
wprzytula, line 215: IIUC, tuples are untyped iff their type is empty?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -116,6 +116,23 @@ pub enum CassDataType {
...
+    /// This takes into account the fact that tuples/collections may be untyped.
+    pub fn typecheck_equals(&self, other: &CassDataType) -> bool {
+        match self {
+            CassDataType::Value(t) => *t == other.get_value_type(),
+            CassDataType::UDT(_) => todo!(),
+            CassDataType::List { .. } => todo!(),
+            CassDataType::Set { .. } => todo!(),
+            CassDataType::Map { .. } => todo!(),
+            CassDataType::Tuple(_) => todo!(),
+            CassDataType::Custom(_) => panic!("Cpp-rust-driver does not support custom types!"),
```
wprzytula: Maybe let's use `unimplemented!` instead of general `panic!`? This conveys our idea specifically.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -87,6 +87,42 @@ impl UDTDataType {
...
+        if !any_string_empty_or_both_equal(&self.name, &other.name) {
+            return false;
+        }
+
+        // A comment from cpp-driver:
+        //// UDT's can be considered equal as long as the mutual first fields shared
+        //// between them are equal. UDT's are append only as far as fields go, so a
+        //// newer 'version' of the UDT data type after a schema change event should be
+        //// treated as equivalent in this scenario, by simply looking at the first N
+        //// mutual fields they should share.
```
wprzytula, line 106: So it resembles Rust driver `SerializeValue` derive macro's logic with `force_exact_match=false`, correct? @Lorak-mmk

scylla-rust-wrapper/src/value.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
+        let run_test_cases = |test_cases: &[TestCase]| {
+            for case in test_cases {
+                for typ in case.compatible_types.iter() {
+                    assert!(
+                        case.value.is_type_compatible(typ),
+                        "Typecheck for failed, when it should pass. Value: {:?}, Type: {:?}",
+                        case.value,
+                        typ
+                    );
+                }
```
wprzytula: `Typecheck for failed` ?

scylla-rust-wrapper/src/value.rs
```diff
@@ -55,8 +55,7 @@ pub enum CassCqlValue {
     Map(Vec<(CassCqlValue, CassCqlValue)>),
     Set(Vec<CassCqlValue>),
     UserDefinedType {
-        keyspace: String,
-        type_name: String,
+        data_type: Arc<CassDataType>,
```
wprzytula, line 67: Why do we store a `CassDataType` inside a `UserDefinedType`? Isn't that enum too broad for just a UDT?

scylla-rust-wrapper/src/collection.rs
```diff
@@ -58,11 +60,15 @@ impl TryFrom<&CassCollection> for CassCqlValue {
...
 
-                Ok(CassCqlValue::Map(grouped_items))
-            }
-            CassCollectionType::CASS_COLLECTION_TYPE_SET => {
-                Ok(CassCqlValue::Set(collection.items.clone()))
+                Ok(CassCqlValue::Map {
+                    data_type,
+                    values: grouped_items,
+                })
             }
```
wprzytula, line 114: Oh, you've fixed a bug here, correct? Map collections were made `CassCqlValue::Set` ...

scylla-rust-wrapper/src/value.rs
```diff
@@ -130,9 +139,27 @@ impl CassCqlValue {
...
+                }
+                // Untyped map.
+                typ.get_value_type() == CassValueType::CASS_VALUE_TYPE_MAP
+            }
+            CassCqlValue::Set { data_type, .. } => {
+                if let Some(dt) = data_type {
+                    return dt.typecheck_equals(typ);
+                }
+                // Untyped set.
+                typ.get_value_type() == CassValueType::CASS_VALUE_TYPE_SET
```
wprzytula: I'd strongly prefer to see this without such imperative `return`s, because such `return`s usually signify an abnormal return (an error, or a shortcut) and here that's no the case. Let's use functional patterns.

wprzytula:
```suggestion
            CassCqlValue::List { data_type, .. } => {
                if let Some(dt) = data_type {
                    dt.typecheck_equals(typ)
                } else {
                    // Untyped list.
                    typ.get_value_type() == CassValueType::CASS_VALUE_TYPE_LIST
                }
            }
            CassCqlValue::Map { data_type, .. } => {
                if let Some(dt) = data_type {
                    dt.typecheck_equals(typ)
                } else {
                    // Untyped map.
                    typ.get_value_type() == CassValueType::CASS_VALUE_TYPE_MAP
                }
            }
            CassCqlValue::Set { data_type, .. } => {
                if let Some(dt) = data_type {
                    dt.typecheck_equals(typ)
                } else {
                    // Untyped set.
                    typ.get_value_type() == CassValueType::CASS_VALUE_TYPE_SET
                }
```

scylla-rust-wrapper/src/value.rs
```diff
@@ -725,5 +756,110 @@ mod tests {
...
+                        data_type_float.clone(),
+                        data_type_int.clone(),
+                        data_type_bool.clone(),
+                        data_type_untyped_set.clone(),
+                        data_type_float_set.clone(),
+                        data_type_untyped_map.clone(),
+                        data_type_typed_key_float_map.clone(),
+                        data_type_float_int_map.clone(),
+                    ],
+                },
```
wprzytula: Let's add an incompatible list type, e.g. `data_type_int_list`.

scylla-rust-wrapper/src/value.rs
```diff
@@ -725,5 +756,110 @@ mod tests {
...
+                        data_type_untyped_list.clone(),
+                        data_type_untyped_set.clone(),
+                        data_type_float_set.clone(),
+                        data_type_float_int_map.clone(),
+                    ],
+                },
+            ];
+
+            run_test_cases(test_cases)
+        }
```
wprzytula: Tests for untyped map and fully typed map are missing. Similarly, tests for set are missing.

scylla-rust-wrapper/src/collection.rs
```diff
@@ -32,8 +32,48 @@ pub struct CassCollection {
...
+                    // Only do the typecheck if both map types are present.
+                    if let (Some(k_typ), Some(v_typ)) = (k_typ, v_typ) {
+                        if index % 2 == 0 && !value::is_type_compatible(value, k_typ) {
+                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                        }
+                        if index % 2 != 0 && !value::is_type_compatible(value, v_typ) {
+                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                        }
+                    }
+                }
```
wprzytula: So you say that we shouldn't check key type compatibility in case of a key-only-typed map?

scylla-rust-wrapper/src/collection.rs
```diff
@@ -163,3 +203,172 @@ make_binders!(decimal, cass_collection_append_decimal);
...
+                );
+                assert_cass_error_eq!(
+                    cass_collection_append_float(untyped_map, 42.42),
+                    CassError::CASS_OK
+                );
+                cass_collection_free(untyped_map);
+            }
+
+            // typed map
+            {
```
wprzytula: What about a half-typed (key-only-typed) map?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -44,13 +44,30 @@ pub struct CassStatement {
...
-            CassError::CASS_OK
+            return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS;
+        }
+        let err = self.typecheck_on_bind(index, &value);
+        if err != CassError::CASS_OK {
+            return err;
         }
+
+        self.bound_values[index] = Set(value);
+        CassError::CASS_OK
```
wprzytula: I dislike that we perform indexing, which may panic. Let's try to refactor this so that we match use `get()` method on slice and pattern match on its result. The pattern matching should involve both `self.bound_values` and `p.variable_col_data_types`, because we index both.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -338,10 +337,19 @@ impl CassDataType {
...
             }
             CassDataType::Map {
-                key_type, val_type, ..
+                typ: MapDataType::Untyped,
+                ..
+            } => None,
+            CassDataType::Map {
+                typ: MapDataType::Key(k),
+                ..
+            } if index == 0 => Some(k),
```
wprzytula: To save the readers from reasoning about remaining, non-exhausted patterns, let's do it this way:

wprzytula:
```suggestion
            CassDataType::Map {
                typ: MapDataType::Key(k),
                ..
            } => (index == 0).then_some(k),
```

scylla-rust-wrapper/src/collection.rs
```diff
@@ -47,19 +46,17 @@ impl CassCollection {
...
-                            return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
-                        }
+                    if index % 2 == 0 && !value::is_type_compatible(value, k_typ) {
+                        return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                    }
+                    if index % 2 != 0 && !value::is_type_compatible(value, v_typ) {
+                        return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
                     }
                 }
                 _ => return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE,
```
wprzytula: Bug spotted! Before this commit, untyped and half-typed map would be accepted, and now they are rejected (they don't match `MapDataType::KeyAndValue`).

wprzytula (changes requested):

scylla-rust-wrapper/src/collection.rs
muzarski: They wouldn't be accepted - previous version included a check for this:
```rust
// Only do the typecheck if both map types are present.
if let (Some(k_typ), Some(v_typ)) = (k_typ, v_typ) {
```

scylla-rust-wrapper/src/collection.rs
muzarski: I know that this is confusing (since we allow users to define such map types), but this is the cpp-driver's logic:
https://github.com/scylladb/cpp-driver/blob/master/src/collection.hpp#L100-L110

muzarski: The `if (data_type_->types().size() == 2` part implies that both types need to be present to perform a typecheck on bind.

scylla-rust-wrapper/src/collection.rs
muzarski: Sorry, I misunderstood. You are right - `match` expression now rejects such map types

muzarski: I'll fix it and introduce additional test cases. Good catch!

muzarski:
Rebased on master

scylla-rust-wrapper/src/user_type.rs
muzarski: Since I moved it to `value` module, I thought that it makes more sense to have a value as first parameter.

scylla-rust-wrapper/src/value.rs
muzarski: That's true - `UDTDataType` would be sufficient. However, this way it's much easier to call `CassDataType::typecheck_equals`. In addition, we would need to clone whole `UDTDataType` since it's owned by a shared reference to `CassDataType` (see `From<&CassUserType> for CassCqlValue` implementation). As of now, we only need to clone an Arc<CassDataType>, which is much cheaper operation.

scylla-rust-wrapper/src/collection.rs
muzarski: Noo, I think it's just a confusing diff generated by git.

scylla-rust-wrapper/src/value.rs
muzarski: Typed maps and sets are included in `[in]compatible_types` for some other types. But I'll introduce separate test cases for them as well.

scylla-rust-wrapper/src/collection.rs
wprzytula: Do we prefer to be mimic cpp-driver as tight as possible in this case, or maybe let's be more eager and do the type check of the key, if present, to save the user from errors?

muzarski: To be honest, I think this is a bug in a cpp-driver. We should not mimic this behaviour

scylla-rust-wrapper/src/statement.rs
muzarski: I refactored the code

muzarski:
v4: addressed @wprzytula comments. Thanks for spotting the bug! I introduced a lot of additional test cases which consider more combinations for typechecks

wprzytula (approved):

PR #146, opened 2024-07-30
types: implement API functions for decimal type

muzarski:
This PR implements following set of API functions related to `decimal/varint` types:

- cass_statement_bind_decimal_*
- cass_user_type_set_decimal_*
- cass_tuple_set_decimal
- cass_value_get_decimal
- cass_value_is_decimal
- cass_collection_append_decimal

The other way (I think it's preferred, looking at the integration tests) to bind `varint` values is treating them as bytes (i.e. cass_statement_bind_bytes). The only thing that is missing is to implement `cass_value_get_bytes`'s support for varint type. This PR does that as well. Thanks to this, we can enable both decimal and varint integration test cases.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -136,9 +136,8 @@ macro_rules! make_appender {
     }
 }
 
-// TODO: Types for which binding is not implemented yet:
+// Types for which binding is not implemented:
 // custom - Not implemented in Rust driver
-// decimal
 
 macro_rules! invoke_binder_maker_macro_with_type {
     (null, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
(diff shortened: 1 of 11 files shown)


scylla-rust-wrapper/src/value.rs
```diff
@@ -191,6 +196,15 @@ fn_serialize_via_writer!(serialize_inet, IpAddr, |me, writer| {
...
 });
+fn_serialize_via_writer!(serialize_decimal, CqlDecimal, |me, writer| {
+    let mut builder = writer.into_value_builder();
+    let (bytes, scale) = me.as_signed_be_bytes_slice_and_exponent();
+    builder.append_bytes(&scale.to_be_bytes());
+    builder.append_bytes(bytes);
+    builder
+        .finish()
+        .map_err(|_| mk_ser_err::<CqlDecimal>(BuiltinSerializationErrorKind::SizeOverflow))?
+});
```
Lorak-mmk: Is this PR up to date? AFAIK there is no longer need for that, now we use existing `SerializeValue` impls.

muzarski: Now it is. I rebased on master

Lorak-mmk (approved):

muzarski:
Rebased on master

Lorak-mmk (approved):

scylla-rust-wrapper/src/binding.rs
```diff
@@ -291,6 +290,19 @@ macro_rules! invoke_binder_maker_macro_with_type {
...
     };
+    (decimal, $macro_name:ident, $this:ty, $consume_v:expr, $fn:ident) => {
+        $macro_name!(
+            $this,
+            $consume_v,
+            $fn,
+            |v, v_size, scale| {
+                use scylla::frame::value::CqlDecimal;
+                let varint_vec = std::slice::from_raw_parts(v, v_size as usize).to_vec();
+                Ok(Some(Decimal(CqlDecimal::from_signed_be_bytes_and_exponent(varint_vec, scale))))
```
wprzytula: Nit: `from_signed_be_bytes_slice_and_exponent` constructor is preferred, as it accepts a slice and creates a vec internally.

wprzytula (approved):

Issue #42, opened 2021-11-18
CassFuture missing functions


Lorak-mmk:
Function list:

- [x] `cass_future_wait_timed`
- [x] `cass_future_tracing_id`
- [ ] `cass_future_coordinator`

First one is easy.
Remaining two also should be easy, if Rust implements this.

muzarski:
`cass_future_coordinator` waits for https://github.com/scylladb/scylla-rust-driver/issues/1030

muzarski:
Closing in favor of https://github.com/scylladb/cpp-rust-driver/issues/249

PR #147, opened 2024-07-31
future: cass future wait timed

muzarski:
ref: https://github.com/scylladb/cpp-rust-driver/issues/42

This PR implements a `cass_future_wait_timed` API function which allows to wait for a future to resolve within a given period of time.

## Pre-review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,11 +166,8 @@ The driver inherits almost all the features of C/C++ and Rust drivers, such as:
             <td colspan=2 align="center" style="font-weight:bold">Future</td>
         </tr>
         <tr>
-            <td>cass_future_wait_timed</td>
-            <td rowspan="3">Unimplemented</td>
-        </tr>
-        <tr>
             <td>cass_future_coordinator</td>
+            <td>Unimplemented</td>
         </tr>
         <tr>
             <td colspan=2 align="center" style="font-weight:bold">Collection</td>
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -412,10 +412,14 @@ pub unsafe extern "C" fn cass_future_tracing_id(
 
 #[cfg(test)]
 mod tests {
-    use crate::testing::assert_cass_future_error_message_eq;
+    use crate::testing::{assert_cass_error_eq, assert_cass_future_error_message_eq};
 
     use super::*;
-    use std::{os::raw::c_char, thread, time::Duration};
+    use std::{
+        os::raw::c_char,
+        thread::{self},
+        time::Duration,
+    };
 
     // This is not a particularly smart test, but if some thread is granted access the value
     // before it is truly computed, then weird things should happen, even a segfault.
@@ -473,6 +477,99 @@ mod tests {
             assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
 
             cass_future_free(cass_fut);
+        }
+    }
+
+    // This test checks whether the future callback is executed correctly when:
+    // - a future is awaited indefinitely
+    // - a future is awaited, after the timeout appeared (_wait_timed)
+    // - a future is not awaited. We simply sleep, and let the tokio runtime resolve
+    //   the future, and execute its callback
+    #[test]
+    #[ntest::timeout(600)]
+    fn test_cass_future_callback() {
+        unsafe {
+            const ERROR_MSG: &str = "NOBODY EXPECTED SPANISH INQUISITION";
+            const HUNDRED_MILLIS_IN_MICROS: u64 = 100 * 1000;
+
+            let create_future_and_flag = || {
+                unsafe extern "C" fn mark_flag_cb(_fut: *const CassFuture, data: *mut c_void) {
+                    let flag = data as *mut bool;
+                    *flag = true;
+                }
+
+                let fut = async move {
+                    tokio::time::sleep(Duration::from_micros(HUNDRED_MILLIS_IN_MICROS)).await;
+                    Err((CassError::CASS_OK, ERROR_MSG.into()))
+                };
+                let cass_fut = CassFuture::make_raw(fut);
+                let flag = Box::new(false);
+                let flag_ptr = Box::into_raw(flag);
+
+                assert_cass_error_eq!(
+                    cass_future_set_callback(cass_fut, Some(mark_flag_cb), flag_ptr as *mut c_void),
+                    CassError::CASS_OK
+                );
+
+                (cass_fut, flag_ptr)
+            };
+
+            // Callback executed after awaiting.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+                cass_future_wait(cass_fut);
+
+                assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
+                assert!(*flag_ptr);
+
+                cass_future_free(cass_fut);
+                let _ = Box::from_raw(flag_ptr);
+            }
+
+            // Future awaited via `assert_cass_future_error_message_eq`.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
+                assert!(*flag_ptr);
+
+                cass_future_free(cass_fut);
+                let _ = Box::from_raw(flag_ptr);
+            }
+
+            // Callback executed after timeouts.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                // This should timeout on tokio::time::timeout.
+                let timed_result = cass_future_wait_timed(cass_fut, HUNDRED_MILLIS_IN_MICROS / 5);
+                assert_eq!(0, timed_result);
+                // This should timeout as well.
+                let timed_result = cass_future_wait_timed(cass_fut, HUNDRED_MILLIS_IN_MICROS / 5);
+                assert_eq!(0, timed_result);
+
+                // Await and check result.
+                assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
+                assert!(*flag_ptr);
+
+                cass_future_free(cass_fut);
+                let _ = Box::from_raw(flag_ptr);
+            }
+
+            // Don't await the future. Just sleep.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                RUNTIME.block_on(async {
+                    tokio::time::sleep(Duration::from_micros(HUNDRED_MILLIS_IN_MICROS + 10 * 1000))
+                        .await
+                });
+
+                assert!(*flag_ptr);
+
+                cass_future_free(cass_fut);
+                let _ = Box::from_raw(flag_ptr);
+            }
         }
     }
 }
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -270,6 +270,8 @@ pub unsafe extern "C" fn cass_future_tracing_id(
 
 #[cfg(test)]
 mod tests {
+    use crate::testing::assert_cass_future_error_message_eq;
+
     use super::*;
     use std::{os::raw::c_char, thread, time::Duration};
 
@@ -293,17 +295,11 @@ mod tests {
         unsafe {
             let handle = thread::spawn(move || {
                 let wrapper = wrapped_cass_fut;
-                let cass_fut = wrapper.0;
-                let mut message: *const c_char = std::ptr::null();
-                let mut msg_len: size_t = 0;
-                cass_future_error_message(cass_fut, &mut message, &mut msg_len);
-                assert_eq!(ptr_to_cstr_n(message, msg_len), Some(ERROR_MSG));
+                let PtrWrapper(cass_fut) = wrapper;
+                assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
             });
 
-            let mut message: *const c_char = std::ptr::null();
-            let mut msg_len: size_t = 0;
-            cass_future_error_message(cass_fut, &mut message, &mut msg_len);
-            assert_eq!(ptr_to_cstr_n(message, msg_len), Some(ERROR_MSG));
+            assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
 
             handle.join().unwrap();
             cass_future_free(cass_fut);
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -108,7 +108,7 @@ impl CassFuture {
         self.with_waited_state(|s| f(s.value.as_mut().unwrap()))
     }
 
-    pub(self) fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
+    fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
         let mut guard = self.state.lock().unwrap();
         let handle = guard.join_handle.take();
         if let Some(handle) = handle {
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -7,6 +7,7 @@ use crate::query_result::CassResult;
 use crate::types::*;
 use crate::uuid::CassUuid;
 use crate::RUNTIME;
+use futures::future;
 use scylla::prepared_statement::PreparedStatement;
 use std::future::Future;
 use std::mem;
@@ -62,7 +63,12 @@ pub struct CassFuture {
 /// An error that can appear during `cass_future_wait_timed`.
 enum FutureError {
     TimeoutError,
+    InvalidDuration,
 }
+
+/// The timeout appeared when we tried to await `JoinHandle`.
+/// This errors contains the original handle, so it can be awaited later again.
+struct JoinHandleTimeout(JoinHandle<()>);
 
 impl CassFuture {
     pub fn make_raw(
@@ -114,22 +120,46 @@ impl CassFuture {
         self.with_waited_state(|s| f(s.value.as_mut().unwrap()))
     }
 
+    /// Awaits the future until completion.
+    ///
+    /// There are two possible cases:
+    /// - noone is currently working on the future -> we take the ownership
+    ///   of JoinHandle (future) and we poll it until completion.
+    /// - some other thread is working on the future -> we wait on the condition
+    ///   variable to get an access to the future's state. Once we are notified,
+    ///   there are two cases:
+    ///     - JoinHandle is consumed -> some other thread already resolved the future.
+    ///       We can return.
+    ///     - JoinHandle is Some -> some other thread was working on the future, but
+    ///       timed out (see [CassFuture::with_waited_state_timed]). We need to
+    ///       take the ownership of the handle, and complete the work.
     fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
         let mut guard = self.state.lock().unwrap();
-        let handle = guard.join_handle.take();
-        if let Some(handle) = handle {
-            mem::drop(guard);
-            // unwrap: JoinError appears only when future either panic'ed or canceled.
-            RUNTIME.block_on(handle).unwrap();
-            guard = self.state.lock().unwrap();
-        } else {
-            guard = self
-                .wait_for_value
-                .wait_while(guard, |state| state.value.is_none())
-                // unwrap: Error appears only when mutex is poisoned.
-                .unwrap();
+        loop {
+            let handle = guard.join_handle.take();
+            if let Some(handle) = handle {
+                mem::drop(guard);
+                // unwrap: JoinError appears only when future either panic'ed or canceled.
+                RUNTIME.block_on(handle).unwrap();
+                guard = self.state.lock().unwrap();
+            } else {
+                guard = self
+                    .wait_for_value
+                    .wait_while(guard, |state| {
+                        state.value.is_none() && state.join_handle.is_none()
+                    })
+                    // unwrap: Error appears only when mutex is poisoned.
+                    .unwrap();
+                if guard.join_handle.is_some() {
+                    // join_handle was none, and now it isn't - some other thread must
+                    // have timed out and returned the handle. We need to take over
+                    // the work of completing the future. To do that, we go into
+                    // another iteration so that we land in the branch with block_on.
+                    continue;
+                }
+            }
+            return f(&mut guard);
         }
-        f(&mut guard)
     }
 
     fn with_waited_result_timed<T>(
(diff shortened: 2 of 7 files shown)


scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +125,43 @@ impl CassFuture {
         f(&mut guard)
     }
 
+    pub(self) fn with_waited_result_timed<T>(
+        &self,
+        f: impl FnOnce(&mut CassFutureResult) -> T,
+        timeout_duration: Duration,
+    ) -> Result<T, ()> {
+        self.with_waited_state_timed(|s| f(s.value.as_mut().unwrap()), timeout_duration)
+    }
```
wprzytula: Hmm, what's the advantage of `pub(self)` over no visibility qualifier?

scylla-rust-wrapper/src/future.rs
```diff
@@ -308,4 +356,38 @@ mod tests {
...
+
+            // This should timeout on cond variable (previous call consumed JoinHandle).
+            let timed_result = cass_future_wait_timed(cass_fut, HUNDRED_MILLIS_IN_MICROS / 5);
+            assert_eq!(0, timed_result);
+
+            // Verify that future eventually resolves, even though timeouts occurred before.
+            let mut message: *const c_char = std::ptr::null();
+            let mut msg_len: size_t = 0;
+            cass_future_error_message(cass_fut, &mut message, &mut msg_len);
+            assert_eq!(ptr_to_cstr_n(message, msg_len), Some(ERROR_MSG));
```
wprzytula: Maybe it would come in handy to have a utility function/macro to do that, similar to `assert_cass_error_eq!` in `testing.rs`?

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +125,43 @@ impl CassFuture {
...
+        timeout_duration: Duration,
+    ) -> Result<T, ()> {
+        let mut guard = self.state.lock().unwrap();
+        let handle = guard.join_handle.take();
+        if let Some(handle) = handle {
+            mem::drop(guard);
+            // Need to wrap it with async{} block, so the timeout is lazily executed inside the runtime.
+            // See mention about panics: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html.
+            let timed = async { tokio::time::timeout(timeout_duration, handle).await };
+            RUNTIME.block_on(timed).map_err(|_| ())?.unwrap();
```
wprzytula: For a moment I was afraid that we can lose callbacks with `tokio::timeout`, but then I understood that the callback is invoked inside a spawned tokio task, which is completed no matter if the `JoinHandle` is awaited until completion or dropped. So I think this is correct.

Do we have unit tests for callbacks in various scenarios? If not, I encourage you to add such to this PR, for all three cases:
- when the future is awaited indefinitely,
- when the future is awaited with a timeout,
- when the future is not awaited by any `cass_future_*` function.

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +125,43 @@ impl CassFuture {
...
+            guard = self.state.lock().unwrap();
+        } else {
+            let wait_result = self
+                .wait_for_value
+                .wait_timeout_while(guard, timeout_duration, |state| state.value.is_none())
+                .unwrap();
+            let timeout_result = wait_result.1;
```
wprzytula: Please prefer tuple deconstruction more than index access - it's usually more readable, as it adds naming for all tuple members.

wprzytula:
```suggestion
            let (elem1, timeout_result) = self
                .wait_for_value
                .wait_timeout_while(guard, timeout_duration, |state| state.value.is_none())
                .unwrap();
```

scylla-rust-wrapper/src/future.rs
muzarski: > ... inside a spawned tokio task, which is completed no matter if the `JoinHandle` is awaited until completion or dropped.

Yes, this is really important. I am not sure, how `tokio::time::timeout` behaves in case of an arbitrary future (whether it is polled up until completion, or not). But in this specific case we receive a `JoinHandle` future from `tokio::runtime::Runtime::spawn`, and as documentation states, this future will be executed upon calling `spawn` (doesn't need to be await'ed):
```
The provided future will start running in the background immediately when spawn is called, even if you don’t await the returned JoinHandle
```

I added the test case that checks that this future will indeed resolve eventually.

I'll add the test cases for callbacks, as you mentioned.

wprzytula: > I am not sure, how tokio::time::timeout behaves in case of an arbitrary future (whether it is polled upon completion, or not)

AFAIK, it cancels the underlying future, so it is abandoned and not completed.

scylla-rust-wrapper/src/future.rs
muzarski: There is no advantage. AFAIK they are equivalent. I just wanted to be consistent, since other methods are marked like that.

wprzytula: If so, let's remove those markers from all of them. They bring no use, just confusion.

muzarski:
Rebased on master.

muzarski:
The async test I enabled started to fail (even when I only rebased on master). I can't reproduce it locally:
- for scylla tests, my local version of ccm for some reason can't create a 5.0.0 scylla cluster
- for cassandra - the test passes

I'll update an integration test to make use of `cass_future_wait` instead of `cass_future_wait_timed` to see if the issue is with the timed version itself, or not.

scylla-rust-wrapper/src/future.rs
muzarski: Implemented tests for callbacks

muzarski:
I changed `wait_timed` to `wait` in integration test. There are multiple client timeout errors:
```
 /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/objects/future.hpp:152: Failure
      Expected: CASS_OK
To be equal to: wait_code
      Which is: CASS_ERROR_LIB_REQUEST_TIMED_OUT [Request timed out]
Request timed out: Request timeout: Request took longer than 12000ms: deadline has elapsed
```
Link to the job: https://github.com/scylladb/cpp-rust-driver/actions/runs/10201719963/job/28224244570?pr=153#step:7:8477.

Recently, we merged a PR that sets a default client timeout to `12000ms` (which is huge!). I'm not sure why the timeouts appear - these are simple insert queries. Link to the test case: https://github.com/scylladb/cpp-rust-driver/blob/master/tests/src/integration/tests/test_async.cpp#L99-L111.

I'll try to investigate further, but it seems that the test failure is not related to `cass_future_wait_timed`. The test fails due to client request timeouts. As I mentioned previously, the test passes for me locally. Maybe it's some GH actions issue - query timing out after 12s is really weird.

cc: @Lorak-mmk @wprzytula

Lorak-mmk:
> The async test I enabled started to fail (even when I only rebased on master). I can't reproduce it locally:
> 
>     * for scylla tests, my local version of ccm for some reason can't create a 5.0.0 scylla cluster

What is the error? also we should update to newer version / test with more than 1 version.

> 
>     * for cassandra - the test passes
> 
> 
> I'll update an integration test to make use of `cass_future_wait` instead of `cass_future_wait_timed` to see if the issue is with the timed version itself, or not.

Lorak-mmk:
> I changed `wait_timed` to `wait` in integration test. There are multiple client timeout errors:
> 
> ```
>  /home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/objects/future.hpp:152: Failure
>       Expected: CASS_OK
> To be equal to: wait_code
>       Which is: CASS_ERROR_LIB_REQUEST_TIMED_OUT [Request timed out]
> Request timed out: Request timeout: Request took longer than 12000ms: deadline has elapsed
> ```
> 
> Link to the job: https://github.com/scylladb/cpp-rust-driver/actions/runs/10201719963/job/28224244570?pr=153#step:7:8477.
> 
> Recently, we merged a PR that sets a default client timeout to `12000ms` (which is huge!). I'm not sure why the timeouts appear - these are simple insert queries. Link to the test case: https://github.com/scylladb/cpp-rust-driver/blob/master/tests/src/integration/tests/test_async.cpp#L99-L111.
> 
> I'll try to investigate further, but it seems that the test failure is not related to `cass_future_wait_timed`. The test fails due to client request timeouts. As I mentioned previously, the test passes for me locally. Maybe it's some GH actions issue - query timing out after 12s is really weird.
> 
> cc: @Lorak-mmk @wprzytula

So previously there was no client request timeout at all? And now that you introduced it, queries are timing out?
Tests are running in valgrind, so it's possible for queries to take a longer time, but it shouldn't be this long.
When running locally, do you also use valgrind?

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +125,42 @@ impl CassFuture {
...
+        timeout_duration: Duration,
+    ) -> Result<T, ()> {
+        let mut guard = self.state.lock().unwrap();
+        let handle = guard.join_handle.take();
+        if let Some(handle) = handle {
+            mem::drop(guard);
+            // Need to wrap it with async{} block, so the timeout is lazily executed inside the runtime.
+            // See mention about panics: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html.
+            let timed = async { tokio::time::timeout(timeout_duration, handle).await };
+            RUNTIME.block_on(timed).map_err(|_| ())?.unwrap();
```
Lorak-mmk: The only error that can be returned is `Elapsed` from Tokio, right? Then why are you mapping it to unit type instead of returning it directly?

Lorak-mmk: Ah I see, it's probably because you use `Err(())` below. Can't you return `[Elapsed](https://docs.rs/tokio/latest/tokio/time/error/struct.Elapsed.html#)` there too?

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +125,42 @@ impl CassFuture {
...
+    ) -> Result<T, ()> {
+        self.with_waited_state_timed(|s| f(s.value.as_mut().unwrap()), timeout_duration)
+    }
+
+    pub(self) fn with_waited_state_timed<T>(
+        &self,
+        f: impl FnOnce(&mut CassFutureState) -> T,
+        timeout_duration: Duration,
+    ) -> Result<T, ()> {
+        let mut guard = self.state.lock().unwrap();
```
Lorak-mmk: Instead of having both `with_waited_state_timed` and `with_waited_state`, sharing most of the code, can we just have one of them, accepting optional timeout as argument?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/future.rs
muzarski: I wanted to, but there is no public constructor for `Elapsed`, and its fields are private (actually the only field is `()`, but it's private, so we can't construct `Elapsed` outside of the crate).

muzarski:
> So previously there was no client request timeout at all? And now that you introduced it, queries are timing out? Tests are running in valgrind, so it's possible for queries to take a longer time, but it shouldn't be this long. When running locally, do you also use valgrind?

Good thing that you noticed that. I missed it. Indeed, when I run the tests locally under valgrind, the timeouts appear.

muzarski:
> > The async test I enabled started to fail (even when I only rebased on master). I can't reproduce it locally:
> > ```
> > * for scylla tests, my local version of ccm for some reason can't create a 5.0.0 scylla cluster
> > ```
> 
> What is the error?

Nvm, it's not ccm issue. I was just missing `openjdk-11` locally which is needed to setup `5.0.0` cluster apparently.

>  also we should update to newer version / test with more than 1 version.

What version would you suggest? 5.4.8?

scylla-rust-wrapper/src/future.rs
muzarski: I wanted to implement it like this. However, notice that in one case we call `Condvar::wait_while`, while in other case we call `Condvar::wait_timeout_while`. It introduces another `if` statement and more branching to the function. I decided to keep it simple - I think two distinct methods are more readable.

Lorak-mmk:
> > > The async test I enabled started to fail (even when I only rebased on master). I can't reproduce it locally:
> > > ```
> > > * for scylla tests, my local version of ccm for some reason can't create a 5.0.0 scylla cluster
> > > ```
> > 
> > 
> > What is the error?
> 
> Nvm, it's not ccm issue. I was just missing `openjdk-11` locally which is needed to setup `5.0.0` cluster apparently.

I recommend using ccm by nix, it removed a lot of my frustrations with ccm.

> 
> > also we should update to newer version / test with more than 1 version.
> 
> What version would you suggest? 5.4.8?

6.0

scylla-rust-wrapper/src/future.rs
Lorak-mmk: This duplication also has an advantage: `with_waited_state` doesn't return result, so it's easier to use.
I'm not a fan of this duplication, but I don't see how to nicely unify some parts of those functions, so let's leave it as is.

scylla-rust-wrapper/src/future.rs
Lorak-mmk: In that case can you create a zero-size struct `TimeoutError` and return it? It will be more descriptive that empty tuple and will take just 1 additional line of code.

muzarski: Done

muzarski:
v2: addressed review comments

There are still timeouts appearing. The conclusion is that the timeouts appear due to tests being run under `valgrind` (reproducing locally as well). I see two solutions here:
- Run the timeout-sensitive tests in a separate CI step without `valgrind` (right now, there is only one failing test due to timeouts).
- Modify the failing test so it disables the global client timeout, and waits for the future indefinitely (change `wait_timed` to `wait`).

I think the former approach is better, as it allows us to test `cass_future_wait_timed`. WDYT? @Lorak-mmk @wprzytula

Lorak-mmk:
> v2: addressed review comments
> 
> There are still timeouts appearing. The conclusion is that the timeouts appear due to tests being run under `valgrind` (reproducing locally as well). I see two solutions here:
> 
>     * Run the timeout-sensitive tests in a separate CI step without `valgrind` (right now, there is only one failing test due to timeouts).
> 
>     * Modify the failing test so it disables the global client timeout, and waits for the future indefinitely (change `wait_timed` to `wait`).
> 
> 
> I think the former approach is better, as it allows us to test `cass_future_wait_timed`. WDYT? @Lorak-mmk @wprzytula

In second approach, if you use wait instead of wait_timed, then how does it test wait_timed?

muzarski:
> > v2: addressed review comments
> > There are still timeouts appearing. The conclusion is that the timeouts appear due to tests being run under `valgrind` (reproducing locally as well). I see two solutions here:
> > ```
> > * Run the timeout-sensitive tests in a separate CI step without `valgrind` (right now, there is only one failing test due to timeouts).
> > 
> > * Modify the failing test so it disables the global client timeout, and waits for the future indefinitely (change `wait_timed` to `wait`).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think the former approach is better, as it allows us to test `cass_future_wait_timed`. WDYT? @Lorak-mmk @wprzytula
> 
> In second approach, if you use wait instead of wait_timed, then how does it test wait_timed?

That's the point - it does not. That's why I prefer to run this specific test without valgrind.

Lorak-mmk:
Sorry, I misread your message. 
Separate step complicates the CI, but in this case it seems reasonable to me.

muzarski:
v3: running the failing async test in a separate CI step without valgrind

I'll update Scylla version in CI in a separate PR.

muzarski:
v3.1: deduplicate error message variable in callback test

muzarski:
v3.2: fix CI step description for timeout-sensitive cassandra integration tests

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +128,42 @@ impl CassFuture {
...
+        timeout_duration: Duration,
+    ) -> Result<T, TimeoutError> {
+        let mut guard = self.state.lock().unwrap();
+        let handle = guard.join_handle.take();
+        if let Some(handle) = handle {
+            mem::drop(guard);
+            // Need to wrap it with async{} block, so the timeout is lazily executed inside the runtime.
+            // See mention about panics: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html.
+            let timed = async { tokio::time::timeout(timeout_duration, handle).await };
+            RUNTIME.block_on(timed).map_err(|_| TimeoutError)?.unwrap();
```
wprzytula: I see that the original non-timed function misses it as well, but please: explain and justify the `unwrap`.

scylla-rust-wrapper/src/future.rs
```diff
@@ -124,6 +128,42 @@ impl CassFuture {
...
+            guard = self.state.lock().unwrap();
+        } else {
+            let (guard_result, timeout_result) = self
+                .wait_for_value
+                .wait_timeout_while(guard, timeout_duration, |state| state.value.is_none())
+                .unwrap();
```
wprzytula: ditto: `unwrap`.

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +349,116 @@ mod tests {
...
+            assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
+
+            cass_future_free(cass_fut);
+        }
+    }
+
+    #[test]
+    #[ntest::timeout(500)]
+    fn test_cass_future_callback() {
+        unsafe {
```
wprzytula: Please add a comment explaining what this test checks and in what scenarios.

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +349,116 @@ mod tests {
...
+                cass_future_free(cass_fut);
+                let _ = Box::from_raw(flag_ptr);
+            }
+
+            // Don't await the future. Just sleep.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                std::thread::sleep(Duration::from_micros(HUNDRED_MILLIS_IN_MICROS + 10 * 1000));
+
```
wprzytula: Won't this fail on current-thread tokio runtime?

wprzytula (changes requested):

muzarski:
Rebased on master

scylla-rust-wrapper/src/future.rs
muzarski: Oh, good catch. I'm not sure if it will fail (but I believe it should fail, lemme check that out of curiosity). I should make use of `tokio::time::sleep` instead.

muzarski: Good point, we should make use of `tokio::time::sleep`, so the thread can execute tasks scheduled to the runtime in the meantime.

Actually, this led me to yet another issue. Notice that currently, we wait on cond variable in some cases. This is a blocking operation, and thus, this will not work in `current-thread` runtime. I pushed a new commit, which modifies the logic, and makes use of `tokio::runtime::Runtime::spawn_blocking`, which is designed to spawn the tasks that are supposed to be blocking operations. In this commit, I also (temporarily) changed the runtime to `current-thread`, to check if the tests are passing.

As you can see, these changes required some `Arc` cloning, and so I am not really sure if we want to change it, or leave it as is. I need your opinion (cc: @Lorak-mmk). If we decide to be compatible with `current-thread` runtime, I'll reorder the commits (and revert the runtime to be the default one - multi-threaded) before merging.

muzarski: Unit tests are passing, but it seems that these changes do not work for integration tests. I'll need to investigate it..

Lorak-mmk: I think that `with_waited_state` (and your new variant) don't need those changes. 
If we are in the branch that waits on condvar, that means that either:
- The future is a ready future, so wait will finish immediately.
- Other thread already claimed the join_handle and is running the future using `block_on`.

You'd need to use block_on here if there was some other spawned task that won't finish without us polling it and that is required for our code to progress - that is not the case here.
The only thing running (apart from other unrelated tasks which we don't need to worry about here) during your new call to block_on is your blocking feature spawned few lines above.

muzarski: > You'd need to use block_on here if there was some other spawned task that won't finish without us polling it and that is required for our code to progress - that is not the case here. The only thing running (apart from other unrelated tasks which we don't need to worry about here) during your new call to block_on is your blocking feature spawned few lines above.

Consider a following scenario:
- we make use of a `current-thread` tokio runtime (`Builder::new_current_thread().enable_all().build()`)
- we create a future which sleeps, let's say, 2s and then sets the value
- we call `cass_future_wait_timed` on this future, with a timeout 1s. This consumes the handle, starts the future, times out after 1s (there is still a task to be polled until completion - the future from above which sets the value)
- now we call `cass_future_wait` on the same future. This waits on cond variable (blocking operation). Now current thread is blocked, it is waiting (on a cond variable) for a value to be set. But there is noone to poll the future that sets the value, since we are using a `current-thread` runtime - deadlock...

This is my understanding of `current-thread` runtime. Please, correct me if I'm wrong. Unit tests reproduced this issue - there indeed was a deadlock (only for `current-thread` runtime obviously).

Lorak-mmk: My first thought was that handle should be returned after timeout, but that won't work if second call starts before first timeouts. I don't have any idea how to solve it right now, I'll think tomorrow.

Lorak-mmk: This is possible by passing the work to another waiting thread after timeout. 
Relevant snippet:
```rust
enum FutureError {
    TimeoutError,
    InvalidDuration,
}

struct JoinHandleTimeout(JoinHandle<()>);

    pub(self) fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
        let mut guard = self.state.lock().unwrap();
        loop {
            let handle = guard.join_handle.take();
            if let Some(handle) = handle {
                mem::drop(guard);
                RUNTIME.block_on(handle).unwrap();
                guard = self.state.lock().unwrap();
            } else {
                guard = self
                    .wait_for_value
                    .wait_while(guard, |state| {
                        state.value.is_none() && state.join_handle.is_none()
                    })
                    .unwrap();
                if guard.join_handle.is_some() {
                    // join_handle was none, and now it isn't - some other thread must
                    // have timed out and returned the handle. We need to take over
                    // the work of completing the feature. To do that, we go into
                    // another iteration so that we land in the branch with block_on.
                    continue;
                }
            }
            return f(&mut guard);
        }
    }

    pub(self) fn with_waited_state_timed<T>(
        &self,
        f: impl FnOnce(&mut CassFutureState) -> T,
        timeout_duration: Duration,
    ) -> Result<T, FutureError> {
        let mut guard = self.state.lock().unwrap();
        let deadline = Instant::now()
            .checked_add(timeout_duration)
            .ok_or(FutureError::InvalidDuration)?;
        loop {
            let handle = guard.join_handle.take();
            if let Some(handle) = handle {
                mem::drop(guard);
                // Need to wrap it with async{} block, so the timeout is lazily executed inside the runtime.
                // See mention about panics: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html.
                let timed = async {
                    let sleep_future =
                        tokio::time::sleep_until(tokio::time::Instant::from(deadline));
                    pin_mut!(sleep_future);
                    let value = future::select(handle, sleep_future).await;
                    match value {
                        Either::Left((result, _)) => Ok(result),
                        Either::Right((_, handle)) => Err(JoinHandleTimeout(handle)),
                    }
                };
                match RUNTIME.block_on(timed) {
                    Err(JoinHandleTimeout(returned_handle)) => {
                        // We timed out. so we can't finish waiting for the future.
                        // The problem is that if current thread executor is used,
                        // then no one will run this future - other threads will
                        // go into the branch with condvar and wait there.
                        // To fix that:
                        //  - Return the join handle, so that next thread can take it
                        //  - Signal one thread, so that if all other consumers are
                        //    already waiting on condvar, one of them wakes up and
                        //    picks up the work.
                        guard = self.state.lock().unwrap();
                        guard.join_handle = Some(returned_handle);
                        self.wait_for_value.notify_one();
                        return Err(FutureError::TimeoutError);
                    }
                    Ok(result) => result.unwrap(),
                    _ => todo!(),
                };
                guard = self.state.lock().unwrap();
            } else {
                let remaining_timeout = deadline.duration_since(Instant::now());
                let (guard_result, timeout_result) = self
                    .wait_for_value
                    .wait_timeout_while(guard, remaining_timeout, |state| {
                        state.value.is_none() && state.join_handle.is_none()
                    })
                    .unwrap();
                if timeout_result.timed_out() {
                    return Err(FutureError::TimeoutError);
                }
                guard = guard_result;
                if guard.join_handle.is_some() {
                    // join_handle was none, and now it isn't - some other thread must
                    // have timed out and returned the handle. We need to take over
                    // the work of completing the feature. To do that, we go into
                    // another iteration so that we land in the branch with block_on.
                    continue;
                }
            }

            return Ok(f(&mut guard));
        }
    }
```

Lorak-mmk: I missed a `_ => todo!(),` in one place, it can be safely removed.

muzarski:
Rebased on master

muzarski:
v4: added proper synchronization between `cass_future_wait` and `cass_future_wait_timed` callers.

scylla-rust-wrapper/src/future.rs
```diff
@@ -108,20 +120,112 @@ impl CassFuture {
         self.with_waited_state(|s| f(s.value.as_mut().unwrap()))
     }
 
-    pub(self) fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
+    fn with_waited_state<T>(&self, f: impl FnOnce(&mut CassFutureState) -> T) -> T {
```
wprzytula: These functions got fairly complicated. Can you add docstrings for them, as well as more comments inside them, explaining what is going on?

scylla-rust-wrapper/src/future.rs
```diff
@@ -108,20 +120,112 @@ impl CassFuture {
...
+                    })
+                    // unwrap: Error appears only when mutex is poisoned.
+                    .unwrap();
+                if guard.join_handle.is_some() {
+                    // join_handle was none, and now it isn't - some other thread must
+                    // have timed out and returned the handle. We need to take over
+                    // the work of completing the feature. To do that, we go into
+                    // another iteration so that we land in the branch with block_on.
+                    continue;
+                }
```
wprzytula: Good explanation. However, let's make sure that a reader of this function is aware from the very beginning (i.e. from its docstring) that there is another - timed - logic for accessing and executing the underlying future, which can interact nonobviously with this - nontimed - logic.

scylla-rust-wrapper/src/future.rs
```diff
@@ -108,20 +120,112 @@ impl CassFuture {
...
+            }
+
+            return Ok(f(&mut guard));
         }
```
wprzytula: I'm in awe of this logic. Great job!

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +413,124 @@ mod tests {
...
 
             handle.join().unwrap();
             cass_future_free(cass_fut);
         }
     }
+
+    // This test makes sure that the future resolves even if timeout happens.
+    //
+    // Tokio's runtime documentation states that Runtime::spawn starts
+    // running a future immediately, even if the future was not await'ed before.
```
wprzytula: Is this true for current thread executor, too?

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +413,124 @@ mod tests {
...
+                (cass_fut, flag_ptr)
+            };
+
+            // Callback executed after awaiting.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+                cass_future_wait(cass_fut);
+
+                assert_cass_future_error_message_eq!(cass_fut, Some(ERROR_MSG));
+                assert!(*flag_ptr);
```
wprzytula, line 523: If we remove `cass_future_wait` from here, should the code remain equivalent? If so, let's add another test case without it.

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +413,124 @@ mod tests {
...
+
+            // Callback executed after timeouts.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                // This should timeout on tokio::time::timeout.
+                let timed_result = cass_future_wait_timed(cass_fut, HUNDRED_MILLIS_IN_MICROS / 5);
+                assert_eq!(0, timed_result);
+                // This should timeout on cond variable (previous call consumed JoinHandle).
+                let timed_result = cass_future_wait_timed(cass_fut, HUNDRED_MILLIS_IN_MICROS / 5);
```
wprzytula: Is the comment still relevant? After the changes introduced, the `JoinHandle` is no longer consumed upon timeout, but returned for other blocking calls to be able to await it.

scylla-rust-wrapper/src/future.rs
```diff
@@ -293,19 +413,124 @@ mod tests {
...
+            }
+
+            // Don't await the future. Just sleep.
+            {
+                let (cass_fut, flag_ptr) = create_future_and_flag();
+
+                RUNTIME.block_on(async {
+                    tokio::time::sleep(Duration::from_micros(HUNDRED_MILLIS_IN_MICROS + 10 * 1000))
+                        .await
+                });
```
wprzytula, line 566: IIUC, on current thread runtime, it's only upon `RUNTIME.block_on` when spawned tasks start to be executed. Am I correct?

wprzytula (changes requested):

scylla-rust-wrapper/src/future.rs
muzarski: Credits to @Lorak-mmk

scylla-rust-wrapper/src/future.rs
muzarski: Yeah, it's because `assert_cass_future_error_message_eq` calls `cass_future_wait` under the hood.

scylla-rust-wrapper/src/future.rs
muzarski: Yes. Thanks to that, when the thread "sleeps", it actually works on a task scheduled earlier (our test future).

scylla-rust-wrapper/src/future.rs
muzarski: I think I should rephrase it. What I meant by that is, this specific future does not have to be await'ed. As long as there is someone working on the tasks, he will be working on resolving this future as well (even if this specific future was not awaited). I believe that in `current-thread` runtime, the thread needs to be inside `Runtime::block_on` call (but it does not have to be called with this specific task/future).

This is why we use `Runtime::block_on(tokio::time::sleep)` in callback test. We don't await the future there, but it is resolved eventually.

wprzytula: 100%

muzarski:
v4.1:
- rebased on master -> adjusted to new CI

README.md
Lorak-mmk, line 1: Comment regarding commit `future: cass_future_wait_timed `:
It says `Next commit will fix that,`, but the next commit is [cargo: add futures 0.3 dependency](https://github.com/scylladb/cpp-rust-driver/pull/147/commits/7eb222f6e85b1d1cd549868597932f966a442fa1). The commit after that is the one that fixes the problem.

Lorak-mmk (changes requested):

wprzytula (approved):

muzarski:
v4.2:
- rebased on master
- adjusted commit message that @Lorak-mmk mentioned

Lorak-mmk (approved):

PR #152, opened 2024-07-31
ser: use rust-driver's builtin serialization for simple types

muzarski:
Previously, we would reimplement a serialization logic. It was a copy from rust driver, with the typechecks excluded.

In this PR, we get rid of the repeated code. To skip the typechecks (or rather, to make the typechecks always pass), we simply provide a valid ColumnType to SerializeValue::serialize().

This is a kind of hack, but thanks to that we don't have to reimplement serialization logic.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/value.rs
+++ b/scylla-rust-wrapper/src/value.rs
@@ -73,18 +73,50 @@ impl CassCqlValue {
         writer: CellWriter<'b>,
     ) -> Result<WrittenCellProof<'b>, SerializationError> {
         match self {
-            CassCqlValue::TinyInt(v) => serialize_i8(v, writer),
-            CassCqlValue::SmallInt(v) => serialize_i16(v, writer),
-            CassCqlValue::Int(v) => serialize_i32(v, writer),
-            CassCqlValue::BigInt(v) => serialize_i64(v, writer),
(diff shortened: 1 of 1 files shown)


wprzytula (approved):

Lorak-mmk (approved):
Isn't it possible to remove CassCqlValue entirely?

muzarski:
> Isn't it possible to remove CassCqlValue entirely?

I think it is possible. The main obstacle are udts, collections and tuples. We would need to construct a valid `ColumnType` for each of these, and a nested `ColumnType` may be expensive to construct (some allocations). I think that there is no other way to use a builtin rust-driver's serialization with typechecks skipped for these types.

Lorak-mmk:
Ok, let's keep the approach from this PR

muzarski:
Actually, I think we could simply have a wrapper over `CqlValue`.

muzarski:
> Actually, I think we could simply have a wrapper over `CqlValue`.

Give me a second, I'll investigate it and update this PR in case it's possible.

muzarski:
Nvm, I don't think there is much sense to introduce a wrapper over CqlValue, since we would need to match variants that can never be created via cpp-rust-driver API (e.g. there is no way to create a `CqlValue::Counter()` value).

I think that we can merge it as is.

PR #155, opened 2024-08-05
ci: fix skipped tests

### LINKED ISSUE #144: CI: some tests are unnecessarily skipped when run against Scylla
A simple example are tests for `duration` cql type:
```
[SKIPPED] Unsupported for Apache Cassandra Version 3.0.8: Server version 3.10+ is required
[OK]      CassandraTypesDurationTests.Integration_Cassandra_MixedValues (1937 ms)
```

When scylla cluster is set up via ccm (no matter the version of scylla), the `ccm <node> version` command always returns a `3.0.8` version. This version is probably hardcoded in `nodetool` (`ccm <node> version` command delegates it to `nodetool`). However, there is a way to print a scylla version using ccm. There is a `ccm <node> versionfrombuild` command which seems to display the correct version of scylla node.

We need to adjust the logic responsible for filtering out the tests based on Scylla/Cassandra version. Testing framework should be aware whether it's used against Cassandra or Scylla cluster, and then execute either `ccm <node> version` or `ccm <node> versionfrombuild` respectively. Minimal versions required for specific tests should be adjusted for scylla, because I believe that there is no 1:1 mapping between Scylla's and Cassandra's versions and some features were probably introduced in different versions.

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/144

This PR addresses multiple issues:
## Skipped tests for scylla clusters
As linked issue mentions, `ccm <node> version` for scylla clusters always returns `3.0.8` version which I believe is hardcoded. Because of that, some tests were unnecessarily skipped. To retrieve scylla version, we should use `ccm <node> versionfrombuild`.

I adjusted the logic for retrieving the Scylla version. Apart from that, I skip the version checks for Scylla, since the feature to version mapping is not the same for Scylla and Cassandra. I left the checks for Cassandra. As of now, I don't think it's really important to check which features should be tested based on Scylla version, since we don't plan to test old Scylla versions. We will add such filters in the future, once they are needed.

## No compact test
Previously, this test was silently passing as it was skipped for both Scylla and Cassandra. However, the function `cass_cluster_set_no_compact` is not even implemented. I removed this test from CI (for both Scylla and Cassandra).

## Map collection iterator
cpp-rust-driver's implementation of `cass_iterator_from_collection` for maps was inconsistent with cpp-driver's implementation. This PR fixes it. I found this bug, because the test for a `map<int, duration>` type was handled a bit differently than for the other types (since apparently, `duration` cannot be used as a primary key). This test was failing due to the bug with map iterators.

## Collection objects in test framework
The `Collection` object was not initialized correctly when constructed from `CassValue`. An issue is described in detail in a corresponding commit.

## Modified tests for collections
Due to the two previous bugs, the tests for collections were modified a bit in the past. Since the issues are fixed, I reverted these changes and the tests now contain original logic from the upstream.

## Duration tests for cassandra
Enabled duration tests for cassandra.

## Scylla version in CI
Bumped Scylla version in CI to `6.0.0`.

## Pre-review checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,8 @@ jobs:
 :CompressionTests.*\
 :LoggingTests.*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
-:ExecutionProfileTest.InvalidName"
+:ExecutionProfileTest.InvalidName\
+:*NoCompactEnabledConnection"
       run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --scylla --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
 
     - name: Upload test logs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
       working-directory: ./scylla-rust-wrapper
       run: cargo test
 
-    - name: Run integration tests on Scylla 5.0.0
+    - name: Run integration tests on Scylla 6.0.0
       env:
 #        Ignored tests are added in the end, after the "-" sign.
         Tests: "ClusterTests.*\
@@ -53,7 +53,7 @@ jobs:
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :ExecutionProfileTest.InvalidName\
 :*NoCompactEnabledConnection"
-      run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --scylla --version=release:5.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
+      run: valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite ./cassandra-integration-tests --scylla --version=release:6.0.0 --category=CASSANDRA --verbose=ccm --gtest_filter="$Tests"
 
     - name: Upload test logs
       uses: actions/upload-artifact@v3
--- a/tests/src/integration/ccm/bridge.cpp
+++ b/tests/src/integration/ccm/bridge.cpp
@@ -849,7 +849,20 @@ CassVersion CCM::Bridge::get_cassandra_version() {
   // Get the version string from CCM
   std::vector<std::string> active_cluster_version_command;
   active_cluster_version_command.push_back(generate_node_name(1));
+
+  if (is_scylla_) {
+    // For scylla, we need to use `versionfrombuild`. `version` always returns 3.0.8.
+    active_cluster_version_command.push_back("versionfrombuild");
+
+    std::string ccm_output = execute_ccm_command(active_cluster_version_command);
+
+    // versionfrombuild returns a version directly
+    return CassVersion(ccm_output);
+  }
+
+  // Use `version` command for Cassandra.
   active_cluster_version_command.push_back("version");
+
   std::string ccm_output = execute_ccm_command(active_cluster_version_command);
 
   // Ensure the version release information exists and return the version
@@ -861,7 +874,7 @@ CassVersion CCM::Bridge::get_cassandra_version() {
 
   // Unable to determine version information from active cluster
   throw BridgeException("Unable to determine version information from active Cassandra cluster \"" +
-                        get_active_cluster() + "\"");
+                  get_active_cluster() + "\"");
 }
 
 DseVersion CCM::Bridge::get_dse_version() {
--- a/tests/src/integration/ccm/cass_version.hpp
+++ b/tests/src/integration/ccm/cass_version.hpp
@@ -281,7 +281,7 @@ private:
     std::string scylla_version_prefix = "release:";
     std::string version(version_string);
     if (version.compare(0, scylla_version_prefix.size(), scylla_version_prefix) == 0) {
-      version = "3.0.8";
+      version.replace(0, scylla_version_prefix.size(), "");
     }
     std::replace(version.begin(), version.end(), '.', ' ');
     std::size_t found = version.find("-");
--- a/tests/src/integration/integration.hpp
+++ b/tests/src/integration/integration.hpp
@@ -83,28 +83,28 @@
   SKIP_TEST("Unsupported for Apache Cassandra Version "          \
             << server_version_string << ": Server version " << version_string << "+ is required")
 
-#define CHECK_VERSION(version)                                                      \
+// Currently, we only check for the version if tests are being run against
+// Cassandra cluster. It's because, some of the tests were unnecessarily
+// skipped for Scylla. In the future (once it's needed) we might
+// do some version restrictions for Scylla clusters as well.
+#define SKIP_IF_CASSANDRA_VERSION_LT(version)                                                      \
   do {                                                                              \
     CCM::CassVersion cass_version = this->server_version_;                          \
     if (!Options::is_cassandra()) {                                                 \
       cass_version = static_cast<CCM::DseVersion>(cass_version).get_cass_version(); \
     }                                                                               \
-    if (cass_version < #version) {                                                  \
+    if (!Options::is_scylla() && cass_version < #version) {                                                  \
       SKIP_TEST_VERSION(cass_version.to_string(), #version)                         \
     }                                                                               \
   } while (0)
 
-#define CHECK_OPTIONS_VERSION(version)                                 \
-  if (Options::server_version() < #version) {                          \
-    SKIP_TEST_VERSION(Options::server_version().to_string(), #version) \
-  }
 
-#define CHECK_VALUE_TYPE_VERSION(type)                                            \
+#define CHECK_VALUE_TYPE_CASSANDRA_VERSION(type)                                            \
   CCM::CassVersion cass_version = this->server_version_;                          \
   if (!Options::is_cassandra()) {                                                 \
     cass_version = static_cast<CCM::DseVersion>(cass_version).get_cass_version(); \
   }                                                                               \
-  if (cass_version < type::supported_server_version()) {                          \
+  if (!Options::is_scylla() && cass_version < type::supported_server_version()) {                          \
     SKIP_TEST_VERSION(cass_version.to_string(), type::supported_server_version()) \
   }
 
--- a/tests/src/integration/objects/collection.hpp
+++ b/tests/src/integration/objects/collection.hpp
@@ -194,15 +194,17 @@ protected:
       } else {
         FAIL() << "Invalid CassValueType: Value type is not a valid collection";
       }
+  
+      // Check if collection is empty (null).
+      // Scylla does not distinguish between an empty collection and NULL value.
+      // See: https://opensource.docs.scylladb.com/stable/cql/types.html#sets
+      size_t collection_size = cass_value_item_count(value);
+      if (collection_size > 0) {
+        is_null_ = false;
+      }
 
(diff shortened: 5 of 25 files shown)


muzarski:
Rebased on master

tests/src/integration/ccm/bridge.cpp
```diff
@@ -849,19 +849,31 @@ CassVersion CCM::Bridge::get_cassandra_version() {
...
   std::vector<std::string> active_cluster_version_command;
   active_cluster_version_command.push_back(generate_node_name(1));
-  active_cluster_version_command.push_back("version");
+  if (!is_scylla_) {
+    active_cluster_version_command.push_back("version");
+  }
+  else {
+    // For scylla, we need to use versionfrombuild. version always returns 3.0.8.
+    active_cluster_version_command.push_back("versionfrombuild");
+  }
```
Lorak-mmk: Why not switch the condition around to avoid unnecessary negation?

tests/src/integration/tests/test_basics.cpp
```diff
@@ -343,7 +343,7 @@ CASSANDRA_INTEGRATION_TEST_F(BasicsTests, NoCompactEnabledConnection) {
     }
```
Lorak-mmk, line 343: Aren't we going to run into this ` if (!Options::is_cassandra()) {
    if (server_version_ >= "6.0.0") {` branch when we update Scylla to 6.0?
If so, can you fix this in this PR?

Lorak-mmk (changes requested):

tests/src/integration/tests/test_basics.cpp
muzarski: No, `Options::is_cassandra()` actually returns `true` for Scylla as well. It's because Scylla identifies as `Apache Cassandra` server (such info is retrieved from `ccm`).

There are 3 possible `CCM:ServerType` enum values:
```cpp

bool Options::is_cassandra() { return server_type_ == CCM::ServerType::CASSANDRA; }

bool Options::is_dse() { return server_type_ == CCM::ServerType::DSE; }

bool Options::is_ddac() { return server_type_ == CCM::ServerType::DDAC; }
```

I'm not sure about the difference between the types, but in Scylla's case, the server type is `CASSANDRA`. `Options::is_scylla_` is a flag responsible for distinguishing between scylla and cassandra.

Lorak-mmk: wth is ddac?! Never heard about it before.

muzarski:
v1.1: addressed review comment

Lorak-mmk (approved):

tests/src/integration/tests/test_basics.cpp
wprzytula: https://www.datastax.com/press-release/datastax-announces-datastax-distribution-apache-cassandratm

tests/src/integration/ccm/bridge.cpp
```diff
@@ -849,19 +849,31 @@ CassVersion CCM::Bridge::get_cassandra_version() {
...
+      size_t version_index = ccm_output.find("ReleaseVersion:");
+      if (version_index != std::string::npos) {
+        ccm_output.replace(0, version_index + 15, "");
+        return CassVersion(trim(ccm_output));
+      }
+
+        // Unable to determine version information from active cluster
+        throw BridgeException("Unable to determine version information from active Cassandra cluster \"" +
+                        get_active_cluster() + "\""); 
   }
```
wprzytula: Inconsistent indentation? `throw` should probably be indented less.

tests/src/integration/ccm/bridge.cpp
```diff
@@ -849,19 +849,31 @@ CassVersion CCM::Bridge::get_cassandra_version() {
...
 
-  // Unable to determine version information from active cluster
-  throw BridgeException("Unable to determine version information from active Cassandra cluster \"" +
-                        get_active_cluster() + "\"");
+  // versionfrombuild returns a version directly
+  return CassVersion(ccm_output);
```
wprzytula: For readability, please make the `if` and `if-else` statement, with this in `if (is_scylla)` branch, and the other code in `else` branch.

tests/src/integration/ccm/cass_version.hpp
```diff
@@ -281,7 +281,7 @@ class CassVersion {
     std::string scylla_version_prefix = "release:";
     std::string version(version_string);
     if (version.compare(0, scylla_version_prefix.size(), scylla_version_prefix) == 0) {
-      version = "3.0.8";
+      version.replace(0, scylla_version_prefix.size(), "");
     }
     std::replace(version.begin(), version.end(), '.', ' ');
```
wprzytula, line 286: How this new code actually retrieves the Scylla version? Or, if it's already retrieved, what variable stores it?

tests/src/integration/integration.hpp
```diff
@@ -83,28 +83,28 @@
...
 
-#define CHECK_VERSION(version)                                                      \
+#define CHECK_CASSANDRA_VERSION(version)                                                      \
   do {                                                                              \
     CCM::CassVersion cass_version = this->server_version_;                          \
     if (!Options::is_cassandra()) {                                                 \
       cass_version = static_cast<CCM::DseVersion>(cass_version).get_cass_version(); \
     }                                                                               \
-    if (cass_version < #version) {                                                  \
+    if (!Options::is_scylla() && cass_version < #version) {                                                  \
```
wprzytula: Please add a comment mentioning why only Cassandra version is checked now and maybe a `TODO: Scylla version checks (when needed)`.

.github/workflows/build.yml
```diff
@@ -51,7 +51,8 @@ jobs:
 :CompressionTests.*\
 :LoggingTests.*\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
-:ExecutionProfileTest.InvalidName"
+:ExecutionProfileTest.InvalidName\
+:*NoCompactEnabledConnection"
```
wprzytula, line 55: IIUC, this commit makes the CI pass again, after the previous commit makes if fail due to disabling version filtering for Scylla.
We shouldn't have any commits that do not pass tests, because those break bisectability. A quick fix is to swap this commit with the previous one.

tests/src/integration/objects/collection.hpp
```diff
@@ -198,11 +198,8 @@ class Collection : public Object<CassCollection, cass_collection_free> {
       // Initialize the iterator
       iterator_ = cass_iterator_from_collection(value);
 
-      // Determine if the collection is empty (null)
-      const CassValue* check_value = cass_iterator_get_value(iterator_.get());
-      if (check_value) {
-        is_null_ = false;
-      }
+      // We already checked that collection is not null.
+      is_null_ = false;
```
wprzytula: While this is an unpleasant peculiarity, empty collections seem to be identical to NULL values for the DB. See https://github.com/scylladb/scylla-rust-driver/pull/1007 for reference.

wprzytula: Have you considered making `position` field a non-`Option`?

wprzytula: I don't want to dive too deep into the code (because I have other tasks to do), but I think we should do our best to make the cpp-rust-driver's behaviour consistent with what the official cpp-driver's tests expect.

wprzytula (changes requested):

tests/src/integration/objects/collection.hpp
Lorak-mmk: > I don't want to dive too deep into the code (because I have other tasks to do), but I think we should do our best to make the cpp-rust-driver's behaviour consistent with what the official cpp-driver's tests expect.

It was my understanding that this is what this PR does. Is that not the case?

tests/src/integration/ccm/cass_version.hpp
muzarski: > Or, if it's already retrieved, what variable stores it?

The `version` variable.

tests/src/integration/objects/collection.hpp
muzarski: > While this is an unpleasant peculiarity, empty collections seem to be identical to NULL values for the DB. See [scylladb/scylla-rust-driver#1007](https://github.com/scylladb/scylla-rust-driver/pull/1007) for reference.

Good catch. I'll add the `is_null_` initialization here. However, I still think that iterator should not be used for checking if collection is empty. The iterator, when created, is UNINITIALIZED - to retrieve the first value, we should firstly advance the iterator via `cass_iterator_next`. I don't want to advance the iterator just to check if it's empty. I'll adjust the code so we make use of `cass_value_item_count` function, which can be used to see if collection is empty.

muzarski:
v2: addressed review comments

tests/src/integration/ccm/bridge.cpp
```diff
@@ -849,19 +849,32 @@ CassVersion CCM::Bridge::get_cassandra_version() {
...
+  if (is_scylla_) {
+    // For scylla, we need to use versionfrombuild. version always returns 3.0.8.
+    active_cluster_version_command.push_back("versionfrombuild");
+
+    std::string ccm_output = execute_ccm_command(active_cluster_version_command);
+
+    // versionfrombuild returns a version directly
+    return CassVersion(ccm_output);
   }
+  else {
```
dkropachev: can you please drop this `else`

tests/src/integration/integration.hpp
```diff
@@ -83,28 +83,32 @@
...
       SKIP_TEST_VERSION(cass_version.to_string(), #version)                         \
     }                                                                               \
   } while (0)
 
-#define CHECK_OPTIONS_VERSION(version)                                 \
-  if (Options::server_version() < #version) {                          \
+#define CHECK_OPTIONS_CASSANDRA_VERSION(version)                                 \
```
dkropachev: I don't see it is ever used, can we drop it ?

tests/src/integration/integration.hpp
```diff
@@ -83,28 +83,32 @@
   SKIP_TEST("Unsupported for Apache Cassandra Version "          \
             << server_version_string << ": Server version " << version_string << "+ is required")
 
-#define CHECK_VERSION(version)                                                      \
+// Currently, we only check for the version if tests are being run against
+// Cassandra cluster. It's because, some of the tests were unnecessarily
+// skipped for Scylla. In the future (once it's needed) we might
+// do some version restrictions for Scylla clusters as well.
+#define CHECK_CASSANDRA_VERSION(version)                                                      \
```
dkropachev: Since we are renaming it, can we rename it to somehting more telling, like `SKIP_IF_CASSANDRA_VERSION_LT`

tests/src/integration/objects/collection.hpp
```diff
@@ -194,15 +194,17 @@ class Collection : public Object<CassCollection, cass_collection_free> {
...
+      // Check if collection is empty (null).
+      // Scylla does not distinguish between an empty collection and NULL value.
+      // See: https://opensource.docs.scylladb.com/stable/cql/types.html#sets
+      size_t collection_size = cass_value_item_count(value);
+      if (collection_size > 0) {
+        is_null_ = false;
+      }
 
       // Initialize the iterator
       iterator_ = cass_iterator_from_collection(value);
```
dkropachev, line 207: Can you drop this line then ?

dkropachev (changes requested):

tests/src/integration/objects/collection.hpp
muzarski: No. `iterator_` is a field. This code is called when constructor is called, and so this line initializes a field of a new object.

muzarski:
v2.1: Addressed @dkropachev 's review comments.

PR #159, opened 2024-09-05
ci: redo cicd

dkropachev:
## Goal of the PR

1. Speed up CICD
2. Make every step to be  easily reproducible locally
3. Separate build files from the code 
4. Run integration tests on different versions of scylla and cassandra.

## Pre-review checklist


dkropachev:
CICD Speedup results:
`Build + Scylla IT`: takes `5` minutes instead of `18`.
`Build + Cassandra IT`: takes `6` minutes instead of `18`.

muzarski:
I see two main issues here:
1. The test binary cache is not working as intended - the binary is not found in the cache in `Scylla ITs` job, and then, it is built again.
2. Something is messed up with the test filter. Notice that we only ran 9 test cases - This is probably the reason why CI sped up so much. When it comes to the filter, I see some unnecessary whitespaces that probably mess things up:
`Note: Google Test filter = ClusterTests.* :BasicsTests.* :PreparedTests.* :CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_* :BatchSingleNodeClusterTests*:BatchCounterSingleNodeClusterTests*:BatchCounterThreeNodeClusterTests* :ErrorTests.* :SslNoClusterTests*:SslNoSslOnClusterTests* :SchemaMetadataTest.*KeyspaceMetadata:SchemaMetadataTest.*MetadataIterator:SchemaMetadataTest.*View* :TracingTests.* :ByNameTests.* :CompressionTests.* :LoggingTests.* :PreparedMetadataTests.* :UseKeyspaceCaseSensitiveTests.* :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare :ExecutionProfileTest.InvalidName :*NoCompactEnabledConnection :PreparedMetadataTests.Integration_Cassandra_AlterDoesntUpdateColumnCount :UseKeyspaceCaseSensitiveTests.Integration_Cassandra_ConnectWithKeyspace:*_DSE_*`

dkropachev:
> I see two main issues here:
> 
> 1. The test binary cache is not working as intended - the binary is not found in the cache in `Scylla ITs` job, and then, it is built again.
> 2. Something is messed up with the test filter. Notice that we only ran 9 test cases - This is probably the reason why CI sped up so much. When it comes to the filter, I see some unnecessary whitespaces that probably mess things up:
>    `Note: Google Test filter = ClusterTests.* :BasicsTests.* :PreparedTests.* :CassandraTypes/CassandraTypesTests/*.Integration_Cassandra_* :BatchSingleNodeClusterTests*:BatchCounterSingleNodeClusterTests*:BatchCounterThreeNodeClusterTests* :ErrorTests.* :SslNoClusterTests*:SslNoSslOnClusterTests* :SchemaMetadataTest.*KeyspaceMetadata:SchemaMetadataTest.*MetadataIterator:SchemaMetadataTest.*View* :TracingTests.* :ByNameTests.* :CompressionTests.* :LoggingTests.* :PreparedMetadataTests.* :UseKeyspaceCaseSensitiveTests.* :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare :ExecutionProfileTest.InvalidName :*NoCompactEnabledConnection :PreparedMetadataTests.Integration_Cassandra_AlterDoesntUpdateColumnCount :UseKeyspaceCaseSensitiveTests.Integration_Cassandra_ConnectWithKeyspace:*_DSE_*`

Thanks, now it looks good, please take a look.

dkropachev:
CICD Speedup results:
Build + Scylla IT: takes `15` minutes instead of `19`.
Build + Cassandra IT: takes `15` minutes instead of `19`.

.github/workflows/build-lint-and-test.yml
```diff
@@ -0,0 +1,227 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: make install-build-dependencies
+
+      - name: Build integration test binary
+        run: make build-integration-test-bin
+
+      - name: Check
```
muzarski, line 32: I would decouple this step into 3 separate steps. Let's say, only a clippy check fails - then the logs from other checks introduce unnecessary noise, making it harder to see which check actually failed.

.github/workflows/build-lint-and-test.yml
```diff
@@ -0,0 +1,227 @@
...
+    build/libscylla-cpp-driver.*
+  INTEGRATION_TEST_BIN_CACHE_KEY: integration-test-bin-${{ github.sha }}
+  # Goes to `Makefile` to let it pickup cached binary
+  DONT_REBUILD_INTEGRATION_BIN: true
+
+jobs:
+  build-lint-and-unit-test:
+    name: Build, lint and run unit tests
+    runs-on: ubuntu-latest
+    steps:
```
muzarski, line 25: I would reorder the steps in this job. It does not make much sense to build an integration test binary, if checks or unit tests for rust wrapper fail. The order I'd recommend:
- checkout
- install dependencies
- rust wrapper checks (3 separate steps - format, cargo, clippy)
- unit and proxy tests
- build IT binary
- save IT binary

.github/workflows/build-lint-and-test.yml
```diff
@@ -0,0 +1,227 @@
...
+        run: make build-integration-test-bin
+
+      - name: Check
+        run: make check
+
+      - name: Run unit and proxy tests
+        run: make run-test-unit
+
+      - name: Save integration test binary
+        if: success()
```
muzarski: Nit: AFAIU, `success()` is used by default. See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions.

.github/workflows/build-lint-and-test.yml
```diff
@@ -0,0 +1,227 @@
...
+        with:
+          name: test-logs-scylla-${{ matrix.scylla-version }}
+          path: ./log/*
+
+      - name: Upload CCM logs
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: ccm-log-scylla-${{ matrix.scylla-version }}
+          path: /tmp/ccm*/ccm*/node*/logs/*
```
muzarski: Question: Where does this path come from? Is it some default path that ccm uses to export logs? The question also applies to the `./log/*` path used in the previous step.

muzarski: Maybe add some constant variables for these paths? They are reused for cassandra integration tests as well.

Makefile
```diff
@@ -0,0 +1,183 @@
+EMPTY :=
+SPACE := ${EMPTY} ${EMPTY}
+
+ifndef SCYLLA_TEST_FILTER
+SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
```
muzarski, line 5: I was looking for alternatives to getting rid of these annoying whitespaces. The only thing I found is last paragraph in https://www.gnu.org/software/make/manual/html_node/Splitting-Lines.html. Basically, replace `\` with `$\` at the end of each line. This is still kinda annoying and tedious to do, but then we don't need to introduce `EMPTY` and `SPACE` definitions. I'm not sure which approach is better. WDYT?

Makefile
```diff
@@ -0,0 +1,183 @@
...
+run-test-integration-scylla: prepare-integration-test build-integration-test-bin-if-missing download-ccm-scylla-image
+	@echo "Running integration tests on scylla ${SCYLLA_VERSION}"
+	valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${SCYLLA_TEST_FILTER}"
+
+run-test-integration-cassandra: prepare-integration-test download-ccm-cassandra-image install-java8-if-missing
+ifndef DONT_REBUILD_INTEGRATION_BIN
+run-test-integration-cassandra: build-integration-test-bin
+else
+run-test-integration-cassandra: build-integration-test-bin-if-missing
+endif
```
muzarski: Why do we have this `ifndef` clause for `run-test-integration-cassandra`, but not for `run-integration-scylla`? I believe that the purpose of `DONT_REBUILD_INTEGRATION_BIN` is to distinguish the use case between a CI and a local development - during CI, we don't want to rebuild a binary, while as a developer, I usually want to rebuild it after introducing some changes. I think that we should do the same for scylla ITs.

muzarski: Small nitpick: I'd change `ifndef` to `ifdef` (and swap the branches) just to improve readability. Double negation is always tricky :)

.github/workflows/build-lint-and-test.yml
```diff
@@ -0,0 +1,227 @@
...
+
+      - name: Install CCM
+        run: |
+          make install-ccm-if-missing
+
+      - name: Get scylla version
+        id: scylla-version
+        run: |
+          if [[ "${{ matrix.scylla-version }}" == "ENTERPRISE-RELEASE" ]]; then
+            echo "value=$(python3 ci/version_fetch.py --version-index 1 scylla-enterprise-stable:1 | tr -d '\"')" >> $GITHUB_OUTPUT
```
muzarski: Maybe `>> $GITHUB_OUTPUT` -> `| tee -a $GITHUB_OUTPUT` so we can quickly check the chosen version in CI logs?

ci/version_fetch.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/python3
```
muzarski, line 1: From what I've checked, we use a very similar script in java-driver's CI. Out of curiosity - are any other drivers/repos making use of similar script?

muzarski (changes requested):
Mostly some nits. Other than that, really nice job!

.github/workflows/build-lint-and-test.yml
dkropachev: Then there is a problem that you have to keep `Makefile` and `build-lint-and-test.yml` in sync.

.github/workflows/build-lint-and-test.yml
dkropachev: Done, but I realy don't want to break `make check` because of https://github.com/scylladb/cpp-rust-driver/pull/159#discussion_r1751762641

.github/workflows/build-lint-and-test.yml
dkropachev: removed

Makefile
dkropachev: I don't see big difference between these two.

Makefile
dkropachev: Yeah, it does not work, swapped it to `ifeq`

.github/workflows/build-lint-and-test.yml
dkropachev: done

ci/version_fetch.py
dkropachev: We are going to have in all the drivers, later on i will make a github action for it.

Makefile
muzarski: Let's leave it as is then

.github/workflows/build-lint-and-test.yml
dkropachev: done

muzarski (approved):

PR #165, opened 2024-09-10
misc: Display consistency

### LINKED ISSUE #149: misc: display consistency
Implement `cass_consistency_string` which stringifies a consistency value.

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/149

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -292,7 +292,8 @@ mod tests {
         let wrapped_cass_fut = PtrWrapper(cass_fut);
         unsafe {
             let handle = thread::spawn(move || {
-                let PtrWrapper(cass_fut) = wrapped_cass_fut;
+                let wrapper = wrapped_cass_fut;
+                let cass_fut = wrapper.0;
                 let mut message: *const c_char = std::ptr::null();
                 let mut msg_len: size_t = 0;
                 cass_future_error_message(cass_fut, &mut message, &mut msg_len);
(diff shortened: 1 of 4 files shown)


dkropachev (approved):

scylla-rust-wrapper/src/future.rs
```diff
@@ -292,7 +292,8 @@ mod tests {
         let wrapped_cass_fut = PtrWrapper(cass_fut);
         unsafe {
             let handle = thread::spawn(move || {
-                let PtrWrapper(cass_fut) = wrapped_cass_fut;
+                let wrapper = wrapped_cass_fut;
+                let cass_fut = wrapper.0;
```
wprzytula, line 296: Nit: the previous syntax made it clear what is the wrapper we are unwrapping, now it's not that explicit. Let's prefer deconstruction syntax (`let Wrapper(inner) = wrapper;`) in the future.

muzarski: This won't work in Rust edition 2021. Refer to: https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html.

```
let Wrapper(inner) = wrapper;
```

In this case, the closure will only capture `inner`. And this is not `Send` in our case (the wrapper is).

muzarski: Ah nvm, I see what you meant.

muzarski: To deconstruct it, after capturing whole wrapper first. Yes, I agree then.

PR #170, opened 2024-09-13
exec_profile: token aware routing with shuffling setting

muzarski:
Implemented:
- `cass_cluster_set_token_aware_routing_shuffle_replicas`
- `cass_execution_profile_set_token_aware_routing_shuffle_replicas`

Notice, that this setting (replica shuffling) is only applicable, if token aware routing is enabled (which is, by default).

This setting is `true` by default.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -40,6 +40,7 @@ const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 #[derive(Clone, Debug)]
 pub(crate) struct LoadBalancingConfig {
     pub(crate) token_awareness_enabled: bool,
+    pub(crate) token_aware_shuffling_replicas_enabled: bool,
     pub(crate) dc_awareness: Option<DcAwareness>,
     pub(crate) latency_awareness_enabled: bool,
     pub(crate) latency_awareness_builder: LatencyAwarenessBuilder,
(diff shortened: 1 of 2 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -49,6 +50,11 @@ impl LoadBalancingConfig {
     // as it results in panic due to DefaultPolicyBuilder::build() spawning a tokio task.
     pub(crate) async fn build(self) -> Arc<dyn LoadBalancingPolicy> {
         let mut builder = DefaultPolicyBuilder::new().token_aware(self.token_awareness_enabled);
+        if self.token_awareness_enabled {
```
dkropachev, line 53: I would recommend to add another option to enable shuffling no matter what:
```suggestion
        if self.shuffling_replicas_enabled || (self.token_awareness_enabled && self.token_aware_shuffling_replicas_enabled) {
```

dkropachev (changes requested):

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -49,6 +50,11 @@ impl LoadBalancingConfig {
...
+            // Cpp-driver enables shuffling replicas only if token aware routing is enabled.
+            builder =
+                builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
+        }
```
dkropachev, line 57: I would recommend add option to enable shuffling no matter what:
```suggestion
        shuffling := self.shuffling_replicas_enabled || (self.token_awareness_enabled && self.token_aware_shuffling_replicas_enabled) 
        if self.shuffling {
            // Cpp-driver enables shuffling replicas only if token aware routing is enabled.
            builder =
                builder.enable_shuffling_replicas(shuffling);
        }
```

dkropachev (changes requested):

scylla-rust-wrapper/src/cluster.rs
muzarski: From what I've just checked, this setting in rust-driver applies only for replicas (when token-awareness is enabled) as well. When token-awareness is disabled, we always pick nodes randomly (respecting the preferred dc, rack settings etc.). I think this setting is rather introduced, to disable shuffling for replicas, if user wishes to (since it's enabled by default). This is obviously not recommended, but some users might want to pick replicas deterministically for some reason.

dkropachev (approved):

PR #174, opened 2024-09-18
makefile: fix rebuild IT binary condition

muzarski:
Currently, we rebuild (even if it exists) IT binary if DONT_REBUILD_INTEGRATION_BIN env var is non-empty. I believe it should be the opposite.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,10 +60,6 @@ ifndef CASSANDRA_VERSION
 	CASSANDRA_VERSION := 3.11.17
 endif
 
-ifndef DONT_REBUILD_INTEGRATION_BIN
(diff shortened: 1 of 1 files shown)


dkropachev (approved):
Please remove:
```
ifndef DONT_REBUILD_INTEGRATION_BIN
	DONT_REBUILD_INTEGRATION_BIN := ${EMPTY}
endif
```

And use:
```
ifdef DONT_REBUILD_INTEGRATION_BIN
run-test-integration-scylla: build-integration-test-bin-if-missing
else
run-test-integration-scylla: build-integration-test-bin
endif
```

muzarski:
CI passed, I'm merging

PR #176, opened 2024-09-19
Makefile: add clang-format

dkropachev:
It adds `clang-format` to the list of linters for c/cpp code.
Also it applies format changes suggested by `clang-format`

## Pre-review checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,28 @@
+---
+BasedOnStyle: LLVM
+Language: Cpp
+Standard: Cpp03
+AccessModifierOffset: -2
+IndentWidth: 2
+TabWidth: 8
+ColumnLimit: 100
+UseTab: Never
+IndentCaseLabels: true
+AlignAfterOpenBracket: true
+AlignEscapedNewlines: Left
+BreakConstructorInitializers: BeforeComma
+AllowShortBlocksOnASingleLine: false
+DerivePointerAlignment: false
+PointerAlignment: Left
+BinPackParameters: true
+BinPackArguments: true
+AllowShortIfStatementsOnASingleLine: true
+CompactNamespaces: true
+AlignOperands: true
+SpacesInContainerLiterals: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+Cpp11BracedListStyle: false
+AlwaysBreakTemplateDeclarations: true
+BreakBeforeInheritanceComma: true
+...
+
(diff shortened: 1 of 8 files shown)


Lorak-mmk:
I wonder if this this something that should be done in this project. C/Cpp files are only test-related, copied verbatim from original cpp-driver and modified a bit. Maybe it's of more value to reduce amount of modifications to them, so that it is easier to diff them eith their original versions?

Lorak-mmk:
Another issue: biggest cause for automated code formatting is consistency of formatting. This PR uses default style of clang-format, without any config. This is consistent neither with scylla / seastar style, nor with original cpp-driver style.
Imo if we were to add automated formatting it should mimic one of those style as close as possible. Personally I'd be for the style of original cpp-driver, becuse it will make it easier to read both codebases together (which is something I know I had to do many times when developing this project).

muzarski:
There is a clang-format config file in cpp-driver's repo: https://github.com/datastax/cpp-driver/blob/master/.clang-format. We should definitely use it,

dkropachev:
> There is a clang-format config file in cpp-driver's repo: https://github.com/datastax/cpp-driver/blob/master/.clang-format. We should definitely use it,

Thanks, I have copied it.
@Lorak-mmk , does it address your concerns ?

Lorak-mmk:
How did you generate the last commit? It looks like it doesn't respect the config.  Example: look at `src/aligned_storage.hpp`. Comment was reflown in your PR, despite not being longer than 100 characters (which is the limit in config

I checked out the PR, dropped the last commit and executed `make fix-clang-format` - this seems to result in correct changes (I don't see wrongly reflown comments).

Another issue: I think that `third_party` folder should be skipped. Those are basically dependencies, not something that we should edit.

dkropachev:
> How did you generate the last commit? It looks like it doesn't respect the config. Example: look at `src/aligned_storage.hpp`. Comment was reflown in your PR, despite not being longer than 100 characters (which is the limit in config
> 
> I checked out the PR, dropped the last commit and executed `make fix-clang-format` - this seems to result in correct changes (I don't see wrongly reflown comments).
> 
> Another issue: I think that `third_party` folder should be skipped. Those are basically dependencies, not something that we should edit.

Thanks, both done, `third_party` excluded, last commit removed and formatting reapplied.

muzarski (approved):
Please squash first and second commits, otherwise LGTM

dkropachev:
> Please squash first and second commits, otherwise LGTM

@Lorak-mmk , are you ok with it ?

Lorak-mmk (approved):

wprzytula (approved):

PR #180, opened 2024-09-26
integration: ccm decommission fix

muzarski:
## ccm <node> decommission
Scylla's fork of ccm does not recognize `--force` flag
for `decommission` command. `force_decommission_node` function
appends such flag to the ccm command.

Replaced occurrences of `force_decommission_node` with
`decommission_node` throughout the tests.

## Test filter
Since we fixed the usage of ccm in integration tests regarding
the decommission, we can enable `ConsistencyThreeNodeCluster` test suite.

Unfortunately, other test suites (`SessionTests` and `ControlConnectionTests`)
affected by this change still need some driver's features to be implemented.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ ifndef SCYLLA_TEST_FILTER
 SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :BasicsTests.*\
 :ConfigTests.*\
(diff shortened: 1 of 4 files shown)


dkropachev:
to add more context, `nodetool decomission` on scylla does not have flag `--force` so no way `scylla-ccm` can get this API near soon.

PR #182, opened 2024-09-30
async: fix AsyncTests.Integration_Cassandra_Close

muzarski:
## Pre-review checklist


dkropachev:
Opened an issue to fix this problem: https://github.com/scylladb/cpp-rust-driver/issues/304

PR #185, opened 2024-10-07
prepared: CassPrepared api

### LINKED ISSUE #54: CassPrepared missing functions
Function list:

- [ ] `cass_prepared_parameter_name`
- [ ] `cass_prepared_parameter_data_type`
- [ ] `cass_prepared_parameter_data_type_by_name`
- [ ] `cass_prepared_parameter_data_type_by_name_n`

muzarski:
fix: https://github.com/scylladb/cpp-rust-driver/issues/54

This PR implements following API functions:
- `cass_prepared_parameter_name`
- `cass_prepared_parameter_data_type`
- `cass_prepared_parameter_data_type_by_name`
- `cass_prepared_parameter_data_type_by_name_n`

## Pre-review checklist

- ~[ ] I have implemented Rust unit tests for the features/changes introduced.~ -> unable to implement unit tests, since there is no way to mock `PreparedStatement` struct from rust-driver

### DIFF
--- a/scylla-rust-wrapper/src/prepared.rs
+++ b/scylla-rust-wrapper/src/prepared.rs
@@ -80,3 +80,16 @@ pub unsafe extern "C" fn cass_prepared_parameter_name(
         None => CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
     }
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_prepared_parameter_data_type(
+    prepared_raw: *const CassPrepared,
+    index: size_t,
+) -> *const CassDataType {
+    let prepared = ptr_to_ref(prepared_raw);
+
+    match prepared.variable_col_data_types.get(index as usize) {
+        Some(dt) => Arc::as_ptr(dt),
+        None => std::ptr::null(),
+    }
+}
--- a/scylla-rust-wrapper/src/prepared.rs
+++ b/scylla-rust-wrapper/src/prepared.rs
@@ -1,10 +1,12 @@
 use scylla::{frame::value::MaybeUnset::Unset, transport::PagingState};
-use std::sync::Arc;
+use std::{os::raw::c_char, sync::Arc};
 
 use crate::{
     argconv::*,
+    cass_error::CassError,
     cass_types::{get_column_type, CassDataType},
     statement::{CassStatement, Statement},
+    types::size_t,
 };
 use scylla::prepared_statement::PreparedStatement;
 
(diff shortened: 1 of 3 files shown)


wprzytula:
> unable to implement unit tests, since there is no way to mock PreparedStatement struct from rust-driver

This seems to be a use case for a `cpp-rust-driver-unstable` feature in Rust driver. Such feature could e.g. expose `pub` constructors of important entities such as `PreparedStatement`.

Lorak-mmk:
> > unable to implement unit tests, since there is no way to mock PreparedStatement struct from rust-driver
> 
> This seems to be a use case for a `cpp-rust-driver-unstable` feature in Rust driver. Such feature could e.g. expose `pub` constructors of important entities such as `PreparedStatement`.

If there is a use case for that, then maybe this should be a public API? I know there was at least 1 Rust Driver user asking for this before.

Lorak-mmk (approved):

scylla-rust-wrapper/src/prepared.rs
```diff
@@ -30,6 +30,26 @@ impl CassPrepared {
...
+        match self.variable_col_data_types.get(index) {
+            Some(dt) => Some(dt),
+            // This is a violation of driver's internal invariant.
+            // Since `self.variable_col_data_types` is created based on prepared statement's
+            // col specs, and we found an index with a corresponding name, we should
+            // find a CassDataType at given index.
+            None => panic!(
+                "Cannot find a data type of parameter with given name: {}. This is a driver bug!",
+                name
+            ),
```
wprzytula, line 50: This makes me wonder if it's possible to somehow zip `variable_col_data_types` with `statement.variable_col_specs`. E.g. make them stored in a common vec of pairs. If it were possible, we would get much stronger type system guarantees.

scylla-rust-wrapper/src/prepared.rs
```diff
@@ -93,3 +113,27 @@ pub unsafe extern "C" fn cass_prepared_parameter_data_type(
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn cass_prepared_parameter_data_type_by_name_n(
+    prepared_raw: *const CassPrepared,
+    name: *const c_char,
+    name_length: size_t,
+) -> *const CassDataType {
+    let prepared = ptr_to_ref(prepared_raw);
+    let parameter_name = ptr_to_cstr_n(name, name_length).expect("Parameter name is not UTF-8");
```
wprzytula: maybe: `Parameter name` -> `Prepared parameter name`

wprzytula (approved):

scylla-rust-wrapper/src/prepared.rs
muzarski: I don't see how we could do it without cloning the `statement.variable_col_specs`.

PR #187, opened 2024-10-09
binding: fix by-name parameter binding and enable corresponding IT

muzarski:
## Motivation
Before this PR, binding values by name to unprepared queries was buggy:
- Main issue was that we were not even reading name-to-index mapping (`SimpleQuery::name_to_bound_index`) - only writing to it. In result, we were not really binding by name, but rather by position, in the same order as `cass_statement_bind_*` calls appeared. The `bound_values` vector was directly passed to `Session::query_*` calls.
- `CassStatement::bind_cql_value` would return `CASS_OK` when size of `SimpleQuery::name_to_bound_index` map exceeded the declared size of parameters (`CassStatement::bound_values.len()`)
- We wouldn't unquote the parameter names for unprepared queries in `CassStatement::bind_cql_value`

This PR solves this issue, and enables another integration test suite - `NamedParametersTests`.

## Changes
Apart from the fixes to `CassStatement::bind_cql_value`, I also introduced a custom implementation of `SerializeRow` for unprepared queries execution. I created a `SimpleQueryRowSerializer` struct which holds necessary information to perform serialization. Now, with following context:
- `bound_values` vector - directly taken from `CassStatement::bound_values`
- an optional name-to-index mapping (from `SimpleQuery::name_to_bound_index`) - if it's non-empty, it means that we bind by names, otherwise we preserve the original order of `bound_values`.
- column specs from `RowSerializationContext`

we are able to serialize values in correct order for unprepared queries.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -3,6 +3,7 @@ use scylla::transport::errors::*;
 // Re-export error types.
 pub(crate) use crate::cass_error_types::{CassError, CassErrorSource};
 use crate::query_error::CassErrorResult;
+use crate::statement::UnknownNamedParameterError;
 
 pub trait ToCassError {
     fn to_cass_error(&self) -> CassError;
@@ -94,7 +95,14 @@ impl ToCassError for BadQuery {
             BadQuery::ValuesTooLongForKey(_usize, _usize2) => CassError::CASS_ERROR_LAST_ENTRY,
             BadQuery::BadKeyspaceName(_bad_keyspace_name) => CassError::CASS_ERROR_LAST_ENTRY,
             BadQuery::Other(_other_query) => CassError::CASS_ERROR_LAST_ENTRY,
-            BadQuery::SerializationError(_) => CassError::CASS_ERROR_LAST_ENTRY,
+            BadQuery::SerializationError(e) => {
+                if e.downcast_ref::<UnknownNamedParameterError>().is_some() {
+                    // It means that our custom `UnknownNamedParameterError` was returned.
+                    CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
+                } else {
+                    CassError::CASS_ERROR_LAST_ENTRY
+                }
+            }
             BadQuery::TooManyQueriesInBatchStatement(_) => CassError::CASS_ERROR_LAST_ENTRY,
             // BadQuery is non_exhaustive
             // For now, since all other variants return LAST_ENTRY,
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -124,7 +124,7 @@ impl BoundSimpleQuery {
         let index = self.name_to_bound_index.get(name);
 
         if let Some(idx) = index {
-            return self.bind_cql_value(*idx, value);
+            self.bind_cql_value(*idx, value)
         } else {
             // Find first free index, i.e. index where value at this index is Unset.
             let free_index = self.bound_values.iter().position(|v| matches!(v, Unset));
@@ -132,11 +132,12 @@ impl BoundSimpleQuery {
             if let Some(index) = free_index {
                 self.name_to_bound_index.insert(name.to_string(), index);
 
-                return self.bind_cql_value(index, value);
+                self.bind_cql_value(index, value)
+            } else {
+                // No free index for a given name. Cpp-driver returns this error in such case.
+                CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
             }
         }
-
-        CassError::CASS_OK
     }
 }
 
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -57,14 +57,12 @@ impl BoundPreparedStatement {
         }
     }
 
-    fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
-        // If the name was quoted, then we should treat it as case sensitive.
-        let (name_unquoted, is_case_sensitive) =
-            match name.strip_prefix('\"').and_then(|s| s.strip_suffix('\"')) {
-                Some(name_unquoted) => (name_unquoted, true),
-                None => (name, false),
-            };
-
+    fn bind_cql_value_by_name(
+        &mut self,
+        name: &str,
+        is_case_sensitive: bool,
+        value: Option<CassCqlValue>,
+    ) -> CassError {
         let indices: Vec<usize> = self
             .statement
             .statement
@@ -72,8 +70,8 @@ impl BoundPreparedStatement {
             .iter()
             .enumerate()
             .filter(|(_, col)| {
-                is_case_sensitive && col.name() == name_unquoted
-                    || !is_case_sensitive && col.name().eq_ignore_ascii_case(name_unquoted)
+                is_case_sensitive && col.name() == name
+                    || !is_case_sensitive && col.name().eq_ignore_ascii_case(name)
             })
             .map(|(i, _)| i)
             .collect();
@@ -166,9 +164,18 @@ impl CassStatement {
     }
 
     fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
+        // If the name was quoted, then we should treat it as case sensitive.
+        let (name_unquoted, is_case_sensitive) =
+            match name.strip_prefix('\"').and_then(|s| s.strip_suffix('\"')) {
+                Some(name_unquoted) => (name_unquoted, true),
+                None => (name, false),
+            };
+
         match &mut self.statement {
-            BoundStatement::Simple(simple) => simple.bind_cql_value_by_name(name, value),
-            BoundStatement::Prepared(prepared) => prepared.bind_cql_value_by_name(name, value),
+            BoundStatement::Simple(simple) => simple.bind_cql_value_by_name(name_unquoted, value),
+            BoundStatement::Prepared(prepared) => {
+                prepared.bind_cql_value_by_name(name_unquoted, is_case_sensitive, value)
+            }
         }
     }
 
--- a/scylla-rust-wrapper/src/statement.rs
+++ b/scylla-rust-wrapper/src/statement.rs
@@ -20,61 +20,71 @@ use std::slice;
 use std::sync::Arc;
 
 #[derive(Clone)]
-pub enum Statement {
-    Simple(SimpleQuery),
-    // Arc is needed, because PreparedStatement is passed by reference to session.execute
-    Prepared(Arc<CassPrepared>),
+pub enum BoundStatement {
+    Simple(BoundSimpleQuery),
+    Prepared(BoundPreparedStatement),
 }
 
 #[derive(Clone)]
-pub struct SimpleQuery {
-    pub query: Query,
-    pub name_to_bound_index: HashMap<String, usize>,
-}
-
-pub struct CassStatement {
-    pub statement: Statement,
+pub struct BoundPreparedStatement {
+    // Arc is needed, because PreparedStatement is passed by reference to session.execute
+    pub statement: Arc<CassPrepared>,
     pub bound_values: Vec<MaybeUnset<Option<CassCqlValue>>>,
-    pub paging_state: PagingState,
-    pub paging_enabled: bool,
-    pub request_timeout_ms: Option<cass_uint64_t>,
-
-    pub(crate) exec_profile: Option<PerStatementExecProfile>,
 }
 
-impl BoxFFI for CassStatement {}
-
-impl CassStatement {
+impl BoundPreparedStatement {
     fn bind_cql_value(&mut self, index: usize, value: Option<CassCqlValue>) -> CassError {
-        let (bound_value, maybe_data_type) = match &self.statement {
-            Statement::Simple(_) => match self.bound_values.get_mut(index) {
-                Some(v) => (v, None),
-                None => return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
-            },
-            Statement::Prepared(p) => match (
-                self.bound_values.get_mut(index),
-                p.variable_col_data_types.get(index),
-            ) {
-                (Some(v), Some(dt)) => (v, Some(dt)),
-                (None, None) => return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
-                // This indicates a length mismatch between col specs table and self.bound_values.
-                //
-                // It can only occur when user provides bad `count` value in `cass_statement_reset_parameters`.
-                // Cpp-driver does not verify that both of these values are equal.
-                // I believe returning CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS is best we can do here.
-                _ => return CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
-            },
-        };
+        match (
+            self.bound_values.get_mut(index),
+            self.statement.variable_col_data_types.get(index),
+        ) {
+            (Some(v), Some(dt)) => {
+                // Perform the typecheck.
+                if !value::is_type_compatible(&value, dt) {
+                    return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                }
 
-        // Perform the typecheck.
-        if let Some(dt) = maybe_data_type {
-            if !value::is_type_compatible(&value, dt) {
-                return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE;
+                *v = Set(value);
+                CassError::CASS_OK
             }
+            (None, None) => CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
+            // This indicates a length mismatch between col specs table and self.bound_values.
+            //
+            // It can only occur when user provides bad `count` value in `cass_statement_reset_parameters`.
+            // Cpp-driver does not verify that both of these values are equal.
+            // I believe returning CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS is best we can do here.
+            _ => CassError::CASS_ERROR_LIB_INDEX_OUT_OF_BOUNDS,
+        }
+    }
+
+    fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
+        let mut name_str = name;
+        let mut is_case_sensitive = false;
+
+        if name_str.starts_with('\"') && name_str.ends_with('\"') {
+            name_str = name_str.strip_prefix('\"').unwrap();
+            name_str = name_str.strip_suffix('\"').unwrap();
+            is_case_sensitive = true;
         }
 
-        *bound_value = Set(value);
-        CassError::CASS_OK
+        let indices: Vec<usize> = self
+            .statement
+            .statement
+            .get_variable_col_specs()
+            .iter()
+            .enumerate()
+            .filter(|(_, col)| {
+                is_case_sensitive && col.name() == name_str
+                    || !is_case_sensitive && col.name().eq_ignore_ascii_case(name_str)
+            })
+            .map(|(i, _)| i)
+            .collect();
+
+        if indices.is_empty() {
+            return CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST;
+        }
+
+        self.bind_multiple_values_by_name(&indices, value)
     }
 
     fn bind_multiple_values_by_name(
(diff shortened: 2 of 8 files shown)


Lorak-mmk:
> 4 out of 6 tests from this suite are passing. We can enable them.
> 
> When it comes to SimpleStatementAnyOrder and SimpleStatementInvalidName, I don't think we will ever enable them. The reason is that rust-driver no longer supports 'WITH_NAMES_FOR_VALUES' flag (https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L379) since serialization refactor, thus does not support "by name" value binding.
> 

I don't get it. Rust Driver prepares the statements underneath and should be able to handle named bind markers in queries.

> The reason that analogous tests for prepared statements pass is because "name-to-indexes" mapping can be retrieved from the prepared statement's metadata received from the server upon preparation. Thanks to that, we can bind the values by name in correct order (in cpp-rust-driver layer).

muzarski:
> I don't get it. Rust Driver prepares the statements underneath and should be able to handle named bind markers in queries.

You are right. Probably, the custom `SerializeRow` implementation will do the job. I'll try it out

muzarski:
I updated the description.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -93,15 +93,16 @@ impl CassStatement {
...
-            is_case_sensitive = true;
-        }
+        // A free index for a value. Used only if the statement is an unprepared query.
+        let free_index: Option<usize>;
+
+        // If the name was quoted, then we should treat it as case sensitive.
+        let (name_unquoted, is_case_sensitive) = if name.starts_with('\"') && name.ends_with('\"') {
+            let name_unquoted = name.strip_prefix('\"').unwrap().strip_suffix('\"').unwrap();
+            (name_unquoted, true)
+        } else {
```
wprzytula: The `unwrap`s can be avoided.

wprzytula:
```suggestion
        // If the name was quoted, then we should treat it as case sensitive.
        let (name_unquoted, is_case_sensitive) = match name.strip_prefix('\"').and_then(|s| s.strip_suffix('\"')) {
            Some(stripped) => (stripped, true),
            None => (name, false),
        }
```

scylla-rust-wrapper/src/statement.rs
```diff
@@ -93,15 +93,16 @@ impl CassStatement {
     }
 
     fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
```
wprzytula, line 229: My suggested rewrite of this function for more readability, fewer `unwrap`s and more FP control flow:
(see how I tamed the borrow checker :sunglasses: )
```rust
fn bind_cql_value_by_name(&mut self, name: &str, value: Option<CassCqlValue>) -> CassError {
    // If the name was quoted, then we should treat it as case sensitive.
    let (name_unquoted, is_case_sensitive) =
        match name.strip_prefix('\"').and_then(|s| s.strip_suffix('\"')) {
            Some(name_unquoted) => (name_unquoted, true),
            None => (name, false),
        };

    match &self.statement {
        Statement::Prepared(prepared) => {
            let indices: Vec<usize> = prepared
                .statement
                .get_variable_col_specs()
                .iter()
                .enumerate()
                .filter(|(_, col)| {
                    is_case_sensitive && col.name == name_unquoted
                        || !is_case_sensitive && col.name.eq_ignore_ascii_case(name_unquoted)
                })
                .map(|(i, _)| i)
                .collect();

            if indices.is_empty() {
                CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
            } else {
                self.bind_multiple_values_by_name(&indices, value)
            }
        }
        Statement::Simple(query) => {
            let index = query.name_to_bound_index.get(name_unquoted);

            if let Some(idx) = index {
                self.bind_cql_value(*idx, value)
            } else {
                // Find first free index, i.e. index where value at this index is Unset.
                let free_index = self.bound_values.iter().position(|v| matches!(v, Unset));
                if let Some(index) = free_index {
                    if let Statement::Simple(query) = &mut self.statement {
                        query
                            .name_to_bound_index
                            .insert(name_unquoted.to_string(), index);
                    }
                    self.bind_cql_value(index, value)
                } else {
                    CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
                }
            }
        }
    }
}
```

scylla-rust-wrapper/src/statement.rs
```diff
@@ -33,6 +35,72 @@ pub struct SimpleQuery {
...
+                        })?;
+
+                    Ok(self.bound_values[*bound_values_idx].clone())
+                })
+                .collect::<Result<_, _>>()?
+        };
+
+        // This will validate the length as well.
+        <_ as SerializeRow>::serialize(bound_values, ctx, writer)
+    }
```
wprzytula: I'm moderately worried about overhead introduced by `by_name` binding. Can we at least perform some optimisations, such as using `VecMap` instead of `HashMap`, and `SmallVec` instead of `Vec`?
(It doesn't have to be in this PR, ofc)

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -59,7 +59,16 @@ impl From<&BadQuery> for CassError {
             BadQuery::ValuesTooLongForKey(_usize, _usize2) => CassError::CASS_ERROR_LAST_ENTRY,
             BadQuery::BadKeyspaceName(_bad_keyspace_name) => CassError::CASS_ERROR_LAST_ENTRY,
             BadQuery::Other(_other_query) => CassError::CASS_ERROR_LAST_ENTRY,
-            BadQuery::SerializationError(_) => CassError::CASS_ERROR_LAST_ENTRY,
+            BadQuery::SerializationError(e) => {
+                // TODO: make the member of SerializationError public in rust-driver.
+                // There is no other way to check the actual error.
```
wprzytula: I encourage you to make a PR to Rust driver that publicises the field of all of: `SerializationError`, `DeserializationError`, `TypeCheckError`. I'd prefer not to merge this as-is, with string textual matching.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -59,7 +59,16 @@ impl From<&BadQuery> for CassError {
...
+                if e.to_string().contains("Unknown named parameter:") {
+                    // It means that our custom `UnknownNamedParameterError` was returned.
```
wprzytula: Oh, allocations upon error conversion, noooo...

tests/src/integration/tests/test_named_parameters.cpp
```diff
@@ -204,7 +204,7 @@ CASSANDRA_INTEGRATION_TEST_F(NamedParametersTests, SimpleStatementInvalidName) {
   insert_statement.bind<Uuid>("named_uuid", value_uuid_);
   insert_statement.bind<Blob>("named_blob", value_blob_);
   insert_statement.bind<List<Float> >("named_list_floats", value_list_floats_);
-  EXPECT_EQ(CASS_ERROR_SERVER_INVALID_QUERY,
+  EXPECT_EQ(CASS_ERROR_LIB_NAME_DOES_NOT_EXIST,
             session_.execute(insert_statement, false).error_code());
 }
```
wprzytula, line 209: Is this a change wrt cpp-driver, or rather a revert to cpp-driver's original test?

wprzytula (changes requested):

tests/src/integration/tests/test_named_parameters.cpp
muzarski: Cpp-driver supports this flag: https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L379.

So what it does for unprepared queries (when binding by name), is it just simply sends the request with bad names - server performs name validation and returns an error. In our case, we do the validation on driver's side (because rust-driver prepares simple queries under the hood, to receive the metadata). We need to adjust the test to this logic - i.e. expect a validation error from the driver (`ERROR_LIB...`), and not server error (`ERROR_SERVER...`).

scylla-rust-wrapper/src/cass_error.rs
muzarski: Opened PR: https://github.com/scylladb/scylla-rust-driver/pull/1087

muzarski:
v2: rebased on https://github.com/scylladb/cpp-rust-driver/pull/200. We don't match the specific `SerializationError` by stringification anymore - `downcast_ref()` method is used.

muzarski:
v3: rewritten the `bind_cql_value_by_name` method according to @wprzytula suggestions - thank you! Now, the method logic is much clearer since we got rid of early returns.

scylla-rust-wrapper/src/statement.rs
muzarski: Opened an issue not to forget about it: https://github.com/scylladb/cpp-rust-driver/issues/201

muzarski:
Rebased on master

wprzytula (approved):

wprzytula:
The PR's name is a bit misleading. It sounds like it changes only tests and enables them; instead, it fixes **bugs** in the driver! This should definitely be indicated in the name, and also prioritised higher.

muzarski:
> The PR's name is a bit misleading. It sounds like it changes only tests and enables them; instead, it fixes **bugs** in the driver! This should definitely be indicated in the name, and also prioritised higher.

Good point. Adjusted the title

muzarski:
Rebased on master

wprzytula (approved):

scylla-rust-wrapper/src/statement.rs
```diff
@@ -110,8 +110,8 @@ impl CassStatement {
                     .iter()
                     .enumerate()
                     .filter(|(_, col)| {
-                        is_case_sensitive && col.name() == name_str
-                            || !is_case_sensitive && col.name().eq_ignore_ascii_case(name_str)
+                        is_case_sensitive && col.name() == name_unquoted
+                            || !is_case_sensitive && col.name().eq_ignore_ascii_case(name_unquoted)
                     })
                     .map(|(i, _)| i)
                     .collect();
```
Lorak-mmk: 📌 Can we make `bind_multiple_values_by_name` get impl Iterator or something like this, and avoid allocating a vector here? We should make prepared statements fast.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -128,17 +128,14 @@ impl CassStatement {
...
                     return self.bind_cql_value(*idx, value);
                 } else {
-                    for (index, bound_val) in self.bound_values.iter().enumerate() {
-                        if let Unset = bound_val {
-                            set_bound_val_index = Some(index);
-                            break;
-                        }
-                    }
+                    // Find first free index, i.e. index where value at this index is Unset.
+                    free_index = self.bound_values.iter().position(|v| matches!(v, Unset));
```
Lorak-mmk: ❓ For unprepared statements the names are always case sensitive with the current implementation. Is that also the case for cpp-driver?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -117,33 +114,32 @@ impl CassStatement {
...
-                    return self.bind_cql_value(*idx, value);
+                    self.bind_cql_value(*idx, value)
                 } else {
                     // Find first free index, i.e. index where value at this index is Unset.
-                    free_index = self.bound_values.iter().position(|v| matches!(v, Unset));
+                    let free_index = self.bound_values.iter().position(|v| matches!(v, Unset));
+                    if let Some(index) = free_index {
+                        if let Statement::Simple(query) = &mut self.statement {
+                            query.name_to_bound_index.insert(name.to_string(), index);
+                        }
```
Lorak-mmk: 🔧 This is all under a `Statement::Simple(query) => {` match clause already. 
Why do you match the same thing again? The presence of "if" suggests that this may not be a Simple variant, which I don't think is possible.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -32,6 +35,72 @@ pub struct SimpleQuery {
...
+}
+
+#[derive(Debug)]
+pub struct UnknownNamedParameterError(String);
+
+impl std::fmt::Display for UnknownNamedParameterError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Unknown named parameter: \"{}\"", self.0)
+    }
+}
```
Lorak-mmk: ❓ In Rust Driver we generate those impls using "thiserror" crate, right? Do you think it is worth it to use it here too or is there not enough such impls to warrant adding a dependency?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -32,6 +35,72 @@ pub struct SimpleQuery {
...
+                        })?;
+
+                    Ok(self.bound_values[*bound_values_idx].clone())
+                })
+                .collect::<Result<_, _>>()?
+        };
+
+        // This will validate the length as well.
+        <_ as SerializeRow>::serialize(bound_values, ctx, writer)
+    }
```
Lorak-mmk: 🔧 In the named variant, instead of collecting to a new vector, can you iterator over the values and serialize them?
This would avoid vector allocation, and possibly the `.clone()` in `self.bound_values[*bound_values_idx].clone()` too.
`.try_for_each()` iterator method would probably be the best fit for this.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/statement.rs
muzarski: The first time, we match the `&self.statement` (non-mutable reference). If you tried matching `&mut self.statement`, we get a compiler error, because inside the match statement we access `& mut self` (e.g. in `self.bind_cql_value`). This `if let ... = &mut self.statement` is a smart trick by @wprzytula. It allows us to handle everything inside the `match` statement in a way so compiler does not complain anymore - we reduce the mutability scope of `&mut self.statement`.

scylla-rust-wrapper/src/statement.rs
muzarski: At the time I was writing this code, we did not have a `thiserror` dependency in cpp-rust-driver. I thought that it was not worth introducing it for this use-case. However, https://github.com/scylladb/cpp-rust-driver/pull/206 introduced `thiserror` dependency.

This is a good observation. I'll adjust it, so we use `thiserror` here.

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: In that case you could use use `let ... = ... { something} else {panic...}` to show the intent better.
1. This removes the confusion of "if"
2. Its better to panic than silently do the wrong thing

cc: @wprzytula

scylla-rust-wrapper/src/statement.rs
muzarski: I really like this suggestion. Applied it locally.

scylla-rust-wrapper/src/statement.rs
muzarski: That's what I tried to do in the first place. However, I stumbled upon the following issue: the iterator borrows from `&self`, and then we try to pass it to the function that accepts `&mut self`. The compiler obviously complains. I wonder if there is any (clean) workaround solution.

scylla-rust-wrapper/src/statement.rs
muzarski: I remember investigating this. A short answer is yes.

If you are interested, here is a thorough explanation:
There is an integration test that checks case insensitivity for prepared statements. Table and statement definitions:
```c
#define TABLE_FORMAT           \
  "CREATE TABLE %s ("          \
  "key timeuuid PRIMARY KEY, " \
  "a int, "                    \
  "b boolean, "                \
  "c text, "                   \
  "abc float, "                \
  "\"ABC\" float, "            \
  "\"aBc\" float"              \
  ")"

#define INSERT_CASE_SENSITIVE_FORMAT \
  "INSERT INTO %s "                  \
  "(key, abc, \"ABC\", \"aBc\") "    \
  "VALUES (?, ?, ?, ?)"
```

It means that non-pk bind markers will have following names: `abc`, `ABC` and `aBc`. 

Then there is a snippet from the test case (`ByNameTests` suite, `MultipleBinds` test case):
```c
Prepared prepared =
      session_.prepare(format_string(INSERT_CASE_SENSITIVE_FORMAT, table_name_.c_str()));
Statement statement = prepared.bind();

...

statement.bind<TimeUuid>("key", key);
statement.bind<Float>("abc", Float(1.23f)); // This should bind to columns `abc`, `ABC`, and `aBc`

...

ASSERT_EQ(Float(1.23f), row.column_by_name<Float>("\"abc\""));
ASSERT_EQ(Float(1.23f), row.column_by_name<Float>("\"ABC\""));
ASSERT_EQ(Float(1.23f), row.column_by_name<Float>("\"aBc\""));
```

There is no such test case for simple (unprepared) statements. I defined such test case, but it fails with following error:
```
Invalid query: Invalid amount of bind variables: expected 4 received 2
```

muzarski: And the behaviour is the same in current (from this PR) cpp-rust-driver implementation - we get an `UnknownNamedParameterError`.

Or rather equivalent, not the same. Keep in mind that cpp-driver uses https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L379-L385. In rust-driver we dropped it in favor of preparing the unprepared statement before execution.

scylla-rust-wrapper/src/statement.rs
muzarski: Applied the suggestion

muzarski:
v3.1: Applied some of the suggestions by @Lorak-mmk

scylla-rust-wrapper/src/statement.rs
wprzytula: Let's have an issue about this, as I think it will require more thorough lifetime/ownership analysis.

muzarski: https://github.com/scylladb/cpp-rust-driver/issues/214

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: I'm curious, do you know if it is possible to use unprepared statements in the batch in cpp-driver? I wonder how it behaves then given that named parameters are broken in batches: https://github.com/apache/cassandra/blob/3078aea1cfc70092a185bab8ac5dc8a35627330f/doc/native_protocol_v4.spec#L444-L450

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: Notice that `bind_cql_value_by_name` matches `self.statement`, then calls `bind_multiple_values_by_name` in Prepared case, which in turn calls `bind_cql_value`, which again matches `self.statement`.
This suggests to me that we should extend `Statement` enum with bind values, it would look like:
```
struct SimpleStatementWithValues {
    statement: SimpleQuery
    values: bound_values: Vec<MaybeUnset<Option<CassCqlValue>>>
}
struct PreparedStatementWithValues {
    statement: Arc<CassPrepared>
    values: bound_values: Vec<MaybeUnset<Option<CassCqlValue>>>
}

enum StatementWithParams
{
 Simple(SimpleStatementWithValues)
 Prepared(PreparedStatementWithValues)
}
```

`bind_cql_value_by_name` and `bind_cql_value` would be defined on `SimpleStatementWithValues` and `PreparedStatementWithValues`. The implementation in  `CassStatement` would match the enum and forward to them.
Now the impls of `bind_cql_value_by_name` don't need to borrow from self for match, and so are free to call another methods. This solves both this issue, and the one where I proposed `let ... = ... { something} else {panic...}`

Lorak-mmk: 📝 `PreparedStatementWithValues` is basically a `BoundStatement` - so when we introduce `BoundStatement` to Rust Driver we might be able to utilize it here.

muzarski: Your suggestion has a lof of sense. Should I do the refactor as a part of this PR?

Lorak-mmk: Unless it turns out to be a lot of work for some reason I don't see why not.

scylla-rust-wrapper/src/statement.rs
muzarski: It's possible, but the values can only be bound by index: https://github.com/scylladb/cpp-driver/blob/master/src/batch_request.cpp#L142-L146

muzarski:
v4:
- Applied a suggestion by @Lorak-mmk. Moved `bound_values` out of `CassStatement`, to new structs `BoundSimpleQuery` and `BoundPreparedStatement`. It reduces the complexity of binding functions, as we do not have to match against `&self.statement` in multiple places.
- Defined two new integration test cases that perform by-name binding to multiple columns using the same bind marker

scylla-rust-wrapper/src/statement.rs
muzarski: I did the refactor. This however, does not solve the issue with borrow-checker. We still try to mutate `&mut self` while iterating over an iterator that borrows from `&self` (it borrows from prepared statement's metadata).

Lorak-mmk: It can be done like this: https://github.com/scylladb/cpp-rust-driver/commit/5d3addd74161e55519c907e56547628a5704e1d2
This is a PoC, probably needs more work

muzarski: Ok, so I believe that it's out of scope of this PR

scylla-rust-wrapper/src/statement.rs
```diff
@@ -139,6 +145,63 @@ impl BoundSimpleQuery {
...
+                            SerializationError::new(UnknownNamedParameterError(
+                                col_spec.name().to_owned(),
+                            ))
+                        })?;
+
+                let val = &self.bound_values[*bound_values_idx];
+
+                <_ as SerializeValue>::serialize(val, col_spec.typ(), writer.make_cell_writer())?;
+
+                Ok(())
```
Lorak-mmk, line 200: This assumes that columns have corresponding name->index mapping, right?
What about a case where used bound some values by name and some others by index (so they don't have the mapping)?

How does cpp-driver handle this case?

muzarski: > and some others by index (so they don't have the mapping)?

It's not necessarily true that there won't be a mapping. For example, consider that user firstly binds by name: `by_name("foo", 5)`. It will store value 5 at `index 0` and remember the mapping `"foo"->0`. Now, when user calls `by_index(0, 10)`, it will overwrite the value, but will not change the name->index mapping - it is still there. So interleaving `by_name` and `by_index` calls are ok, as long as user provides mapping for each bind marker's name. If there is any missing name in `name->index` mapping, then we return `UnknownNamedParameterError`.

The behaviour I just described is cpp-driver's semantics.

muzarski:
Rebased on master

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: I'm talking about a different case.
User creates unprepared statement with 2 bound variables.
User binds `by_name("foo", 5)`, we have mapping `"foo"->0`.
User binds `by_index(1, 37)`.

Now the value at index 1 is bound, but has no name mapping, right?
The code in your PR doesn't handle this case I think. How does cpp-driver behave?

muzarski: cpp-driver then ignores the value at index `1`. What cpp-driver does during serialization, is it iterates over the hashmap (their custom implementation):
```c
// Format: [<name_1><value_1>...<name_n><value_n>]
// where:
// <name> is a [string]
// <value> is a [bytes]
int32_t QueryRequest::encode_values_with_names(ProtocolVersion version, RequestCallback* callback,
                                               BufferVec* bufs) const {
  int32_t size = 0;
  for (size_t i = 0; i < value_names_->size(); ++i) {
    const Buffer& name_buf = (*value_names_)[i].buf;
    bufs->push_back(name_buf);

    Buffer value_buf(elements()[i].get_buffer());
    bufs->push_back(value_buf);

    size += name_buf.size() + value_buf.size();
  }
  return size;
}
```

For some context: `value_names_` is their implementation of hashmap (our `name_to_bound_index`). `elements()` returns the vector that stores the values (our `bound_values`).

In the case you provided, the value at index `1` will be simply ignored. This is what we do in cpp-rust-driver - we iterate over column specs, then based on them and `name_to_bound_index`, we retrieve the values (ignoring the values that do not have the mapping).

I see one issue, though...
Assume that user declared 2 parameters. Then he does:
- `by_name("foo", 5)` <- stores value 5 at index 0, remembers `"foo"->0` mapping
- `by_index(1, 19)` <- stores value 19 at index 1
- `by_name("bar", 10)` <- tries to find a free index for "bar", but does not find one! It's because we search for FIRST value that is `Unset`, but our vector is already filled (user declared 2 parameters).

In case of cpp-driver, it would overwrite the value at index 1, and remember the mapping `"bar"->1`.

I see two solutions.
1. Fix the behaviour so we are compatible with cpp-driver, and introduce some integration test for this
2. Completely drop the compatibility with cpp-driver in regards to interleaving by_name and by_index bindings.
We could introduce some "binding state" e.g.:
```rust
enum BindingState {
    Init,
    ByName,
    ByIndex { name_to_bound_index: ... },
}
```

The initial state would obviously be `BindingState::Init`. The transitions are obvious:
- Init -> ByName on first by_name bind
- Init -> ByIndex on first by_index bind
- By[Name/Index] -> Init on `cass_statement_reset_parameters`

We would return an error when user binds by name in `ByIndex` state, and vice versa.

I'm slightly leaning towards the second option, since this is safer and who knows how many edge cases there are that we haven't discovered yet.

cc: @wprzytula

muzarski:
v4.1: Thanks to @Lorak-mmk I discovered a small inconsistency between cpp-rust and cpp-driver in regards to interleaving by name and by index bindings. I fixed it (commit: `binding: adjust finding new index to cpp-driver semantics`) and introduced integration test cases that verify this (last commit). I checked that the same test cases pass for cpp-driver.

scylla-rust-wrapper/src/statement.rs
muzarski: I went with the first approach. Already pushed the fixes (and integration tests)

Lorak-mmk (approved):

wprzytula (approved):
I think I'd feel more secure about these changes with at least a number of unit tests of the `BoundStatement`. There are integration tests, that's true, but for some reason I tend to trust Rust unit tests targetting the specific area more.

muzarski:
> I think I'd feel more secure about these changes with at least a number of unit tests of the `BoundStatement`. There are integration tests, that's true, but for some reason I tend to trust Rust unit tests targetting the specific area more.

I think it would be possible to implement unit tests for simple queries, if I extracted another struct containing binding payload (everything from `BoundSimpleQuery` apart from `Query`). It would be possible because `Query` is not required during binding, and we could mock metadata for serialization.

However, I don't see how we could implement unit tests for prepared statements. Does rust driver allow to somehow mock `PreparedStatement` structure?

Anyhow, I'll merge it as is. There are multiple places where Rust unit tests would be useful, but currently (AFAIK) we don't have a way to mock structures from rust-driver.

PR #188, opened 2024-10-10
build: generate bindings for basic types

muzarski:
## Motivation

All of the numeric types were for some reason defined by us in `types.rs`. We should let `bindgen` generate bindings for them.

This is what this PR does. Removes basic types definitions from `types.rs` and tells bindgen to generate bindings for these types.

## size_t
Second commit addresses the issue with `size_t` binding. There is a question regarding whether we should treat it as `usize` or not. Please, take a look at that.

## Modules refactor
Moved all of the `include!`s throughout code to one place - `lib.rs`. Each `include!` now corresponds to its module.

## Pre-review checklist


Lorak-mmk (approved):
I think the current version of the PR is fine (meaning `size_t_is_usize` should remain false). I see it as a safer option - there is no way for C definition and our definition to diverge.

Note: now the `types.rs` file doesn't actually define any types. We should probably remove the of the definitions from it.

wprzytula (approved):
Great to see this fixed.

Lorak-mmk:
> I think the current version of the PR is fine (meaning `size_t_is_usize` should remain false). I see it as a safer option - there is no way for C definition and our definition to diverge.
> 
> Note: now the `types.rs` file doesn't actually define any types. We should probably remove the of the definitions from it.

Actually we could, for each generated module, have an entry in `lib.rs`:
```
mod generated_mod {
    include!(concat!(env!("OUT_DIR"), "/generated_mod.rs"));
}
```

That way, `lib.rs` would be the only file in which we use `include!` macro - the rest of the project could just use the modules as usual.

wprzytula:
> That way, `lib.rs` would be the only file in which we use `include!` macro - the rest of the project could just use the modules as usual.

Yes, please. I hate those `include!`s.

muzarski:
> > I think the current version of the PR is fine (meaning `size_t_is_usize` should remain false). I see it as a safer option - there is no way for C definition and our definition to diverge.
> > Note: now the `types.rs` file doesn't actually define any types. We should probably remove the of the definitions from it.
> 
> Actually we could, for each generated module, have an entry in `lib.rs`:
> 
> ```
> mod generated_mod {
>     include!(concat!(env!("OUT_DIR"), "/generated_mod.rs"));
> }
> ```
> 
> That way, `lib.rs` would be the only file in which we use `include!` macro - the rest of the project could just use the modules as usual.

Done. Adjusted the names of some of the generated .rs files as well.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -57,6 +57,11 @@ pub mod cass_data_types {
...
 }
 
+/// CassConsistency and CassWriteType
+pub mod cass_misc_types {
+    include!(concat!(env!("OUT_DIR"), "/cppdriver_misc_types.rs"));
+}
+
 lazy_static! {
     pub static ref RUNTIME: Runtime = Runtime::new().unwrap();
     pub static ref LOGGER: RwLock<Logger> = RwLock::new(Logger {
```
Lorak-mmk: I'm not a fun of introducing this. "misc" is like "utils" - a pleace for people to put stuff when they don't want to think about where the proper place would be.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -62,6 +62,11 @@ pub mod cass_misc_types {
     include!(concat!(env!("OUT_DIR"), "/cppdriver_misc_types.rs"));
 }
 
+/// CassCompressionType
+pub mod cass_compression_types {
+    include!(concat!(env!("OUT_DIR"), "/cppdriver_compression_types.rs"));
+}
+
 lazy_static! {
```
Lorak-mmk: There was a file generated for batch types, which you put into "misc", but here you retain the module for compression types. Why?
Can't we retain both?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/lib.rs
muzarski: Yeah, I'm gonna be honest that I wasn't able to come up with a proper name for this module. But I'm sure it should not be called `cppdriver_data_query_error` as it was previously. Maybe we should decouple it into multiple modules. Do you have any suggestions?

muzarski: How it looks in rust-driver:
- `Consistency` is defined in `scyla-cql::frame::types.rs`
- `WriteType` is defined in `scylla_cql::frame::response::error.rs`
- `BatchType` is defined in `scylla_cql::frame::request::batch.rs`

muzarski: I think the most reasonable approach would be to introduce a separate module for each type. WDYT?

Lorak-mmk: Makes sense

muzarski:
v2: Created `cass_consistency_types` and `cass_batch_types` module (instead of having common `cass_misc_types` module). Moved `CassWriteType` to `cass_error_types` module.

scylla-rust-wrapper/build.rs
```diff
@@ -67,22 +85,29 @@ fn main() {
...
+        "cppdriver_error_types.rs",
         &[
             "CassErrorSource_",
             "CassErrorSource",
             "CassError_",
             "CassError",
+            "CassWriteType",
+            "CassWriteType_",
         ],
         &out_path,
```
Lorak-mmk, line 97: Is `CassWriteType` an error?

muzarski: It is only used in server errors. In rust-driver, `CassWriteType` is defined in a module corresponding to server errors as well.

Lorak-mmk (approved):

scylla-rust-wrapper/src/lib.rs
```diff
@@ -32,11 +33,76 @@ pub mod statement;
...
 pub mod value;
 
+/// All numeric types are defined here.
+pub mod types {
+    #![allow(non_camel_case_types)]
+    // for `cass_false` and `cass_true` globals.
+    #![allow(non_upper_case_globals)]
+
+    // Definition for size_t (and possibly other types in the future)
+    include!(concat!(env!("OUT_DIR"), "/basic_types.rs"));
```
wprzytula: Possible micro-enhancement: maybe we could introduce a macro, e.g. `include_bindgen_generated(filename)`, that would do the concatenation with `OUT_DIR` underneath? Just not to repeat ourselves everywhere we include such a file.

wprzytula (approved):

muzarski:
Rebased on main. I'll apply @wprzytula 's suggestion in the next push

wprzytula (approved):

PR #190, opened 2024-10-14
ci: add explicit step to update apt cache in jobs

muzarski:
## Pre-review checklist


muzarski:
@dkropachev I adjusted this PR, according to what we discussed during our call

PR #191, opened 2024-10-15
ci: use of ubuntu 22.04

muzarski:
## Runner image regression (`ubuntu-latest`)
With recent update of `ubuntu-latest` pointing to `ubuntu-24.04`, multiple issues started to occur during CI. As a temporary measure, to make CI green again, let's keep on using ubuntu-22.04 (previously ubuntu-latest).

## Scylla-ccm
We will now fetch most recent version of `scylla-ccm` (master) during CI

Opened an issue to update runners to `ubuntu-latest` in the future: https://github.com/scylladb/cpp-rust-driver/issues/192

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ CASSANDRA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 endif
 
 ifndef CCM_COMMIT_ID
-	export CCM_COMMIT_ID := 81076bce792a0fb3f2050e4c209a93e4a62ab55f
+	export CCM_COMMIT_ID := master
 endif
 
 ifndef SCYLLA_VERSION
--- a/.github/workflows/build-lint-and-test.yml
+++ b/.github/workflows/build-lint-and-test.yml
@@ -21,7 +21,7 @@ env:
 jobs:
   build-lint-and-unit-test:
     name: Build, lint and run unit tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -48,7 +48,7 @@ jobs:
 
   scylla-integration-tests:
     name: Scylla ITs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     needs: [build-lint-and-unit-test]
     timeout-minutes: 90
 
@@ -142,7 +142,7 @@ jobs:
           path: ${{ env.CCM_LOGS_PATTERN }}
 
   cassandra-integration-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     needs: [build-lint-and-unit-test]
 
     strategy:
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -12,7 +12,7 @@ env:
 jobs:
   build-rpm-pkgs:
     name: Build rpm packages
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     container:
       image: fedora:latest
       # Required by `mock`:
@@ -49,7 +49,7 @@ jobs:
   
   build-deb-pkgs:
     name: Build deb packages
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     strategy:
       matrix:


Makefile
```diff
@@ -55,7 +55,7 @@ CASSANDRA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 endif
 
 ifndef CCM_COMMIT_ID
-	export CCM_COMMIT_ID := 81076bce792a0fb3f2050e4c209a93e4a62ab55f
+	export CCM_COMMIT_ID := master
```
dkropachev, line 58: Don't do that. it can lead to unpredictable failures.

dkropachev (changes requested):

Makefile
Lorak-mmk: CCM is mostly used by our internal repos. It is updated often, to keep up with changes to Scylla.
If we don't use latest version then
- bugs in ccm will be unnoticed, instead of being quickly reported. CCM doesn't have good test suite, so being used in other projects is a main way to test it.
- it will lead to even harder to debug failures when our hardcoded version stops working with new scylla

In Python Driver we use master, and it doesn't cause problems.

Lorak-mmk (approved):

Makefile
muzarski: I agree with Karol.

dkropachev: You could say the exact same thing about `ubuntu-latest`, couldn't you ?

muzarski: But runner images are not maintained by Scylla. ccm is a different matter - as Karol mentioned, it's a great way to test `scylla-ccm` itself. When an issue arises, we can inform `scylla-ccm` maintainers about it. If there is no quick fix for the issue, nothing stops us from hardcoding the ccm version in CI in the future.

dkropachev: @muzarski , aren't you having a problem with `ccm` not working on `ubuntu-24.04` ?

muzarski: Only on GH runners using ubuntu-24.04. When I tried same commands on local docker ubuntu-24.04 container, everything worked as expected.

muzarski: `ubuntu-24.04` GH runner image seems to be still unstable (https://github.com/actions/runner-images/issues). I don't think it's an issue with `ccm` (but I may be wrong).

dkropachev: what I mean is this, test and build environment is a function of many variables, if you set all of them to constant you going to have constant results, more variables are variable more chances you have for it to fail. 
ccm can have best test suit in the world, but if it is not tested against exact environement you are running on, you going to have a chance for it to fail.
Having that said, I don't see any gains in pointing ccm to `master`, only loses.

Lorak-mmk: The gain is that
- We help find and fix bugs in ccm, which is a benefit for us, and any other repo that uses ccm.
- We don't gain debt by staying on years-old versions (let's be honest, no one is going to remember to update the commit sha regularly)
- If someone updates it, then they need to deal with new bugs anyway - so we either deal with problems, or stay on ancient version
- If we stay on ancient version, it is going to break at some point because of new scylla version - and then it will be much harder to fix compared to always being on master

dkropachev: > The gain is that
> 
> * We help find and fix bugs in ccm, which is a benefit for us, and any other repo that uses ccm.

Following this logic we need to run tests on `scylla:master` to help find bugs there too, why don't we do that ?
There is only one proper place for it - ccm repo and cicd. 

> * We don't gain debt by staying on years-old versions (let's be honest, no one is going to remember to update the commit sha regularly)

You don't need to, unless it does not provide functionality you need or don't work on new scylla image.
And when it happens, you are going to have nice flow of issue -> PR connected together and you will be able to see reasoning why has ccm version changed.
Opposite to silent update and no way to reason why tests are failing localy while working ok on cicd.

> * If someone updates it, then they need to deal with new bugs anyway - so we either deal with problems, or stay on ancient version
> * If we stay on ancient version, it is going to break at some point because of new scylla version - and then it will be much harder to fix compared to always being on master

But at least you will have exact reason why and what is failing, oposite to the case when cicd starts failing on a new PR, while you have not changed anything just because some bug slipped through cracks in ccm cicd.

dkropachev (approved):
@muzarski , @Lorak-mmk , i left my my reasoning here, if you still think to bind it to master, do it.

wprzytula (approved):

PR #194, opened 2024-10-16
ci: enable NullStringApiArgsTest suite

muzarski:
Makes `cass_session_prepare` logic consistent with cpp-driver (even though, I am not a fan of it) and enables `NullStringApiArgsTest` test suite.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -555,10 +555,12 @@ pub unsafe extern "C" fn cass_session_prepare_n(
     query: *const c_char,
     query_length: size_t,
 ) -> *const CassFuture {
-    let query_str = match ptr_to_cstr_n(query, query_length) {
-        Some(v) => v,
-        None => return std::ptr::null(),
-    };
+    let query_str = ptr_to_cstr_n(query, query_length)
+        // Apparently nullptr denotes an empty statement string.
+        // It seems to be intended (for some weird reason, why not save a round-trip???)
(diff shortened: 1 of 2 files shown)


muzarski:
As a small optimization, to safe a round-trip for a request that we know will fail anyway, we could return an early "server error" (syntax error) without sending the acutal request.

scylla-rust-wrapper/src/session.rs
```diff
@@ -555,10 +555,12 @@ pub unsafe extern "C" fn cass_session_prepare_n(
...
     query_length: size_t,
 ) -> *const CassFuture {
-    let query_str = match ptr_to_cstr_n(query, query_length) {
-        Some(v) => v,
-        None => return std::ptr::null(),
-    };
+    let query_str = ptr_to_cstr_n(query, query_length)
+        // Apparently nullptr denotes an empty statement string.
+        // It seems to be intended (for some weird reason, why not safe a round-trip???)
+        // to receive a server error in such case (CASS_ERROR_SERVER_SYNTAX_ERROR).
```
Lorak-mmk: safe -> save

wprzytula (approved):
Weird indeed. This is one of moments when I wonder if we should follow cpp-driver, or rather be wiser.

Lorak-mmk (approved):

Lorak-mmk:
> Weird indeed. This is one of moments when I wonder if we should follow cpp-driver, or rather be wiser.

I think we could have different (better) behavior if we are sure that it will not break apps that try to migrate to our driver.

Issue #116, opened 2023-06-21
Avoid redundant cassandra.h file


Gor027:
Currently, for building purposes, there is a duplicated `cassandra.h` file. Perhaps, it is possible to use symlinks to have a single file.

PR #195, opened 2024-10-17
lbp: rack awareness

### LINKED ISSUE #156: Implement RackAwarePolicy loadbalancing policy
We need to implement an API that matches `RackAwarePolicy` of `cpp-driver` that was implemented at https://github.com/scylladb/cpp-driver/pull/73

muzarski:
Fix: https://github.com/scylladb/cpp-rust-driver/issues/156
See: https://github.com/scylladb/cpp-driver/pull/73

## Changes
Implemented `cass_[cluster/exeuction_profile]_set_load_balance_rack_aware[_n]`. Adjusted the documentation, and extended unit tests for API related to LBP.

## Important note about documentation and default behaviour of rack aware policy
I removed this part of the docstring:
```
 * With empty local_rack and local_dc,  default local_dc and local_rack
 * is chosen from the first connected contact point,
 * and no remote hosts are considered in query plans.
 * If relying on this mechanism, be sure to use only contact
 * points from the local rack.
```

There are multiple reasons for this:
- this behaviour does not make much sense, and we should not mimic it IMO
- rust-driver does not behave like this
- this is not even true for cpp-driver

Why it's not true for cpp-driver:
If you carefully study the changes introduced to cpp-driver in the aforementioned
commit, you will notice that it's not possible for the driver to use rack aware
policy with an empty strings. This is because API functions reject
empty string, thus RackAwarePolicy object is never constructed in such case.
```cpp
CassError cass_cluster_set_load_balance_rack_aware_n(CassCluster* cluster, const char* local_dc,
                                                   size_t local_dc_length,
                                                   const char* local_rack,
                                                   size_t local_rack_length) {
  if (local_dc == NULL || local_dc_length == 0 || local_rack == NULL || local_rack_length == 0) {
    return CASS_ERROR_LIB_BAD_PARAMS;
  }
  cluster->config().set_load_balancing_policy(new RackAwarePolicy(
      String(local_dc, local_dc_length), String(local_rack, local_rack_length)));
  return CASS_OK;
}
```

Why is this part of docstring included in cpp-driver then? No idea. Maybe,
because `cass_cluster_set_load_balance_dc_aware` mentions something similar
for empty (non-specified) dc. However, in this case it's true, since dc awareness is enabled
by default in cpp-driver. See the docstring:
```
 * Configures the cluster to use DC-aware load balancing.
 * For each query, all live nodes in a primary 'local' DC are tried first,
 * followed by any node from other DCs.
 *
 * <b>Note:</b> This is the default, and does not need to be called unless
 * switching an existing from another policy or changing settings.
 * Without further configuration, a default local_dc is chosen from the
 * first connected contact point, and no remote hosts are considered in
 * query plans. If relying on this mechanism, be sure to use only contact
 * points from the local DC.
```

## Default node location preference
Cpp-driver, is dc-aware by default. This is not true for rust-driver, and probably will never be. In rust-driver, by default there are no node location preferences. I adjusted the documentation of `cass_cluster_set_load_balance_dc_aware` by removing the mention of dc awareness being enabled by default.

## Pre-review checklist

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -2233,6 +2233,46 @@ cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster,
                                          unsigned used_hosts_per_remote_dc,
                                          cass_bool_t allow_remote_dcs_for_local_cl);
 
+
+/**
+ * Configures the cluster to use Rack-aware load balancing.
+ * For each query, all live nodes in a primary 'local' rack are tried first,
+ * followed by nodes from local DC and then nodes from other DCs.
+ *
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] local_dc The primary data center to try first
+ * @param[in] local_rack The primary rack to try first
+ * @return CASS_OK if successful, otherwise an error occurred
+ */
+CASS_EXPORT CassError
+cass_cluster_set_load_balance_rack_aware(CassCluster* cluster,
+                                       const char* local_dc,
+                                       const char* local_rack);
+
+
+/**
+ * Same as cass_cluster_set_load_balance_rack_aware(), but with lengths for string
+ * parameters.
+ *
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] local_dc
+ * @param[in] local_dc_length
+ * @return same as cass_cluster_set_load_balance_dc_aware()
+ *
+ * @see cass_cluster_set_load_balance_dc_aware()
+ */
+CASS_EXPORT CassError
+cass_cluster_set_load_balance_rack_aware_n(CassCluster* cluster,
+                                         const char* local_dc,
+                                         size_t local_dc_length,
+                                         const char* local_rack,
+                                         size_t local_rack_length);
+
+
 /**
  * Configures the cluster to use token-aware request routing or not.
  *
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -1087,6 +1087,50 @@ cass_execution_profile_set_load_balance_dc_aware_n(CassExecProfile* profile,
                                                    unsigned used_hosts_per_remote_dc,
                                                    cass_bool_t allow_remote_dcs_for_local_cl);
 
+
+/**
+ * Configures the execution profile to use Rack-aware load balancing.
+ * For each query, all live nodes in a primary 'local' rack are tried first,
+ * followed by nodes from local DC and then nodes from other DCs.
+ *
+ * <b>Note:</b> Profile-based load balancing policy is disabled by default.
+ * cluster load balancing policy is used when profile does not contain a policy.
+ *
+ * @public @memberof CassExecProfile
+ *
+ * @param[in] profile
+ * @param[in] local_dc The primary data center to try first
+ * @param[in] local_rack The primary rack to try first
+ * @return CASS_OK if successful, otherwise an error occurred
+ */
+CASS_EXPORT CassError
+cass_execution_profile_set_load_balance_rack_aware(CassExecProfile* profile,
+                                                   const char* local_dc,
+                                                   const char* local_rack);
+
+
+/**
+ * Same as cass_execution_profile_set_load_balance_rack_aware(), but with lengths for string
+ * parameters.
+ *
+ * @public @memberof CassExecProfile
+ *
+ * @param[in] profile
+ * @param[in] local_dc
+ * @param[in] local_dc_length
+ * @return same cass_execution_profile_set_load_balance_rack_aware()
+ *
+ * @see cass_execution_profile_set_load_balance_rack_aware()
+ * @see cass_cluster_set_load_balance_rack_aware_n()
+ */
+CASS_EXPORT CassError
+cass_execution_profile_set_load_balance_rack_aware_n(CassExecProfile* profile,
+                                                     const char* local_dc,
+                                                     size_t local_dc_length,
+                                                     const char* local_rack,
+                                                     size_t local_rack_length);
+
+
 /**
  * Configures the execution profile to use token-aware request routing or not.
  *
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -2220,13 +2220,6 @@ cass_cluster_set_load_balance_round_robin(CassCluster* cluster);
  * For each query, all live nodes in a primary 'local' DC are tried first,
  * followed by any node from other DCs.
  *
- * <b>Note:</b> This is the default, and does not need to be called unless
- * switching an existing from another policy or changing settings.
- * Without further configuration, a default local_dc is chosen from the
- * first connected contact point, and no remote hosts are considered in
- * query plans. If relying on this mechanism, be sure to use only contact
- * points from the local DC.
- *
  * @deprecated The remote DC settings for DC-aware are not suitable for most
  * scenarios that require DC failover. There is also unhandled gap between
  * replication factor number of nodes failing and the full cluster failing. Only
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -75,9 +75,22 @@ impl LoadBalancingConfig {
             builder =
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let LoadBalancingKind::DcAware { local_dc } = load_balancing_kind {
-            builder = builder.prefer_datacenter(local_dc).permit_dc_failover(true)
+
+        match load_balancing_kind {
+            LoadBalancingKind::DcAware { local_dc } => {
+                builder = builder.prefer_datacenter(local_dc).permit_dc_failover(true)
+            }
+            LoadBalancingKind::RackAware {
+                local_dc,
+                local_rack,
+            } => {
+                builder = builder
+                    .prefer_datacenter_and_rack(local_dc, local_rack)
+                    .permit_dc_failover(true)
+            }
+            LoadBalancingKind::RoundRobin => {}
         }
+
         if self.latency_awareness_enabled {
             builder = builder.latency_awareness(self.latency_awareness_builder);
         }
@@ -99,7 +112,13 @@ impl Default for LoadBalancingConfig {
 #[derive(Clone, Debug)]
 pub(crate) enum LoadBalancingKind {
     RoundRobin,
-    DcAware { local_dc: String },
+    DcAware {
+        local_dc: String,
+    },
+    RackAware {
+        local_dc: String,
+        local_rack: String,
+    },
 }
 
 #[derive(Clone)]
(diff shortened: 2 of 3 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -55,11 +56,18 @@ impl LoadBalancingConfig {
...
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let Some(dc_awareness) = self.dc_awareness {
+
+        // Prioritize rack awareness if user for some reason set both rack+dc and dc preferencies.
+        if let Some(rack_awareness) = self.rack_awareness {
+            builder = builder
+                .prefer_datacenter_and_rack(rack_awareness.local_dc, rack_awareness.local_rack)
+                .permit_dc_failover(true)
+        } else if let Some(dc_awareness) = self.dc_awareness {
```
Lorak-mmk: When I see a comment like this the immediate first thought is: maybe this should be an enum, to handle this on type level instead of in runtime?

Right now the semantics are different than cpp-driver in an edge case: if you set dc aware, then rack aware, then dc aware again, cpp driver would use dc aware, but this would use rack aware.

include/cassandra.h
```diff
@@ -2206,13 +2206,6 @@ cass_cluster_set_load_balance_round_robin(CassCluster* cluster);
...
- * <b>Note:</b> This is the default, and does not need to be called unless
- * switching an existing from another policy or changing settings.
- * Without further configuration, a default local_dc is chosen from the
- * first connected contact point, and no remote hosts are considered in
- * query plans. If relying on this mechanism, be sure to use only contact
- * points from the local DC.
- *
  * @deprecated The remote DC settings for DC-aware are not suitable for most
  * scenarios that require DC failover. There is also unhandled gap between
  * replication factor number of nodes failing and the full cluster failing. Only
```
Lorak-mmk, line 2225: I generally dislike the machanism of using first contact point DC as local DC (/rack), but I wonder if this won't cause compatibility issues.

We could probably implement it by, in case dc name is empty, following mechanism:
- After session is connected check the DC of first contact point
- Then create new LBP with this DC set as local
- Swap it in session using something like this:
```rust
let handle = session.get_default_execution_profile_handle().clone();
let new_profile = handle.pointee_to_builder().load_balancing_policy(...).build();
handle.map_to_another_profile(new_profile);
```

I'm not sure preserving this mechanism is worth introducing such trickery - probably not.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cluster.rs
muzarski: > Right now the semantics are different than cpp-driver in an edge case: if you set dc aware, then rack aware, then dc aware again, cpp driver would use dc aware, but this would use rack aware.

This scenario convinced me. I'll apply the suggestion (an enum) and introduce a test for this specific edge case.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -55,9 +55,9 @@ impl LoadBalancingConfig {
             builder =
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let Some(dc_awareness) = self.dc_awareness.as_ref() {
+        if let Some(dc_awareness) = self.dc_awareness {
             builder = builder
-                .prefer_datacenter(dc_awareness.local_dc.clone())
+                .prefer_datacenter(dc_awareness.local_dc)
                 .permit_dc_failover(true)
         }
```
wprzytula: That's my boy! (that's I believe the best translation of Polish "Moja krew!") :smile:

include/cassandra.h
```diff
@@ -2219,6 +2219,46 @@ cass_cluster_set_load_balance_dc_aware_n(CassCluster* cluster,
...
+ * @param[in] local_rack The primary rack to try first
+ * @return CASS_OK if successful, otherwise an error occurred
+ */
+CASS_EXPORT CassError
+cass_cluster_set_load_balance_rack_aware(CassCluster* cluster,
+                                       const char* local_dc,
+                                       const char* local_rack);
+
+
+/**
```
wprzytula, line 2292: This is error-prone that we have two instances of `cassandra.h`. Let's try to unify them; it's too easy to make them differ.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -55,11 +56,18 @@ impl LoadBalancingConfig {
             builder =
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let Some(dc_awareness) = self.dc_awareness {
+
+        // Prioritize rack awareness if user for some reason set both rack+dc and dc preferencies.
```
wprzytula: preferencies -> preferences

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -477,6 +492,68 @@ pub unsafe extern "C" fn cass_cluster_set_load_balance_dc_aware_n(
...
+    let (local_dc, local_rack) = match (
+        ptr_to_cstr_n(local_dc_raw, local_dc_length),
+        ptr_to_cstr_n(local_rack_raw, local_rack_length),
+    ) {
+        (Some(local_dc_str), Some(local_rack_str))
+            if local_dc_length > 0 && local_rack_length > 0 =>
+        {
+            (local_dc_str.to_owned(), local_rack_str.to_owned())
+        }
+        // One of them is either null pointer, is an empty string or is not a proper utf-8.
```
wprzytula: nit: `is either null pointer` -> `either is a null pointer`

include/cassandra.h
```diff
@@ -1087,6 +1087,50 @@ cass_execution_profile_set_load_balance_dc_aware_n(CassExecProfile* profile,
...
                                                    cass_bool_t allow_remote_dcs_for_local_cl);
 
+
+/**
+ * Configures the execution profile to use Rack-aware load balancing.
+ * For each query, all live nodes in a primary 'local' rack are tried first,
+ * followed by nodes from local DC and then nodes from other DCs.
+ *
+ * <b>Note:</b> Profile-based load balancing policy is disabled by default.
+ * cluster load balancing policy is used when profile does not contain a policy.
```
wprzytula, line 1097: How can we make a profile contain a load balancing policy? This is unclear to me from this docstring; is it clear from some other documentation?

muzarski:
v2: Introduced `NodeLocationPreferences` enum. Removed `DcAwareness` and `RackAwareness` structs.

include/cassandra.h
wprzytula: Wow! Is it going to be the first time when profile mapping feature is actually useful (and used by anyone!) ?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -929,6 +929,49 @@ mod tests {
                     }
                     assert!(!cluster.load_balancing_config.token_awareness_enabled);
                     assert!(cluster.load_balancing_config.latency_awareness_enabled);
+
+                    // set preferred rack+dc
+                    assert_cass_error_eq!(
+                        cass_cluster_set_load_balance_rack_aware(
+                            cluster_raw,
+                            "eu-east\0".as_ptr() as *const i8,
+                            "rack1\0".as_ptr() as *const i8,
```
wprzytula: Let's use C str literals, as they've been stabilised: https://doc.rust-lang.org/nightly/edition-guide/rust-2021/c-string-literals.html

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -929,6 +929,49 @@ mod tests {
...
+                            assert_eq!(local_rack, "rack1");
+                        }
+                        _ => panic!("Expected preferred dc and rack"),
+                    }
+
+                    // set back to preferred dc
+                    assert_cass_error_eq!(
+                        cass_cluster_set_load_balance_dc_aware(
+                            cluster_raw,
+                            "eu\0".as_ptr() as *const i8,
```
wprzytula: ditto: C str

wprzytula (changes requested):

include/cassandra.h
wprzytula: (ofc out of scope of this PR)

muzarski: There is an issue regarding it https://github.com/scylladb/cpp-rust-driver/issues/116. I haven't investigated it yet - I'm not exactly sure why, but it looks like currently we need both of them, i.e. we can't simply remove one of them without making any additional adjustments.

include/cassandra.h
Lorak-mmk: As I said, I'm not convinced we should do this - actually, I'm more and more convinced that we shouldn't. So unfortunately it won't :(

include/cassandra.h
wprzytula: What about symlinks?

include/cassandra.h
muzarski: cpp-rust-driver's session wrapper additionally holds a map of execution profiles, indexed by their custom names. You can firstly construct a profile, and provide custom options to it, and then pass it to `CassCluster` via `cass_cluster_set_execution_profile` with a name chosen for this profile.

Then, having multiple execution profiles defined and stored (each with unique name), you can "bind" one of your choice to some statement (see `cass_statement_set_execution_profile`), identifying the chosen profile by its name.

include/cassandra.h
muzarski: I'll try it out.

muzarski: Looks good. Build is successful: https://github.com/scylladb/cpp-rust-driver/pull/196

scylla-rust-wrapper/src/cluster.rs
muzarski: Comment removed after refactor introducing `NodeLocationPreference`.

scylla-rust-wrapper/src/cluster.rs
muzarski: Done. BTW, I cannot replace `let empty_str = "\0".as_ptr() as *const i8;` with `let empty_str = c"";` It results in a following `ntest::timeout` proc macro panic:

```
error: custom attribute panicked
   --> src/cluster.rs:884:5
    |
884 |     #[ntest::timeout(100)]
    |     ^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: message: Unrecognized literal: `c""`
```

It happens for any c str literal constructed outside of some other macro (e.g. `assert_cass_error_eq`).

include/cassandra.h
wprzytula: I know the mechanism you described, it was me who implemented it :)
The thing is, what does it mean that a profile contains a policy?

scylla-rust-wrapper/src/cluster.rs
wprzytula: Oh, maybe you should open an issue to `ntest`?

include/cassandra.h
muzarski: Ok, now I understand your question. So I studied the logic of cpp-driver, and it seems that this statement is true for cpp-driver. See: https://github.com/scylladb/cpp-driver/blob/master/src/request_processor.cpp#L196-L209

If user does not define an LBP for execution profile (i.e. exec profile does not CONTAIN lbp), then the session-wide LBP is used for this execution profile. So it means, we are inconsistent with cpp-driver. Consider following scenario:
- user sets session-wide LBP to dc aware: `cass_cluster_set_load_balance_dc_aware(...)`
- user creates some exeuction profile, but does not define LBP for it (let's call this exec profile "foo")
- now, when user executes a statement with a "foo" exec profile:
  - in cpp-driver, the session-wide DC-aware LBP is chosen for this statement execution
  - in cpp-rust-driver/rust-driver, the `DefaultPolicy::default()` is always chosen, no matter what session-wide LBP is set to

I believe that if we wanted to mimic this behaviour, then it's relatively easy to implement - no changes to `rust-driver` are required. The question is, however, whether we want to mimic such behaviour?

cc: @Lorak-mmk

Lorak-mmk: I don't see anything wrong with such behavior. It actually seems quite useful - if you want to overwrite some things (e.g. CL and timeout) for some statements but don't want to overwrite everything. As this behavior is not buggy / footgun etc I see no reason to change it.


Actually I didn't even know that Rust Driver worked this way, I thought it behaved like cpp-driver in this regard :D

muzarski: Opened a PR: https://github.com/scylladb/cpp-rust-driver/pull/197

wprzytula: > Actually I didn't even know that Rust Driver worked this way, I thought it behaved like cpp-driver in this regard :D

I can vaguely remember that we decided that overriding the global per-session execution only partly with per-statement exec profile is too complex (for a user). A profile should be considered a full configuration. If only some options are to be changed for a statement compared to the session globally, it's adviced to derive an execution profile - either by cloning the builder beforehand or by calling `to_builder()` on an existing profile.

muzarski:
v2.1: rebased on master. (now there is only one `cassadra.h` header.

Still waiting for: https://github.com/scylladb/cpp-rust-driver/pull/197

muzarski:
v3: Rebased on master, and addressed all comments. Moving out of draft, since it's ready for review again.

include/cassandra.h
Lorak-mmk, line 1: Commit: " cluster: define and implement cass_cluster_set_load_balance_rack_aware[_n]"

Nit: First line of commit message is too long which causes GH to display it badly. I think you can drop "define and".

After reading the commit message, I wonder - do we handle DC aware like cpp-driver? Meaning that we treat first contact point as local DC. If not, should we even try? Or just remove this from docs?

Lorak-mmk: Ah, I see you removed it in the third commit. Maybe we should have an issue to discuss this? I'm generally all for improving weird behavior of cpp-driver, even if the change would be slightly breaking. In this case I wonder if it isn't too breaking. Isn't this something that users could depend on?

I guess providing a migration guide could be an option? That would still benefit from an issue so we don't forget about this incompatibility.

Lorak-mmk (changes requested):

include/cassandra.h
muzarski: Opened an issue: https://github.com/scylladb/cpp-rust-driver/issues/204

muzarski:
v3.1: fixed commit titles and opened and issue regarding default local DC

Lorak-mmk (approved):

PR #197, opened 2024-10-21
exec_profile: Use cluster lbp for exec profiles by default

muzarski:
The documentation of `cass_execution_profile_set_load_balance_...` functions states:
```
 <b>Note:</b> Profile-based load balancing policy is disabled by default.
 cluster load balancing policy is used when profile does not contain a policy.
```
It seems to be true, see: https://github.com/scylladb/cpp-driver/blob/master/src/request_processor.cpp#L196-L209.

We should adjust cpp-rust-driver so it behaves the same way.

This PR consist mostly of some refactor commits that prepare for the logic change in the last commit.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -65,9 +65,9 @@ impl LoadBalancingConfig {
             builder =
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let Some(dc_awareness) = self.dc_awareness.as_ref() {
+        if let Some(dc_awareness) = self.dc_awareness {
             builder = builder
-                .prefer_datacenter(dc_awareness.local_dc.clone())
+                .prefer_datacenter(dc_awareness.local_dc)
                 .permit_dc_failover(true)
         }
         if self.latency_awareness_enabled {
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -51,7 +51,7 @@ const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 pub(crate) struct LoadBalancingConfig {
     pub(crate) token_awareness_enabled: bool,
     pub(crate) token_aware_shuffling_replicas_enabled: bool,
-    pub(crate) dc_awareness: Option<DcAwareness>,
+    pub(crate) load_balancing_kind: LoadBalancingKind,
     pub(crate) latency_awareness_enabled: bool,
     pub(crate) latency_awareness_builder: LatencyAwarenessBuilder,
 }
@@ -65,10 +65,8 @@ impl LoadBalancingConfig {
             builder =
                 builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
         }
-        if let Some(dc_awareness) = self.dc_awareness {
-            builder = builder
-                .prefer_datacenter(dc_awareness.local_dc)
-                .permit_dc_failover(true)
+        if let LoadBalancingKind::DcAware { local_dc } = self.load_balancing_kind {
+            builder = builder.prefer_datacenter(local_dc).permit_dc_failover(true)
         }
         if self.latency_awareness_enabled {
             builder = builder.latency_awareness(self.latency_awareness_builder);
@@ -81,7 +79,7 @@ impl Default for LoadBalancingConfig {
         Self {
             token_awareness_enabled: true,
             token_aware_shuffling_replicas_enabled: true,
-            dc_awareness: None,
+            load_balancing_kind: LoadBalancingKind::RoundRobin,
             latency_awareness_enabled: false,
             latency_awareness_builder: Default::default(),
         }
@@ -89,8 +87,9 @@ impl Default for LoadBalancingConfig {
 }
 
 #[derive(Clone, Debug)]
-pub(crate) struct DcAwareness {
-    pub(crate) local_dc: String,
+pub(crate) enum LoadBalancingKind {
+    RoundRobin,
+    DcAware { local_dc: String },
 }
 
 #[derive(Clone)]
@@ -457,7 +456,7 @@ pub unsafe extern "C" fn cass_cluster_set_credentials_n(
 #[no_mangle]
 pub unsafe extern "C" fn cass_cluster_set_load_balance_round_robin(cluster_raw: *mut CassCluster) {
     let cluster = ptr_to_ref_mut(cluster_raw);
-    cluster.load_balancing_config.dc_awareness = None;
+    cluster.load_balancing_config.load_balancing_kind = LoadBalancingKind::RoundRobin;
 }
 
 #[no_mangle]
@@ -496,7 +495,7 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
         .unwrap()
         .to_string();
 
-    load_balancing_config.dc_awareness = Some(DcAwareness { local_dc });
+    load_balancing_config.load_balancing_kind = LoadBalancingKind::DcAware { local_dc };
 
     CassError::CASS_OK
 }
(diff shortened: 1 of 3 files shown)


scylla-rust-wrapper/src/cluster.rs
Lorak-mmk, line 1: `bp_config: move LoadBalancingKind to upper layer`:
After reading code changes and commit message it is still not clear to me why this change is necessary or helpful - please explain it in commit message.

This moves one field out of LoadBalancingConfig and it is not immediately obvious to me why: code users of `CassCluster` can access this field just as well not matter if it is in `CassCluster` or in `LoadBalancingConfig` (which is in `CassCluster`.

scylla-rust-wrapper/src/session.rs
```diff
@@ -112,13 +112,14 @@ impl CassSessionInner {
...
             ));
         }
+
+        let (mut session_builder, default_lbp) = session_builder_and_default_lbp_fut.await;
+
         let mut exec_profile_map = HashMap::with_capacity(exec_profile_builder_map.len());
         for (name, builder) in exec_profile_builder_map {
-            exec_profile_map.insert(name, builder.build().await.into_handle());
+            exec_profile_map.insert(name, builder.build(default_lbp.clone()).await.into_handle());
         }
```
Lorak-mmk: Why go through all the trouble of having `default_lbp` returned from other functions instead of extracting it here using `session.get_default_execution_profile_handle().pointee_to_builder()`? It looks like the code that needs default_lbp can be moved below session creation, so it should work.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/session.rs
muzarski: That's what I wanted to do in the first place. However, there is no much we can do having the builder. All of its fields are private, and we don't provide any public getters for its fields.

Lorak-mmk: Huh, you are right, I'm surprised.

Maybe we should add getters for the builder? I don't see any disadvantages of that.

cc: @wprzytula

scylla-rust-wrapper/src/cluster.rs
muzarski: That's right. We move a field from `LoadBalancingConfig` to both structs that use it: `CassCluster` and `CassExecProfile`. The reason is that cluster's LBP is not optional (and set to `LoadBalancingKind::RoundRobin` by default), while it is optional for `CassExecProfile` (transformed to `Option<LoadBalancingKind>` in last commit). We can't easily represent it while having this field in a common struct - `LoadBalancingConfig`.

Sure, we could have `LoadBalancingConfig` hold `Option<LoadBalancingKind>`, but then we would need to introduce some `unwrap()`s to the code.

If this explanation is enough, I'll update the commit message appropriately.

Lorak-mmk: I'm still not sure I understand.

CassCluster is only used to create Session, right?

Why can't it have `Option<...>`, and call `.unwrap_or_default` / `.unwrap_or_else` to provide default LBP during session building?

muzarski: Ok, then `LoadBalancingConfig::build` would look like:
```rust
    // This is `async` to prevent running this function from beyond tokio context,
    // as it results in panic due to DefaultPolicyBuilder::build() spawning a tokio task.
    pub(crate) async fn build(self) -> Arc<dyn LoadBalancingPolicy> {
        let load_balancing_kind = self
            .load_balancing_kind
            .unwrap_or(LoadBalancingKind::RoundRobin);
        let mut builder = DefaultPolicyBuilder::new().token_aware(self.token_awareness_enabled);
        if self.token_awareness_enabled {
            // Cpp-driver enables shuffling replicas only if token aware routing is enabled.
            builder =
                builder.enable_shuffling_replicas(self.token_aware_shuffling_replicas_enabled);
        }
        if let LoadBalancingKind::DcAware { local_dc } = load_balancing_kind {
            builder = builder.prefer_datacenter(local_dc).permit_dc_failover(true)
        }
        if self.latency_awareness_enabled {
            builder = builder.latency_awareness(self.latency_awareness_builder);
        }
        builder.build()
    }
```

I guess we could then adjust the logic of `CassExecProfile::build` accordingly (notice that `if let Some(...)` corresponds to optional `load_balancing_kind` contained in `LoadBalancingConfig`):
```rust
    pub(crate) async fn build(
        self,
        cluster_default_lbp: Arc<dyn LoadBalancingPolicy>,
    ) -> ExecutionProfile {
        let load_balacing = if let Some(_) = self.load_balancing_config.load_balancing_kind {
            self.load_balancing_config.build().await
        } else {
            // We don't want to call self.load_balancing_config.build()
            // as it would result in round robin lbp by default due to `unwrap_or` call inside this function.
            cluster_default_lbp
        };

        self.inner.load_balancing_policy(load_balacing).build()
    }
```

muzarski: So we can either use the code samples I provided, or move the field from `LoadBalancingConfig` (current version).

scylla-rust-wrapper/src/session.rs
muzarski: It sounds fine. But I have a minor doubt about it. All of the fields of `ExecutionProfileBuilder` are optional. It results in some weird types, such as `Option<Option<Arc<dyn SpeculativeExecutionPolicy>>>`. In this case:
- `None` indicates that no spec execution policy was chosen yet, and the default one (`None` - no speculative execution) will be chosen when building
- `Some(None)` indicates that speculative execution policy has been set to None (user called `builder.speculative_execution_policy(None)`), i.e. driver will not perform a speculative execution for this profile
- `Some(Some(...))` is obvious

Does it make sense to implement a getter that returns `Option<Option<...>>`? Is it really helpful for the user? Maybe the getter should remove one layer of `Option`ness, indicating whether the field has been set in the builder, by returning default value in case the field is not set in the builder. For example:
```rust
pub fn get_speculative_execution_policy(&self) -> Option<Arc<dyn SpeculativeExecutionPolicy>> {
    self.speculative_exeuction_policy
        .clone()
        .unwrap_or_else(defaults::speculative_execution_policy)
}
```

This obviously could cost us an additional allocation if `defaults::speculative_execution_policy()` allocated an `Arc`. This is not the case here, since it returns `None`. But `defaults::[load_balancing_policy/retry_policy]()` both allocate if we wanted to follow the same approach for lbp and retry policy getters.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: Those snippets look good to me. Will those allow you to revert `connect: pass default LBP to connect_fut `?

scylla-rust-wrapper/src/session.rs
Lorak-mmk: We can replace nested Option with an enum (NotSet / Fallback, Disabled, Enabled(policy)).

Which I think is a good idea in general - we have several `Option<Option<...>>` in the driver and it is never obvious what they mean.

scylla-rust-wrapper/src/cluster.rs
muzarski: For this, we need `ExecutionProfileBuilder` getters.

Lorak-mmk: Ok. For the getters I'd like to know what @wprzytula thinks about them.

muzarski:
v2: Reverted the changes introduced by `lbp_config: move LoadBalancingKind to upper layer` (removed this commit). Now, the `load_balancing_kind` field is optional in `LoadBalancingConfig`.

muzarski:
v3: Rebased on https://github.com/scylladb/cpp-rust-driver/pull/200. Addressed all of @Lorak-mmk comments

Lorak-mmk (approved):

muzarski:
Rebased on master

PR #199, opened 2024-10-24
cluster: tracing config options

muzarski:
Implemented:
- `cass_cluster_set_tracing_consistency`
- `cass_cluster_set_tracing_max_wait_time`
- `cass_cluster_set_tracing_retry_wait_time`

Set the defaults for this options as well.

Notice, that we needed to make use of a small hack, since rust-driver does not expose the option to set tracing info fetch timeout. Instead, it allows user to define maximum number of retries. This can be computed based on timeout and retry interval provided by user via cpp-rust-driver API.

## Pre-review checklist

- ~[ ] I have implemented Rust unit tests for the features/changes introduced.`~

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -855,6 +855,8 @@ impl TryFrom<CassConsistency> for Consistency {
             CassConsistency::CASS_CONSISTENCY_LOCAL_QUORUM => Ok(Consistency::LocalQuorum),
             CassConsistency::CASS_CONSISTENCY_EACH_QUORUM => Ok(Consistency::EachQuorum),
             CassConsistency::CASS_CONSISTENCY_LOCAL_ONE => Ok(Consistency::LocalOne),
+            CassConsistency::CASS_CONSISTENCY_LOCAL_SERIAL => Ok(Consistency::LocalSerial),
+            CassConsistency::CASS_CONSISTENCY_SERIAL => Ok(Consistency::Serial),
             _ => Err(()),
         }
     }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -22,6 +22,7 @@ use scylla::{SessionBuilder, SessionConfig};
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::future::Future;
+use std::num::NonZeroU32;
 use std::os::raw::{c_char, c_int, c_uint};
 use std::sync::Arc;
 use std::time::Duration;
@@ -43,6 +44,8 @@ const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_millis(5000);
 const DEFAULT_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(30);
 // - keepalive timeout is 60 secs
 const DEFAULT_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(60);
+// - tracing info fetch timeout is 15 millis
+const DEFAULT_TRACING_INFO_FETCH_TIMEOUT: Duration = Duration::from_millis(15);
 // - tracing info fetch interval is 3 millis
 const DEFAULT_TRACING_INFO_FETCH_INTERVAL: Duration = Duration::from_millis(3);
 // - tracing consistency is ONE
@@ -113,6 +116,11 @@ pub struct CassCluster {
     auth_password: Option<String>,
 
     client_id: Option<uuid::Uuid>,
+
+    /// The default timeout for tracing info fetch.
+    /// Rust-driver only defines the number of retries.
+    /// However, this can be easily computed: `tracing_max_wait_time / tracing_retry_wait_time`.
+    tracing_max_wait_time: Duration,
 }
 
 impl CassCluster {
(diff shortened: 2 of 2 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -855,6 +855,8 @@ impl TryFrom<CassConsistency> for Consistency {
             CassConsistency::CASS_CONSISTENCY_LOCAL_QUORUM => Ok(Consistency::LocalQuorum),
```
wprzytula, line 855: typo in commit name: `CassConsistency: adjust converion to serial variants of Consistency` -> `conversion`

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -855,6 +855,8 @@ impl TryFrom<CassConsistency> for Consistency {
...
             CassConsistency::CASS_CONSISTENCY_EACH_QUORUM => Ok(Consistency::EachQuorum),
             CassConsistency::CASS_CONSISTENCY_LOCAL_ONE => Ok(Consistency::LocalOne),
+            CassConsistency::CASS_CONSISTENCY_LOCAL_SERIAL => Ok(Consistency::LocalSerial),
+            CassConsistency::CASS_CONSISTENCY_SERIAL => Ok(Consistency::Serial),
```
wprzytula, line 859: This is a bug fix! Let's prioritise such things more, instead of packing them into PRs with new features. Bugs should be fixed ASAP!

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -159,6 +167,19 @@ pub fn build_session_builder(
         session_builder = session_builder.user(username, password)
     }
 
+    // Compute the number of retries for tracing info fetch
+    // based on the timeout and interval provided by user.
+    let tracing_info_fetch_attemps = {
```
wprzytula, line 172: typo: `attemps` -> `attempts`

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -159,6 +167,19 @@ pub fn build_session_builder(
...
+        let attemps = cluster.tracing_max_wait_time.as_millis()
```
wprzytula, line 173: ditto

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -414,6 +436,16 @@ pub unsafe extern "C" fn cass_cluster_set_request_timeout(
...
+#[no_mangle]
+pub unsafe extern "C" fn cass_cluster_set_tracing_max_wait_time(
+    cluster_raw: *mut CassCluster,
+    max_wait_time_ms: c_uint,
+) {
+    let cluster = ptr_to_ref_mut(cluster_raw);
+
+    cluster.tracing_max_wait_time = Duration::from_millis(max_wait_time_ms.into());
+}
+
```
wprzytula, line 448: Could you please add a unit test for setting all tracing parameters?

wprzytula (changes requested):

scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk: For a driver still in alpha packing a bugfix in a small PR like this is imo perfectly acceptable.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -43,6 +43,8 @@ const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_millis(5000);
 const DEFAULT_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(30);
 // - keepalive timeout is 60 secs
 const DEFAULT_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(60);
+// - tracing info fetch interval is 3 millis
+const DEFAULT_TRACING_INFO_FETCH_INTERVAL: Duration = Duration::from_millis(3);
 // - tracing consistency is ONE
```
Lorak-mmk, line 51: ❓ 3ms is the defualt in Rust Driver. Is that the default in cpp-driver too?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -159,6 +167,19 @@ pub fn build_session_builder(
...
+        let attemps = cluster.tracing_max_wait_time.as_millis()
+            / session_builder
+                .config
+                .tracing_info_fetch_interval
+                .as_millis();
+
+        NonZeroU32::new(attemps as u32).unwrap_or_else(|| NonZeroU32::new(1).unwrap())
+    };
+    session_builder = session_builder.tracing_info_fetch_attempts(tracing_info_fetch_attemps);
+
```
Lorak-mmk, line 182: ❓ 🤔💭 What is the semantics of `cass_cluster_set_tracing_max_wait_time` in cpp-driver?
If it puts a time limit on whole tracing fetch, then this is very different to what you did.
Rust Driver, afaict, doesn't have a timeout for a single attempt - only a limit on number of attempts.
So a single attempt can take arbitrarily long time, and you assume they are instant.

Maybe a correct solution here is to wrap `get_tracing_info` calls into `tokio::time::timeout`?
Actually, what is the API in cpp-driver to retrieve tracing info? I don't see any call to `get_tracing_info`.
And if there are none, then what those parameters even do?

📌 As a side note, maybe we should add a timeout for single attempt to Rust Driver?

muzarski: >  I don't see any call to `get_tracing_info`. And if there are none, then what those parameters even do?

Great observation - I was sure that there was already an implemented method which fetches the tracing info. There is no such method (even in cassandra.h) - cpp-driver docs say:
```
**Note**: The driver does not return the actual tracing data for the request. The
application itself must use the returned tracing identifier to query the tables.

## Configuration

By default, when tracing is enabled, the driver will wait for the query's tracing
data to become available in the server-side tables before setting the request's
future. The amount of time it will wait, retry, and the consistency level of the
tracing data can be controls by setting `CassCluster` configuration options.
```

So this is something we have to implement in cpp-rust-driver.

Lorak-mmk:
Closing this - we decided to not implement those functions, because of the weird semantics.
If we introduce support for tracing, we'll do it in a better way (for example allowing user to access the result).

PR #200, opened 2024-10-28
cargo: point to rust-driver 64b4afcd

muzarski:
Currently, there are some open PRs that depend on changes recently introduced to rust-driver.
Let's bump the version of rust-driver dependency.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -422,7 +422,7 @@ impl CassDataType {
 
 pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
     match column_type {
-        ColumnType::Custom(s) => CassDataType::Custom((*s).clone()),
+        ColumnType::Custom(s) => CassDataType::Custom(s.clone().into_owned()),
         ColumnType::Ascii => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_ASCII),
         ColumnType::Boolean => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BOOLEAN),
         ColumnType::Blob => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_BLOB),
@@ -459,10 +459,15 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
         } => CassDataType::UDT(UDTDataType {
             field_types: field_types
                 .iter()
-                .map(|(name, col_type)| ((*name).clone(), Arc::new(get_column_type(col_type))))
+                .map(|(name, col_type)| {
+                    (
+                        name.clone().into_owned(),
+                        Arc::new(get_column_type(col_type)),
+                    )
+                })
                 .collect(),
-            keyspace: (*keyspace).clone(),
-            name: (*type_name).clone(),
+            keyspace: keyspace.clone().into_owned(),
+            name: type_name.clone().into_owned(),
             frozen: false,
         }),
         ColumnType::SmallInt => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_SMALL_INT),
--- a/scylla-rust-wrapper/src/exec_profile.rs
+++ b/scylla-rust-wrapper/src/exec_profile.rs
@@ -376,13 +376,13 @@ pub unsafe extern "C" fn cass_execution_profile_set_retry_policy(
     profile: *mut CassExecProfile,
     retry_policy: *const CassRetryPolicy,
 ) -> CassError {
-    let retry_policy: &dyn RetryPolicy = match ptr_to_ref(retry_policy) {
-        DefaultRetryPolicy(default) => default.as_ref(),
-        FallthroughRetryPolicy(fallthrough) => fallthrough.as_ref(),
-        DowngradingConsistencyRetryPolicy(downgrading) => downgrading.as_ref(),
+    let retry_policy: Arc<dyn RetryPolicy> = match ptr_to_ref(retry_policy) {
+        DefaultRetryPolicy(default) => Arc::clone(default) as _,
+        FallthroughRetryPolicy(fallthrough) => Arc::clone(fallthrough) as _,
+        DowngradingConsistencyRetryPolicy(downgrading) => Arc::clone(downgrading) as _,
     };
     let profile_builder = ptr_to_ref_mut(profile);
-    profile_builder.modify_in_place(|builder| builder.retry_policy(retry_policy.clone_boxed()));
+    profile_builder.modify_in_place(|builder| builder.retry_policy(retry_policy));
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -23,7 +23,7 @@ pub struct CassResult {
 
 pub struct CassResultData {
     pub paging_state_response: PagingStateResponse,
-    pub col_specs: Vec<ColumnSpec>,
+    pub col_specs: Vec<ColumnSpec<'static>>,
     pub col_data_types: Arc<Vec<Arc<CassDataType>>>,
     pub tracing_id: Option<Uuid>,
 }
@@ -31,7 +31,7 @@ pub struct CassResultData {
 impl CassResultData {
     pub fn from_result_payload(
         paging_state_response: PagingStateResponse,
-        col_specs: Vec<ColumnSpec>,
+        col_specs: Vec<ColumnSpec<'static>>,
         maybe_col_data_types: Option<Arc<Vec<Arc<CassDataType>>>>,
         tracing_id: Option<Uuid>,
     ) -> CassResultData {
@@ -43,7 +43,7 @@ impl CassResultData {
             Arc::new(
                 col_specs
                     .iter()
-                    .map(|col_spec| Arc::new(get_column_type(&col_spec.typ)))
+                    .map(|col_spec| Arc::new(get_column_type(col_spec.typ())))
                     .collect(),
             )
         });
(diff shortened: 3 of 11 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -43,11 +43,24 @@ impl CassResultData {
...
+            .iter()
+            .map(|col_spec| {
+                // TODO: introduce ColumnSpec::[in]to_owned() methods in rust-driver.
+                let name = col_spec.name().to_owned();
+                let typ = col_spec.typ().clone().into_owned();
+                let table_spec = col_spec.table_spec().to_owned();
+                ColumnSpec::owned(name, typ, table_spec)
+            })
+            .collect();
+
```
Lorak-mmk: When is this called? On each query?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -43,11 +43,24 @@ impl CassResultData {
...
+        let col_specs = col_specs
+            .iter()
+            .map(|col_spec| {
+                // TODO: introduce ColumnSpec::[in]to_owned() methods in rust-driver.
+                let name = col_spec.name().to_owned();
+                let typ = col_spec.typ().clone().into_owned();
+                let table_spec = col_spec.table_spec().to_owned();
+                ColumnSpec::owned(name, typ, table_spec)
+            })
+            .collect();
```
wprzytula: I haven't included `ColumnSpec::into_owned` on purpose: it would be inefficient, as another Vec would have to be allocated.
This shouldn't be needed; the `ColumnSpec` stored in `QueryResult` (soon `LegacyQueryResult`) is `'static`.
Actually, there is a bug in `QueryResult`:
```rust
pub fn col_specs(&self) -> &[ColumnSpec] {
```

...while it should be:
```rust
pub fn col_specs(&self) -> &[ColumnSpec<'static>] {
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -23,15 +23,15 @@ pub struct CassResult {
...
+    pub col_specs: Vec<ColumnSpec<'static>>,
     pub col_data_types: Arc<Vec<Arc<CassDataType>>>,
     pub tracing_id: Option<Uuid>,
 }
 
 impl CassResultData {
     pub fn from_result_payload(
         paging_state_response: PagingStateResponse,
-        col_specs: Vec<ColumnSpec>,
+        col_specs: &[ColumnSpec],
```
wprzytula:
```suggestion
        col_specs: &[ColumnSpec<'static>],
```

scylla-rust-wrapper/src/query_result.rs
wprzytula: Yes, unfortunately. But soon it will be irrelevant, when we switch to new `QueryResult` with its `RowDeserializerOwned`.

wprzytula (changes requested):

scylla-rust-wrapper/src/query_result.rs
muzarski: Yeah... It's because current `QueryResult` API is in awkward state.

The main issue is that ColumnSpec lifetime is bound by lifetime of result:
```rust
    /// Returns column specifications.
    #[inline]
    pub fn col_specs(&self) -> &[ColumnSpec] {
        self.metadata
            .as_ref()
            .map(|metadata| metadata.col_specs())
            .unwrap_or_default()
    }
```

This is equivalent to:
```rust
pub fn col_specs<'s>(&'s self) -> &'s [ColumnSpec<'s>]
```
even though this version of `QueryResult` holds metadata consisting of `ColumnSpec<'static>` (owned column specs). To return `&[ColumnSpec<'static>]`, we also need fix introduced by this commit: https://github.com/scylladb/scylla-rust-driver/pull/1093/commits/e53a92c1721738dce4a6a4d257e5e0e48edaee21

Then we wouldn't need to call this `map` in cpp-rust-driver:
```rust
  .map(|col_spec| {
      // TODO: introduce ColumnSpec::[in]to_owned() methods in rust-driver.
      let name = col_spec.name().to_owned();
      let typ = col_spec.typ().clone().into_owned();
      let table_spec = col_spec.table_spec().to_owned();
      ColumnSpec::owned(name, typ, table_spec)
  })
```

I could open a small PR in rust-driver that does that (so we get rid of this mapping from borrowed col spec to owned col spec). This would be a temporary fix, before the final version of `QueryResult` is introduced in rust-driver.

As a side note: I believe that once final version of lazy deserialization is merged, the corresponding structures in cpp-rust-driver will need to be adjusted/refactored appropriately so we are as efficient as we can.

scylla-rust-wrapper/src/query_result.rs
muzarski: > Actually, there is a bug in `QueryResult`:
> 
> ```rust
> pub fn col_specs(&self) -> &[ColumnSpec] {
> ```
> 
> ...while it should be:
> 
> ```rust
> pub fn col_specs(&self) -> &[ColumnSpec<'static>] {
> ```

Yes, that's what I mentioned in response to @Lorak-mmk 's comment.

scylla-rust-wrapper/src/query_result.rs
muzarski: This also needs adjustment in `QueryResult::col_specs()`, so it returns `&[ColumnSpec<'static>]`

muzarski:
v1.1: In rust-driver, we changed an error message emitted when statement id changes after repreparation. I adjusted the corresponding intergration test (`PreparedTests.Integration_Cassandra_FailFastWhenPreparedIDChangesDuringReprepare`) to this new error message.

muzarski:
v2: Adjusted to https://github.com/scylladb/scylla-rust-driver/commit/da199a182e0ab95efd2caae495ad3583f09bdd85. Now there is no need to convert `ColumnSpec<'query_result>` to `ColumnSpec<'static>`

scylla-rust-wrapper/src/query_result.rs
muzarski: I reverted the change (from accepting `Vec<...>` to accepting slice), so now we accept `Vec<ColumnSpec<'static>>`

muzarski:
v2.1: Bumped again to https://github.com/scylladb/scylla-rust-driver/commit/64b4afcdb4286b21f6cc1acb55266d6607f250e0

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -422,7 +422,7 @@ impl CassDataType {
 
 pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
     match column_type {
-        ColumnType::Custom(s) => CassDataType::Custom((*s).clone()),
+        ColumnType::Custom(s) => CassDataType::Custom(s.clone().into_owned()),
         ColumnType::Ascii => CassDataType::Value(CassValueType::CASS_VALUE_TYPE_ASCII),
```
Lorak-mmk, line 426: `s` is a `&String`, right? So why do we need both `.clone()` and `.into_owned()`? Isn't one of those enough?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -459,10 +459,15 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
         } => CassDataType::UDT(UDTDataType {
             field_types: field_types
                 .iter()
-                .map(|(name, col_type)| ((*name).clone(), Arc::new(get_column_type(col_type))))
+                .map(|(name, col_type)| {
+                    (
+                        name.clone().into_owned(),
+                        Arc::new(get_column_type(col_type)),
+                    )
+                })
```
Lorak-mmk, line 467: ditto

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -459,10 +459,15 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
...
                 .collect(),
-            keyspace: (*keyspace).clone(),
-            name: (*type_name).clone(),
+            keyspace: keyspace.clone().into_owned(),
+            name: type_name.clone().into_owned(),
```
Lorak-mmk, line 470: ditto

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -376,13 +376,13 @@ pub unsafe extern "C" fn cass_execution_profile_set_retry_policy(
...
     retry_policy: *const CassRetryPolicy,
 ) -> CassError {
-    let retry_policy: &dyn RetryPolicy = match ptr_to_ref(retry_policy) {
-        DefaultRetryPolicy(default) => default.as_ref(),
-        FallthroughRetryPolicy(fallthrough) => fallthrough.as_ref(),
-        DowngradingConsistencyRetryPolicy(downgrading) => downgrading.as_ref(),
+    let retry_policy: Arc<dyn RetryPolicy> = match ptr_to_ref(retry_policy) {
+        DefaultRetryPolicy(default) => default.clone(),
+        FallthroughRetryPolicy(fallthrough) => fallthrough.clone(),
+        DowngradingConsistencyRetryPolicy(downgrading) => downgrading.clone(),
```
Lorak-mmk: Please use `Arc::clone` if you need to clone `Arc`. `.clone` is less obvious.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cass_types.rs
muzarski: `s` is `&Cow<'_, str>`. We need to `clone` the `Cow` and then convert it to `String`.

muzarski: `str::to_owned(s)` or `s.to_string()` seems to work as well. In case you prefer something shorter compared to current version. Notice that `s.to_owned()` returns an owned `Cow` (not `String`).

muzarski:
v2.2: replaced `.clone()` with `Arc::clone()` and rebased on master

scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk: > `s` is `&Cow<'_, str>`. We need to `clone` the `Cow` and then convert it to `String`.

I didn't realize that. In that case the current version is fine.

Lorak-mmk (approved):

wprzytula (approved):

PR #205, opened 2024-11-20
result: small cleanups before bump to 0.15

muzarski:
## Tracing id and paging state response
These were moved from `CassResultData` to `CassResult`.

### AlterProperlyUpdatesColumnCount test
This test should not be enabled. It expects that prepared statement's result metadata was updated after table schema was altered. It worked only because of a driver bug described later. I disabled this test. It's expected to be working for CQL protocol v5 and above. I think it can be addressed later.

### Driver bug
Before this PR, there was an inconsistency of metadata stored in `CassResultData`:
- `column_specs: Vec<ColumnSpec<'static>>` was obtained from a query result
- `col_data_types: Arc<Vec<Arc<CassDataType>>>` was obtained from the prepared statement's cached result metadata

Because of this, these vector could be of a different length when table schema was altered - query result provided the updated metadata, while prepared statement still held an outdated metadata.

Why was the aforementioned test working?
It's because `cass_result_column_count` would return the size of `column_specs` (i.e. updated metadata).

This PR fixes the issue by unifying these two vectors into a single one. `CassColumnSpec` was introduced to address it. It holds a column name (`String`) and a column data type (`Arc<CassDataType`).

### AlterDoesntUpdateColumnCount
On the other hand, this test should be enabled. It expects, that for protocol version 4 or less, the prepared statement's metadata is not updated after table schema was altered. It's enabled at the end of PR, once the bug mentioned above is fixed.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -23,6 +23,7 @@ pub struct CassResult {
     pub paging_state_response: PagingStateResponse,
 }
 
+#[derive(Debug)]
 pub struct CassResultData {
     pub col_specs: Vec<CassColumnSpec>,
 }
--- a/scylla-rust-wrapper/src/query_result.rs
+++ b/scylla-rust-wrapper/src/query_result.rs
@@ -1,7 +1,7 @@
 use crate::argconv::*;
 use crate::cass_error::CassError;
 use crate::cass_types::{
-    cass_data_type_type, get_column_type, CassDataType, CassValueType, MapDataType,
+    cass_data_type_type, get_column_type, CassColumnSpec, CassDataType, CassValueType, MapDataType,
 };
 use crate::inet::CassInet;
 use crate::metadata::{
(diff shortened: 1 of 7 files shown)


muzarski:
I found a bug in the driver. The fix will be part of this PR.

muzarski:
v2: addressed the bug and updated the cover letter. It's ready for review.

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk, line 1: Commit: "result: move tracing_id from CassResultData to CassResult "

Idea: You wrote "CassResultData should represent a result metadata". If that is the case then maybe it should be called "CassResultMetadata"?

Lorak-mmk (approved):
Looks very good, I really like the introduction of CassColumnSpec - it both makes the code easier to understand, and prevents the bug you encountered on a type level!

muzarski:
v2.1: Renamed `CassResultData` -> `CassResultMetadata`

Lorak-mmk (approved):

PR #207, opened 2024-11-25
type safety: ffi pointer manipulation API

muzarski:
## CassDataType interior mutability
`CassDataType` is the only type that is both shared, and mutable. We use of interior mutability pattern for this type. Even if the current code is sound in this matter, we should wrap the underlying data in `UnsafeCell`, to make (possibly)
invalid accesses to it more verbose and easier to detect by forcing the user to use an explicit `unsafe` block.

## [Box/Arc/Ref]FFI API
The main motivation behind this PR is to:

- reduce the lifetime of reference obtained from the pointer
Current pointer-to-ref conversion API (ptr_to_ref_[mut]) returns a
reference with a static lifetime. New API is parameterized by a lifetime,
which is not 'static.

- ensure that IF the memory associated with the pointer was allocated a certain way,
  the raw pointer will be converted to the corresponding Rust pointer primitive i.e. Box or Arc
  (or reference if the pointer was not obtained from an explicit allocation).

This is achieved, by (in addition, to the new API) disallowing the usage of `[Box/Arc]` API that operates on raw pointers.

However, this does not give us any guarantees about the origin of the pointer.
Consider following example:

// Rust
```rust
impl ArcFFI for Foo {}

fn extern "C" f1() -> *const Foo {
    // a pointer to stack variable
    // Also applies to some valid pointer obtained from the reference
    // to the field of some already heap-allocated object.
    // Decided to go with a stack variable to keep the example simple.
    let foo = Foo;
    &foo
}

fn extern "C" f2(foo: *const Foo) {
    let foo = ArcFFI::cloned_from_ptr(foo);
}
```

// C
```c
Foo *foo = f1();
f2(foo);

// Segfault.
// Even if f1() returned a valid pointer, that points to some
// heap-allocated memory. The pointer was not obtained from an Arc allocation.
// I.e., it was not obtained via Arc::into_raw().
```

To guarantee this, we need to introduce a special type for pointer that
would represent the pointer's properties. This will be done in a follow-up PR.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -97,3 +97,77 @@ macro_rules! make_c_str {
 
 #[cfg(test)]
 pub(crate) use make_c_str;
+
+/// Defines a pointer manipulation API for non-shared heap-allocated data.
+///
+/// Implement this trait for types that are allocated by the driver via [`Box::new`],
+/// and then returned to the user as a pointer. The user is responsible for freeing
+/// the memory associated with the pointer using corresponding driver's API function.
+pub trait BoxFFI {
+    fn into_ptr(self: Box<Self>) -> *mut Self {
+        Box::into_raw(self)
+    }
+    unsafe fn from_ptr(ptr: *mut Self) -> Box<Self> {
+        Box::from_raw(ptr)
+    }
+    unsafe fn as_maybe_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
+        ptr.as_ref()
+    }
+    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
+        ptr.as_ref().unwrap()
+    }
+    unsafe fn as_mut_ref<'a>(ptr: *mut Self) -> &'a mut Self {
+        ptr.as_mut().unwrap()
+    }
+    unsafe fn free(ptr: *mut Self) {
+        std::mem::drop(BoxFFI::from_ptr(ptr));
+    }
+}
+
+/// Defines a pointer manipulation API for shared heap-allocated data.
+///
+/// Implement this trait for types that require a shared ownership of data.
+/// The data should be allocated via [`Arc::new`], and then returned to the user as a pointer.
+/// The user is responsible for freeing the memory associated
+/// with the pointer using corresponding driver's API function.
+pub trait ArcFFI {
+    fn as_ptr(self: &Arc<Self>) -> *const Self {
+        Arc::as_ptr(self)
+    }
+    fn into_ptr(self: Arc<Self>) -> *const Self {
+        Arc::into_raw(self)
+    }
+    unsafe fn from_ptr(ptr: *const Self) -> Arc<Self> {
+        Arc::from_raw(ptr)
+    }
+    unsafe fn cloned_from_ptr(ptr: *const Self) -> Arc<Self> {
+        Arc::increment_strong_count(ptr);
+        Arc::from_raw(ptr)
+    }
+    unsafe fn as_maybe_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
+        ptr.as_ref()
+    }
+    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
+        ptr.as_ref().unwrap()
+    }
+    unsafe fn free(ptr: *const Self) {
+        std::mem::drop(ArcFFI::from_ptr(ptr));
+    }
+}
+
+/// Defines a pointer manipulation API for data owned by some other object.
+///
+/// Implement this trait for the types that do not need to be freed (directly) by the user.
+/// The lifetime of the data is bound to some other object owning it.
+///
+/// For example: lifetime of CassRow is bound by the lifetime of CassResult.
+/// There is no API function that frees the CassRow. It should be automatically
+/// freed when user calls cass_result_free.
+pub trait RefFFI {
+    fn as_ptr(&self) -> *const Self {
+        self as *const Self
+    }
+    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
+        ptr.as_ref().unwrap()
+    }
+}
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -4,33 +4,6 @@ use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::sync::Arc;
 
-pub unsafe fn ptr_to_ref<T>(ptr: *const T) -> &'static T {
-    ptr.as_ref().unwrap()
-}
-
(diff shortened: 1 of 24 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -226,17 +241,53 @@ impl CassDataType {
...
+
+/// PartialEq and Eq for test purposes.
+impl PartialEq for CassDataType {
+    fn eq(&self, other: &Self) -> bool {
+        unsafe { self.get_unchecked() == other.get_unchecked() }
+    }
+}
+impl Eq for CassDataType {}
+
+unsafe impl Sync for CassDataType {}
```
wprzytula, line 277: Why is this needed? AFAIK, raw pointers do not implement Send nor Sync; for the same reason, I think `CassDataType` also should not, unless it's required due to some other important reasons.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -226,17 +241,53 @@ impl CassDataType {
...
+
+impl CassDataType {
+    pub unsafe fn get_unchecked(&self) -> &CassDataTypeInner {
+        &*self.0.get()
+    }
+
+    #[allow(clippy::mut_from_ref)]
+    pub unsafe fn get_mut_unchecked(&self) -> &mut CassDataTypeInner {
+        &mut *self.0.get()
+    }
```
wprzytula, line 287: For a moment, I was afraid that this is unsound. We're creating a `&mut` that may alias the same memory as some other accesses!
Tbh, I'm in two minds if this is sound or not.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -226,17 +241,53 @@ impl CassDataType {
...
+
+    pub const fn new(inner: CassDataTypeInner) -> CassDataType {
+        CassDataType(UnsafeCell::new(inner))
+    }
+
+    pub fn new_alloc(inner: CassDataTypeInner) -> Arc<CassDataType> {
+        Arc::new(CassDataType(UnsafeCell::new(inner)))
+    }
```
wprzytula: Nit: I'd prefer `new_arced` as the name, as it seems more idiomatic to me.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +97,77 @@ macro_rules! make_c_str {
...
+pub trait BoxFFI {
+    fn into_ptr(self: Box<Self>) -> *mut Self {
+        Box::into_raw(self)
+    }
+    unsafe fn from_ptr(ptr: *mut Self) -> Box<Self> {
+        Box::from_raw(ptr)
+    }
+    unsafe fn maybe_as_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
+        ptr.as_ref()
+    }
```
wprzytula: Nit: `as_maybe_ref` makes more sense to me. `as` is the name of the operation, and `maybe_ref` is the target of the conversion.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -4,32 +4,32 @@ use std::ffi::CStr;
...
+//     Arc::increment_strong_count(ptr);
+//     Arc::from_raw(ptr)
+// }
+
+// pub unsafe fn free_arced<T>(ptr: *const T) {
+//     if !ptr.is_null() {
+//         // This decrements the arc's internal counter and potentially drops it
+//         Arc::from_raw(ptr);
+//     }
+// }
```
wprzytula: Nit: let's delete those lines instead of commenting them out in this commit.

wprzytula (changes requested):
In general, looks good.

scylla-rust-wrapper/src/cass_types.rs
muzarski: Before that, we would create a `&mut` directly from a pointer passed to the function. With new approach, we are more explicit, and the user needs to wrap the call with an `unsafe` block. Is it sound? Well, it's the caller's responsibility to ensure the soundness.

muzarski:
Rebased on master

scylla-rust-wrapper/src/cass_types.rs
muzarski: There are multiple places in the code where the compiler complains when I comment it out. One of them is the place where we define types with a static lifetime - an optimization for untyped collections/tuples. There are also some types that are indirect owners of `Arc<CassDataType>`, and are passed to an async block which is required to be `Send`.

muzarski:
v2: addressed @wprzytula comments

wprzytula (approved):

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -584,25 +640,25 @@ pub unsafe extern "C" fn cass_data_type_type_name(
...
 ) -> CassError {
     cass_data_type_set_type_name_n(data_type, type_name, strlen(type_name))
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_data_type_set_type_name_n(
-    data_type_raw: *mut CassDataType,
+    data_type_raw: *const CassDataType,
     type_name: *const c_char,
     type_name_length: size_t,
```
Lorak-mmk, line 659: ❓ Can you remind me / cite some source on what are the semantics of `*mut` vs `*const`? `CassDataType* data_type` is used on C side, which is not a const object, right? If that is the case, why do we use `*const` here?

Lorak-mmk: This applied to other places where `*mut` was changed to `*const`.

wprzytula: The semantics of `*const` means that no one is going to modify the pointee through the pointer. Here, this is **not** the case.

Lorak-mmk: What does modify mean? Let's say we get a `*const Mutex<something>`. Are we allowed to lock the mutex and modify its contents? It only requires creating `&` from `*const`.

wprzytula: _Modify_ on structs is defined by induction, by modifying one of its fields. And _modyfing_ a value of a primitive type involves changing bits of their memory representation.

However, _interior mutability_ is what makes an exception: if a type allows for _internal mutability_ (in Rust, it basically means that it holds an `UnsafeCell`), then mutation inside that type are not considered mutation from outside. So answering your question, we are allowed to lock the Mutex through `*const Mutex`.

muzarski: I believe that Wojciech answered your question. Notice, that all these `*const` and `*mut` will not matter after a follow-up PR, since raw pointers are going to be replaced by our new type representing a pointer.

Lorak-mmk (approved):

PR #208, opened 2024-11-26
safety: introduce pointer types and their restrictions

muzarski:
This is a follow-up to: https://github.com/scylladb/cpp-rust-driver/pull/207.

It defines a `CassPtr` type, wrapper over `Option<NonNull<T>>`. The pointer is parameterized by:
## Lifetime
Thanks to that, we can distinguish between owned (should be freed) and borrowed pointers.

## Properties
There are currently two properties:
1. `Ownership` - it tells whether the pointer is mutable from Rust perspective. In other words, `Exclusive` = mutable pointer, `Shared` = immutable pointer.
2. `CMutability` - the semantics of this one are very simple. It just tells whether the pointer is `T*` or `const T*` in C API definition. It's introduced so we can distinguish **shared**, **logically mutable** types (`Mutability::Const` and `CMutability::CMut`) from **shared**, **logically immutable** types (`Mutability::Const` and `CMutability::CConst`). This property will help to write safe unit tests where we play a role of C API user.

During review, please pay attention to that if some pointer was originally `const T*` (or `*const T` in Rust), then it is `CConst` after this PR. Analogously, if it was `T*` (or `*mut T` in Rust), then it should be `CMut`.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -1,7 +1,9 @@
 use crate::types::size_t;
 use std::cmp::min;
 use std::ffi::CStr;
+use std::marker::PhantomData;
 use std::os::raw::c_char;
+use std::ptr::NonNull;
 use std::sync::Arc;
 
 pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
@@ -70,6 +72,259 @@ macro_rules! make_c_str {
 
 #[cfg(test)]
 pub(crate) use make_c_str;
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A trait representing ownership (i.e. Rust mutability) of the pointer.
+///
+/// Pointer can either be [`Exclusive`] or [`Shared`].
+///
+/// ## Shared pointers
+/// Shared pointers can only be converted to **immutable** Rust referential types.
+/// There is no way to obtain a mutable reference from such pointer.
+///
+/// In some cases, we need to be able to mutate the data behind a shared pointer.
+/// There is an example of such use case - namely [`crate::cass_types::CassDataType`].
+/// argconv API does not provide a way to mutate such pointer - one can only convert the pointer
+/// to [`Arc`] or &. It is the API user's responsibility to implement sound interior mutability
+/// pattern in such case. This is what we currently do - CassDataType wraps CassDataTypeInner
+/// inside an `UnsafeCell` to implement interior mutability for the type.
+/// Other example is [`crate::future::CassFuture`] which uses Mutex.
+///
+/// ## Exclusive pointers
+/// Exclusive pointers can be converted to both immutable and mutable Rust referential types.
+pub trait Ownership: sealed::Sealed {}
+
+/// Represents shared (immutable) pointer.
+pub struct Shared;
+impl sealed::Sealed for Shared {}
+impl Ownership for Shared {}
+
+/// Represents exclusive (mutable) pointer.
+pub struct Exclusive;
+impl sealed::Sealed for Exclusive {}
+impl Ownership for Exclusive {}
+
+/// A trait representing mutability of the pointer in C semantics.
+///
+/// Pointer can be either [`CConst`] or [`CMut`].
+///
+/// The semantics of this property are very simple: it tells whether
+/// the pointer is const or not on the C side. If the pointer is `T*`, then
+/// corresponding [`CassPtr`] is [`CMut`], and if the pointer is `const T*`, then
+/// corresponding [`CassPtr`] is [`CConst`].
+///
+/// ## Why [`Ownership`] property is not enough?
+/// This is because mutability semantics differ between C and Rust.
+/// - In C/C++: mut = logically mutable, const = logically immutable
+/// - In Rust: mut = exclusive ownership (e.g. Box, &mut), const = shared ownership (e.g. Arc, &)
+///
+/// In cpp-driver API there are some types that are both shared and mutable. Take for example `CassDataType`.
+/// From rust perspective, the pointer is always [`Shared`], because it is shared behind an [`Arc`].
+/// But there are some methods, e.g:
+/// ```rust,ignore
+/// cass_data_type_add_sub_type(CassDataType* data_type, const CassDataType* sub_data_type);
+/// ```
+/// which accept a mutable pointer. Such method is implemented by employing interior mutability pattern (mutating a shared object).
+/// If we did not have [`CMutability`] property defined, we would not be able to
+/// distinguish between `CassDataType*` and `const CassDataType*`. In such case, user would be able to obtain
+/// `const CassDataType*` pointer (e.g. from `cass_tuple_data_type`) and pass it to the method that expects `CassDataType*`.
+///
+/// This property comes useful for implementing safe Rust unit tests where we play a role of C API user.
+pub trait CMutability: sealed::Sealed {}
+
+/// Represents const C pointer.
+pub struct CConst;
+impl sealed::Sealed for CConst {}
+impl CMutability for CConst {}
+
+/// Represents mutable (non-const) C pointer.
+pub struct CMut;
+impl sealed::Sealed for CMut {}
+impl CMutability for CMut {}
+
+/// Represents additional properties of the pointer.
+pub trait Properties: sealed::Sealed {
+    type Onwership: Ownership;
+    type CMutability: CMutability;
+}
+
+impl<O: Ownership, CM: CMutability> sealed::Sealed for (O, CM) {}
+impl<O: Ownership, CM: CMutability> Properties for (O, CM) {
+    type Onwership = O;
+    type CMutability = CM;
+}
+
+/// Represents a valid non-dangling pointer.
+///
+/// ## Safety and validity guarantees
+/// Apart from trivial constructors such as [`CassPtr::null()`] and [`CassPtr::null_mut()`], there
+/// is only one way to construct a [`CassPtr`] instance - from raw pointer via [`CassPtr::from_raw()`].
+/// This constructor is `unsafe`. It is user's responsibility to ensure that the raw pointer
+/// provided to the constructor is **valid**. In other words, the pointer comes from some valid
+/// allocation, or from some valid reference.
+///
+/// ## Generic lifetime and aliasing guarantees
+/// We distinguish two types of pointers: shared ([`Shared`]) and exclusive ([`Exclusive`]).
+/// Shared pointers can be converted to immutable (&) references, while exclusive pointers
+/// can be converted to either immutable (&) or mutable (&mut) reference. User needs to pick
+/// the correct mutability property of the pointer during construction. This is yet another
+/// reason why [`CassPtr::from_raw`] is `unsafe`.
+///
+/// Pointer is parameterized by the lifetime. Thanks to that, we can tell whether the pointer
+/// **owns** or **borrows** the pointee. Once again, user is responsible for "picking"
+/// the correct lifetime when creating the pointer. For example, when raw pointer
+/// comes from [`Box::into_raw()`], user could create a [`CassPtr<'static, T, (Exclusive,)>`].
+/// `'static` lifetime represents that user is the exclusive **owner** of the pointee, and
+/// is responsible for freeing the memory (e.g. via [`Box::from_raw()`]).
+/// On the other hand, when pointer is created from some immutable reference `&'a T`,
+/// the correct choice of CassPtr would be [`CassPtr<'a, T, (Shared,)>`]. It means that
+/// holder of the created pointer **borrows** the pointee (with some lifetime `'a`
+/// inherited from the immutable borrow `&'a T`).
+///
+/// Both [`CassPtr::into_ref()`] and [`CassPtr::into_mut_ref()`] consume the pointer.
+/// At first glance, it seems impossible to obtain multiple immutable reference from one pointer.
+/// This is why pointer reborrowing mechanism is introduced. There are two methods: [`CassPtr::borrow()`]
+/// and [`CassPtr::borrow_mut()`]. Both of them cooperate with borrow checker and enforce
+/// aliasing XOR mutability principle at compile time.
+///
+/// ## Safe conversions to referential types
+/// Thanks to the above guarantees, conversions to referential types are **safe**.
+/// See methods [`CassPtr::into_ref()`] and [`CassPtr::into_mut_ref()`].
+///
+/// ## Memory layout
+/// We use repr(transparent), so the struct has the same layout as underlying [`Option<NonNull<T>>`].
+/// Thanks to https://doc.rust-lang.org/std/option/#representation optimization,
+/// we are guaranteed, that for `T: Sized`, our struct has the same layout
+/// and function call ABI as simply [`NonNull<T>`].
+#[repr(transparent)]
+pub struct CassPtr<'a, T: Sized, P: Properties> {
+    ptr: Option<NonNull<T>>,
+    _phantom: PhantomData<&'a P>,
+}
+
+/// Owned shared pointer.
+/// Can be used for pointers with shared ownership - e.g. pointers coming from [`Arc`] allocation.
+pub type CassOwnedSharedPtr<T, CM> = CassPtr<'static, T, (Shared, CM)>;
+
+/// Borrowed shared pointer.
+/// Can be used for pointers created from some immutable reference.
+pub type CassBorrowedSharedPtr<'a, T, CM> = CassPtr<'a, T, (Shared, CM)>;
+
+/// Owned exclusive pointer.
+/// Can be used for pointers with exclusive ownership - e.g. pointers coming from [`Box`] allocation.
+pub type CassOwnedExclusivePtr<T, CM> = CassPtr<'static, T, (Exclusive, CM)>;
+
+/// Borrowed exclusive pointer.
+/// This can be for example obtained from mutable reborrow of some [`CassOwnedExclusivePtr`].
+pub type CassBorrowedExclusivePtr<'a, T, CM> = CassPtr<'a, T, (Exclusive, CM)>;
+
+/// Utility method for tests. Useful when some method returns `T*`,
+/// and then another method accepts `const T*`.
+#[cfg(test)]
+impl<'a, T: Sized, P: Properties> CassPtr<'a, T, P> {
+    pub fn into_c_const(self) -> CassPtr<'a, T, (P::Onwership, CConst)> {
+        CassPtr {
+            ptr: self.ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+/// Pointer constructors.
+impl<T: Sized, P: Properties> CassPtr<'_, T, P> {
+    fn null() -> Self {
+        CassPtr {
+            ptr: None,
+            _phantom: PhantomData,
+        }
+    }
+
+    fn is_null(&self) -> bool {
+        self.ptr.is_none()
+    }
+
+    /// Constructs [`CassPtr`] from raw pointer.
+    ///
+    /// ## Safety
+    /// User needs to ensure that the pointer is **valid**.
+    /// User is also responsible for picking correct ownership property and lifetime
+    /// of the created pointer. For more information, see the documentation of [`CassPtr`].
+    unsafe fn from_raw(raw: *const T) -> Self {
+        CassPtr {
+            ptr: NonNull::new(raw as *mut T),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+/// Conversion to raw pointer.
+impl<T: Sized, P: Properties> CassPtr<'_, T, P> {
+    fn to_raw(&self) -> Option<*mut T> {
+        self.ptr.map(|ptr| ptr.as_ptr())
+    }
+}
+
+/// Constructors for to exclusive pointers.
+impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
+    fn null_mut() -> Self {
+        CassPtr {
+            ptr: None,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<'a, T: Sized, P: Properties> CassPtr<'a, T, P> {
+    /// Converts a pointer to an optional valid reference.
+    /// The reference inherits the lifetime of the pointer.
+    fn into_ref(self) -> Option<&'a T> {
+        // SAFETY: Thanks to the validity and aliasing ^ mutability guarantees,
+        // we can safely convert the pointer to valid immutable reference with
+        // correct lifetime.
+        unsafe { self.ptr.map(|p| p.as_ref()) }
+    }
+}
+
+impl<'a, T: Sized, CM: CMutability> CassPtr<'a, T, (Exclusive, CM)> {
+    /// Converts a pointer to an optional valid mutable reference.
+    /// The reference inherits the lifetime of the pointer.
+    fn into_mut_ref(self) -> Option<&'a mut T> {
+        // SAFETY: Thanks to the validity and aliasing ^ mutability guarantees,
+        // we can safely convert the pointer to valid mutable (and exclusive) reference with
+        // correct lifetime.
+        unsafe { self.ptr.map(|mut p| p.as_mut()) }
+    }
+}
+
+impl<T: Sized, P: Properties> CassPtr<'_, T, P> {
+    /// Immutably reborrows the pointer.
+    /// Resulting pointer inherits the lifetime from the immutable borrow
+    /// of original pointer.
+    #[allow(clippy::needless_lifetimes)]
+    pub fn borrow<'a>(&'a self) -> CassPtr<'a, T, (Shared, P::CMutability)> {
+        CassPtr {
+            ptr: self.ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
+    /// Mutably reborrows the pointer.
+    /// Resulting pointer inherits the lifetime from the mutable borrow
+    /// of original pointer. Since the method accepts a mutable reference
+    /// to the original pointer, we enforce aliasing ^ mutability principle at compile time.
+    #[allow(clippy::needless_lifetimes)]
+    pub fn borrow_mut<'a>(&'a mut self) -> CassPtr<'a, T, (Exclusive, CMut)> {
+        CassPtr {
+            ptr: self.ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
 
 /// Defines a pointer manipulation API for non-shared heap-allocated data.
 ///
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -4,7 +4,7 @@ use std::ffi::CStr;
 use std::marker::PhantomData;
 use std::os::raw::c_char;
 use std::ptr::NonNull;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 
 pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
     CStr::from_ptr(ptr).to_str().ok()
@@ -331,29 +331,71 @@ impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
 /// Implement this trait for types that are allocated by the driver via [`Box::new`],
 /// and then returned to the user as a pointer. The user is responsible for freeing
 /// the memory associated with the pointer using corresponding driver's API function.
-pub trait BoxFFI {
-    fn into_ptr(self: Box<Self>) -> *mut Self {
+pub trait BoxFFI: Sized {
+    /// Consumes the Box and returns a pointer with exclusive ownership.
+    /// The pointer needs to be freed. See [`BoxFFI::free()`].
+    fn into_ptr<CM: CMutability>(self: Box<Self>) -> CassPtr<'static, Self, (Exclusive, CM)> {
         #[allow(clippy::disallowed_methods)]
-        Box::into_raw(self)
+        let ptr = Box::into_raw(self);
+
+        // SAFETY:
+        // 1. validity guarantee - pointer is obviously valid. It comes from box allocation.
+        // 2. pointer's lifetime - we choose 'static lifetime. It is ok, because holder of the
+        //    pointer becomes the owner of pointee. He is responsible for freeing the memory
+        //    via BoxFFI::free() - which accepts 'static pointer. User is not able to obtain
+        //    another pointer with 'static lifetime pointing to the same memory.
+        // 3. ownership - user becomes an exclusive owner of the pointee. Thus, it's ok
+        //    for the pointer to be `Exclusive`.
+        unsafe { CassPtr::from_raw(ptr) }
     }
-    unsafe fn from_ptr(ptr: *mut Self) -> Box<Self> {
-        #[allow(clippy::disallowed_methods)]
-        Box::from_raw(ptr)
+
+    /// Consumes the pointer with exclusive ownership back to the Box.
+    fn from_ptr<CM: CMutability>(
+        ptr: CassPtr<'static, Self, (Exclusive, CM)>,
+    ) -> Option<Box<Self>> {
+        // SAFETY:
+        // The only way to obtain an owned pointer (with 'static lifetime) is BoxFFI::into_ptr().
+        // It creates a pointer based on Box allocation. It is thus safe to convert the pointer
+        // back to owned `Box`.
+        unsafe {
+            ptr.to_raw().map(|p| {
+                #[allow(clippy::disallowed_methods)]
+                Box::from_raw(p)
+            })
+        }
     }
-    unsafe fn as_maybe_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref()
+
+    /// Creates a reference from an exclusive pointer.
+    /// Reference inherits the lifetime of the pointer's borrow.
+    #[allow(clippy::needless_lifetimes)]
+    fn as_ref<'a, O: Ownership, CM: CMutability>(
+        ptr: CassPtr<'a, Self, (O, CM)>,
+    ) -> Option<&'a Self> {
+        ptr.into_ref()
     }
-    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref().unwrap()
+
+    /// Creates a mutable from an exlusive pointer.
+    /// Reference inherits the lifetime of the pointer's mutable borrow.
+    #[allow(clippy::needless_lifetimes)]
+    fn as_mut_ref<'a, CM: CMutability>(
+        ptr: CassPtr<'a, Self, (Exclusive, CM)>,
+    ) -> Option<&'a mut Self> {
+        ptr.into_mut_ref()
     }
-    unsafe fn as_mut_ref<'a>(ptr: *mut Self) -> &'a mut Self {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_mut().unwrap()
-    }
-    unsafe fn free(ptr: *mut Self) {
+
+    /// Frees the pointee.
+    fn free<CM: CMutability>(ptr: CassPtr<'static, Self, (Exclusive, CM)>) {
         std::mem::drop(BoxFFI::from_ptr(ptr));
+    }
+
+    // Currently used only in tests.
+    #[allow(dead_code)]
+    fn null<'a, CM: CMutability>() -> CassPtr<'a, Self, (Shared, CConst)> {
+        CassPtr::null()
+    }
+
+    fn null_mut<'a>() -> CassPtr<'a, Self, (Exclusive, CMut)> {
+        CassPtr::null_mut()
     }
 }
 
@@ -363,35 +405,85 @@ pub trait BoxFFI {
 /// The data should be allocated via [`Arc::new`], and then returned to the user as a pointer.
 /// The user is responsible for freeing the memory associated
 /// with the pointer using corresponding driver's API function.
-pub trait ArcFFI {
-    fn as_ptr(self: &Arc<Self>) -> *const Self {
+pub trait ArcFFI: Sized {
+    /// Creates a pointer from a valid reference to Arc-allocated data.
+    /// Holder of the pointer borrows the pointee.
+    fn as_ptr<'a, CM: CMutability>(self: &'a Arc<Self>) -> CassPtr<'a, Self, (Shared, CM)> {
         #[allow(clippy::disallowed_methods)]
-        Arc::as_ptr(self)
+        let ptr = Arc::as_ptr(self);
+
+        // SAFETY:
+        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation.
+        // 2. pointer's lifetime - pointer inherits the lifetime of provided Arc's borrow.
+        //    What's important is that the returned pointer borrows the data, and is not the
+        //    shared owner. Thus, user cannot call ArcFFI::free() on such pointer.
+        // 3. ownership - we always create a `Shared` pointer.
+        unsafe { CassPtr::from_raw(ptr) }
     }
-    fn into_ptr(self: Arc<Self>) -> *const Self {
+
+    /// Creates a pointer from a valid Arc allocation.
+    fn into_ptr<CM: CMutability>(self: Arc<Self>) -> CassPtr<'static, Self, (Shared, CM)> {
         #[allow(clippy::disallowed_methods)]
-        Arc::into_raw(self)
+        let ptr = Arc::into_raw(self);
+
+        // SAFETY:
+        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation
+        // 2. pointer's lifetime - returned pointer has a 'static lifetime. It is a shared
+        //    owner of the pointee. User has to decrement the RC of the pointer (and potentially free the memory)
+        //    via ArcFFI::free().
+        // 3. ownership - we always create a `Shared` pointer.
+        unsafe { CassPtr::from_raw(ptr) }
     }
-    unsafe fn from_ptr(ptr: *const Self) -> Arc<Self> {
-        #[allow(clippy::disallowed_methods)]
-        Arc::from_raw(ptr)
+
+    /// Converts shared owned pointer back to owned Arc.
+    fn from_ptr<CM: CMutability>(ptr: CassPtr<'static, Self, (Shared, CM)>) -> Option<Arc<Self>> {
+        // SAFETY:
+        // The only way to obtain a pointer with shared ownership ('static lifetime) is
+        // ArcFFI::into_ptr(). It converts an owned Arc into the pointer. It is thus safe
+        // to convert such pointer back to owned Arc.
+        unsafe {
+            ptr.to_raw().map(|p| {
+                #[allow(clippy::disallowed_methods)]
+                Arc::from_raw(p)
+            })
+        }
     }
-    unsafe fn cloned_from_ptr(ptr: *const Self) -> Arc<Self> {
-        #[allow(clippy::disallowed_methods)]
-        Arc::increment_strong_count(ptr);
-        #[allow(clippy::disallowed_methods)]
-        Arc::from_raw(ptr)
+
+    /// Increases the reference count of the pointer, and returns an owned Arc.
+    fn cloned_from_ptr<CM: CMutability>(ptr: CassPtr<'_, Self, (Shared, CM)>) -> Option<Arc<Self>> {
+        // SAFETY:
+        // All pointers created via ArcFFI API are originated from Arc allocation.
+        // It is thus safe to increase the reference count of the pointer, and convert
+        // it to Arc. Because of the borrow-checker, it is not possible for the user
+        // to provide a pointer that points to already deallocated memory.
+        unsafe {
+            ptr.to_raw().map(|p| {
+                #[allow(clippy::disallowed_methods)]
+                Arc::increment_strong_count(p);
+                #[allow(clippy::disallowed_methods)]
+                Arc::from_raw(p)
+            })
+        }
     }
-    unsafe fn as_maybe_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref()
+
+    /// Converts a shared borrowed pointer to reference.
+    /// The reference inherits the lifetime of pointer's borrow.
+    #[allow(clippy::needless_lifetimes)]
+    fn as_ref<'a, CM: CMutability>(ptr: CassPtr<'a, Self, (Shared, CM)>) -> Option<&'a Self> {
+        ptr.into_ref()
     }
-    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref().unwrap()
-    }
-    unsafe fn free(ptr: *const Self) {
+
+    /// Decreases the reference count (and potentially frees) of the owned pointer.
+    fn free<CM: CMutability>(ptr: CassPtr<'static, Self, (Shared, CM)>) {
         std::mem::drop(ArcFFI::from_ptr(ptr));
+    }
+
+    fn null<'a, CM: CMutability>() -> CassPtr<'a, Self, (Shared, CM)> {
+        CassPtr::null()
+    }
+
+    fn is_null<CM: CMutability>(ptr: &CassPtr<'_, Self, (Shared, CM)>) -> bool {
+        ptr.is_null()
     }
 }
 
@@ -403,12 +495,134 @@ pub trait ArcFFI {
 /// For example: lifetime of CassRow is bound by the lifetime of CassResult.
 /// There is no API function that frees the CassRow. It should be automatically
 /// freed when user calls cass_result_free.
-pub trait RefFFI {
-    fn as_ptr(&self) -> *const Self {
-        self as *const Self
+pub trait RefFFI: Sized {
+    /// Creates a borrowed pointer from a valid reference.
+    #[allow(clippy::needless_lifetimes)]
+    fn as_ptr<'a, CM: CMutability>(&'a self) -> CassPtr<'a, Self, (Shared, CM)> {
+        // SAFETY:
+        // 1. validity guarantee - pointer is valid, since it's obtained a valid reference.
+        // 2. pointer's lifetime - pointer inherits the lifetime of provided reference's borrow.
+        // 3. ownerhsip - we always create a `Shared` pointer.
+        unsafe { CassPtr::from_raw(self) }
     }
-    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref().unwrap()
+
+    /// Creates a borrowed pointer from a weak reference.
+    ///
+    /// ## SAFETY
+    /// User needs to ensure that the pointee is not freed when pointer is being
+    /// dereferenced.
+    ///
+    /// ## Why this method is unsafe? - Example
+    /// ```
+    /// # use scylla_cpp_driver::argconv::{RefFFI, CConst, CassBorrowedSharedPtr};
+    /// # use std::sync::{Arc, Weak};
+    ///
+    /// struct Foo;
+    /// impl RefFFI for Foo {}
+    ///
+    /// let arc = Arc::new(Foo);
+    /// let weak = Arc::downgrade(&arc);
+    /// let ptr: CassBorrowedSharedPtr<Foo, CConst> = unsafe { RefFFI::weak_as_ptr(&weak) };
+    /// std::mem::drop(arc);
+    ///
+    /// // The ptr is now dangling. The user can "safely" dereference it using RefFFI API.
+    ///
+    /// ```
+    #[allow(clippy::needless_lifetimes)]
+    unsafe fn weak_as_ptr<'a, CM: CMutability>(
+        w: &'a Weak<Self>,
+    ) -> CassPtr<'a, Self, (Shared, CM)> {
+        match w.upgrade() {
+            Some(a) => {
+                #[allow(clippy::disallowed_methods)]
+                let ptr = Arc::as_ptr(&a);
+                unsafe { CassPtr::from_raw(ptr) }
+            }
+            None => CassPtr::null(),
+        }
+    }
+
+    /// Converts a borrowed pointer to reference.
+    /// The reference inherits the lifetime of pointer's borrow.
+    #[allow(clippy::needless_lifetimes)]
+    fn as_ref<'a, CM: CMutability>(ptr: CassPtr<'a, Self, (Shared, CM)>) -> Option<&'a Self> {
+        ptr.into_ref()
+    }
+
+    fn null<'a, CM: CMutability>() -> CassPtr<'a, Self, (Shared, CM)> {
+        CassPtr::null()
+    }
+
+    fn is_null<CM: CMutability>(ptr: &CassPtr<'_, Self, (Shared, CM)>) -> bool {
+        ptr.is_null()
     }
 }
+
+/// ```compile_fail,E0499
+/// # use scylla_cpp_driver::argconv::{CassOwnedExclusivePtr, CassBorrowedExclusivePtr, CMut};
+/// # use scylla_cpp_driver::argconv::BoxFFI;
+/// struct Foo;
+/// impl BoxFFI for Foo {}
+///
+/// let mut ptr: CassOwnedExclusivePtr<Foo, CMut> = BoxFFI::into_ptr(Box::new(Foo));
+/// let borrowed_mut_ptr1: CassBorrowedExclusivePtr<Foo, CMut> = ptr.borrow_mut();
+/// let borrowed_mut_ptr2: CassBorrowedExclusivePtr<Foo, CMut> = ptr.borrow_mut();
+/// let mutref1 = BoxFFI::as_mut_ref(borrowed_mut_ptr2);
+/// let mutref2 = BoxFFI::as_mut_ref(borrowed_mut_ptr1);
+/// ```
+fn _test_box_ffi_cannot_have_two_mutable_references() {}
+
+/// ```compile_fail,E0502
+/// # use scylla_cpp_driver::argconv::{CassOwnedExclusivePtr, CassBorrowedSharedPtr, CassBorrowedExclusivePtr, CConst, CMut};
+/// # use scylla_cpp_driver::argconv::BoxFFI;
+/// struct Foo;
+/// impl BoxFFI for Foo {}
+///
+/// let mut ptr: CassOwnedExclusivePtr<Foo, CMut> = BoxFFI::into_ptr(Box::new(Foo));
+/// let borrowed_mut_ptr: CassBorrowedExclusivePtr<Foo, CMut> = ptr.borrow_mut();
+/// let borrowed_ptr: CassBorrowedSharedPtr<Foo, CMut> = ptr.borrow();
+/// let immref = BoxFFI::as_ref(borrowed_ptr);
+/// let mutref = BoxFFI::as_mut_ref(borrowed_mut_ptr);
+/// ```
+fn _test_box_ffi_cannot_have_mutable_and_immutable_references_at_the_same_time() {}
+
+/// ```compile_fail,E0505
+/// # use scylla_cpp_driver::argconv::{CassOwnedExclusivePtr, CassBorrowedSharedPtr, CMut};
+/// # use scylla_cpp_driver::argconv::BoxFFI;
+/// struct Foo;
+/// impl BoxFFI for Foo {}
+///
+/// let ptr: CassOwnedExclusivePtr<Foo, CMut> = BoxFFI::into_ptr(Box::new(Foo));
+/// let borrowed_ptr: CassBorrowedSharedPtr<Foo, CMut> = ptr.borrow();
+/// BoxFFI::free(ptr);
+/// let immref = BoxFFI::as_ref(borrowed_ptr);
+/// ```
+fn _test_box_ffi_cannot_free_while_having_borrowed_pointer() {}
+
+/// ```compile_fail,E0505
+/// # use scylla_cpp_driver::argconv::{CassOwnedSharedPtr, CassBorrowedSharedPtr, CConst};
+/// # use scylla_cpp_driver::argconv::ArcFFI;
+/// # use std::sync::Arc;
+/// struct Foo;
+/// impl ArcFFI for Foo {}
+///
+/// let ptr: CassOwnedSharedPtr<Foo, CConst> = ArcFFI::into_ptr(Arc::new(Foo));
+/// let borrowed_ptr: CassBorrowedSharedPtr<Foo, CConst> = ptr.borrow();
+/// ArcFFI::free(ptr);
+/// let immref = ArcFFI::cloned_from_ptr(borrowed_ptr);
+/// ```
+fn _test_arc_ffi_cannot_clone_after_free() {}
+
+/// ```compile_fail,E0505
+/// # use scylla_cpp_driver::argconv::{CassBorrowedSharedPtr, CConst};
+/// # use scylla_cpp_driver::argconv::ArcFFI;
+/// # use std::sync::Arc;
+/// struct Foo;
+/// impl ArcFFI for Foo {}
+///
+/// let arc = Arc::new(Foo);
+/// let borrowed_ptr: CassBorrowedSharedPtr<Foo, CConst> = ArcFFI::as_ptr(&arc);
+/// std::mem::drop(arc);
+/// let immref = ArcFFI::cloned_from_ptr(borrowed_ptr);
+/// ```
+fn _test_arc_ffi_cannot_dereference_borrowed_after_drop() {}
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -326,12 +326,18 @@ impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
     }
 }
 
+mod origin_sealed {
+    pub trait FromBoxSealed {}
+    pub trait FromArcSealed {}
+    pub trait FromRefSealed {}
+}
+
 /// Defines a pointer manipulation API for non-shared heap-allocated data.
 ///
 /// Implement this trait for types that are allocated by the driver via [`Box::new`],
 /// and then returned to the user as a pointer. The user is responsible for freeing
 /// the memory associated with the pointer using corresponding driver's API function.
-pub trait BoxFFI: Sized {
+pub trait BoxFFI: Sized + origin_sealed::FromBoxSealed {
     /// Consumes the Box and returns a pointer with exclusive ownership.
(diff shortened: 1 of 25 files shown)


scylla-rust-wrapper/src/collection.rs
```diff
@@ -4,22 +4,28 @@ use crate::cass_types::{CassDataType, CassDataTypeInner, MapDataType};
...
-});
-static UNTYPED_SET_TYPE: CassDataType = CassDataType::new(CassDataTypeInner::Set {
-    typ: None,
-    frozen: false,
-});
-static UNTYPED_MAP_TYPE: CassDataType = CassDataType::new(CassDataTypeInner::Map {
-    typ: MapDataType::Untyped,
-    frozen: false,
-});
+lazy_static! {
```
wprzytula: I've seen that [LazyLock](https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html) is already stable. Maybe it's better to transition to it instead of an external `lazy_static` dependency?

scylla-rust-wrapper/src/collection.rs
```diff
@@ -498,4 +508,26 @@ mod tests {
...
+            let empty_list_dt = cass_collection_data_type(empty_list);
+
+            let empty_set_dt = cass_data_type_new(CassValueType::CASS_VALUE_TYPE_SET);
+            // This will try to increment the reference count of `empty_list_dt`.
+            // Previously, this would fail, because `empty_list_dt` did not originate from an Arc allocation.
+            cass_data_type_add_sub_type(empty_set_dt, empty_list_dt);
+
+            // Cast to *mut, because `cass_data_type_new` returns a *const. See the comment
+            // in this function to see why.
+            cass_data_type_free(empty_set_dt as *mut _)
```
wprzytula: I think this mentioned comment in `cass_data_type_new` is not yet there in this commit (`collection: allocate static empty collections`).

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -433,7 +433,11 @@ pub unsafe extern "C" fn cass_materialized_view_meta_base_table(
     view_meta: *const CassMaterializedViewMeta,
 ) -> *const CassTableMeta {
     let view_meta = RefFFI::as_ref(view_meta);
-    view_meta.base_table.as_ptr()
+
+    match view_meta.base_table.upgrade() {
+        Some(arc) => RefFFI::as_ptr(&arc),
+        None => std::ptr::null(),
+    }
```
wprzytula: Why are we calling `RefFFI::as_ptr` on an `Arc`? I would expect `ArcFFI::as_ptr`.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,250 @@ macro_rules! make_c_str {
 #[cfg(test)]
 pub(crate) use make_c_str;
 
-/// Defines a pointer manipulation API for non-shared heap-allocated data.
+mod sealed {
+    pub trait Sealed {}
+}
+
+pub trait Ownership: sealed::Sealed {}
```
wprzytula: This commit (`argconv: pointers`) misses description, which is much needed here for future readers to understand reasons of adding this.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,250 @@ macro_rules! make_c_str {
...
-/// and then returned to the user as a pointer. The user is responsible for freeing
-/// the memory associated with the pointer using corresponding driver's API function.
-pub trait BoxFFI {
-    fn into_ptr(self: Box<Self>) -> *mut Self {
-        #[allow(clippy::disallowed_methods)]
-        Box::into_raw(self)
+/// Notice that this type does not provide interior mutability itself.
+/// It is the responsiblity of the user of this API, to provide soundness
+/// in this matter (aliasing ^ mutability).
+pub type CassSharedPtr<T> = CassPtr<T, (Shared, Const)>;
```
wprzytula: Maybe let's point to the `CassDataType` here, so that future readers can reach there for a reference on how to provide sound interior mutability to `CassSharedPtr` semantics?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,250 @@ macro_rules! make_c_str {
...
+        }
+    }
+
+    fn is_null(&self) -> bool {
+        self.ptr.is_none()
     }
-    unsafe fn maybe_as_ref<'a>(ptr: *const Self) -> Option<&'a Self> {
+
+    /// Convets a pointer to an optional valid reference.
+    unsafe fn into_ref<'a>(self) -> Option<&'a T> {
```
wprzytula: Converts

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,250 @@ macro_rules! make_c_str {
...
+        // Thus, if the pointer is non-null, we are guaranteed that it's valid.
+        self.ptr.map(|p| p.as_ref())
+    }
+}
+
+/// Methods for any exclusive pointers (no matter the mutability).
+impl<T, M: Mutability> CassPtr<T, (Exclusive, M)> {
+    /// Creates a pointer based on a VALID Box allocation.
+    /// This is the only way to obtain such pointer.
+    fn new(b: Box<T>) -> Self {
```
wprzytula: I'm not sure about this suggestion, because I haven't seen usage yet, but my idea is:
Wouldn't `from_box` be clearer about what's expected here instead of a generic `new`? With the constructor named `new` for every `CassPtr` instantiation but accepting different arguments, this may resemble the loathed (at least by me) method overloading (which adds complexity to understanding things, because a method of a given name may in reality be multiple different methods).

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,250 @@ macro_rules! make_c_str {
...
+    /// Converts the pointer back to the Arc.
+    unsafe fn into_arc(self) -> Option<Arc<T>> {
+        // SAFETY: The pointer can only be obtained via new() or from_ref(),
+        // both of which accept an Arc.
+        // It means that the pointer comes from a valid allocation.
+        self.ptr.map(|p| {
+            #[allow(clippy::disallowed_methods)]
+            Arc::from_raw(p.as_ptr())
+        })
+    }
```
wprzytula: Does this need to be `unsafe`, if you argue in the comment that it's safe?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -374,3 +380,19 @@ pub trait RefFFI: Sized {
...
+impl<T> own_sealed::ExclusiveSealed for T where T: FFI<Ownerhsip = OwnershipExclusive> {}
+impl<T> BoxFFI for T where T: FFI<Ownerhsip = OwnershipExclusive> {}
+
+pub struct OwnershipShared;
+impl<T> own_sealed::SharedSealed for T where T: FFI<Ownerhsip = OwnershipShared> {}
+impl<T> ArcFFI for T where T: FFI<Ownerhsip = OwnershipShared> {}
+
+pub struct OwnershipBorrowed;
+impl<T> own_sealed::BorrowedSealed for T where T: FFI<Ownerhsip = OwnershipBorrowed> {}
+impl<T> RefFFI for T where T: FFI<Ownerhsip = OwnershipBorrowed> {}
```
wprzytula: Missing docstrings.

wprzytula: At least `FFI` should be documented.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -374,3 +380,19 @@ pub trait RefFFI: Sized {
         ptr.is_null()
     }
 }
+
+pub trait FFI {
+    type Ownerhsip;
```
wprzytula: typo: `Ownerhsip` -> `Ownership`.

wprzytula (changes requested):
The approach is very impressive.
I couldn't check all the changes, there were too much of them; I just peeked through them. I trust your meticulousness :)

muzarski:
Rebased on master

scylla-rust-wrapper/src/collection.rs
muzarski: Good point. Please, see: https://github.com/scylladb/cpp-rust-driver/pull/212 (removes `lazy_static` dependency altogether)

scylla-rust-wrapper/src/collection.rs
muzarski: The comment was already there a long time ago:
```
// Changed return type to const ptr - ArcFFI::into_ptr is const.
// It's probably not a good idea - but cppdriver doesn't guarantee
// thread safety apart from CassSession and CassFuture.
// This comment also applies to other functions that create CassDataType.
```

However, it's going to be outdated once we migrate to new pointer types later in this PR. But your comment reminded me that I did not remove the outdated comment next to `cass_data_type_new` :).

scylla-rust-wrapper/src/metadata.rs
muzarski: See the implementation (and the comment):
```rust
// Either:
// - owned by CassMaterializedViewMeta - won't be given to user
// - Owned by CassKeyspaceMeta (in Arc), referenced (Weak) by CassMaterializedViewMeta
impl FFI for CassTableMeta {
    type Ownerhsip = OwnershipBorrowed;
}
```

By implementing the `RefFFI` for this type, we purposely narrow the set of functionalities exposed on the pointer type representing `CassTableMeta`. Specifically, we do not let do the following:
- convert the pointer back to `Arc` (i.e. free)
- obtain an `Arc` by cloning (increasing RC)

Thus, when implementing `RefFFI` for this type, we make a promise (expressed in Rust type system), that we do not make any assumption about the pointer (specifically, we do not care whether it was allocated via `Arc`, or it was simply obtained from some valid reference). When implementing `RefFFI`, we only care about the validity of the pointer - the fact that it can be converted to some **valid** reference.

It's exactly why `RefFFI` is a good choice for `CassTableMeta` - the user does not need to free the memory associated with the pointer (it's automatically done, when either `CassMaterializedViewMeta` or `CassKeyspaceMeta` is freed). The user also cannot clone (increase the reference count) it in any way.

The reason we store `CassTableMeta` inside an `Arc` in `CassKeyspaceMeta` (the owner), is because we want to have a weak reference to it in `CassMaterializedViewMeta`.

scylla-rust-wrapper/src/argconv.rs
muzarski: Good observation - adjusted remaining occurrences as well.

scylla-rust-wrapper/src/argconv.rs
muzarski: Applied the suggestion. Did the same for `CassSharedPtr` (`from_arc`)

scylla-rust-wrapper/src/metadata.rs
wprzytula: Great explanation, thank you.
I believe it deserves being put in a comment!

muzarski:
v1.1: Addressed @wprzytula comments

scylla-rust-wrapper/src/argconv.rs
muzarski: Added docs for the `FFI` trait and the structs

wprzytula:
Waiting with review for a rebase on #207.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +72,398 @@ macro_rules! make_c_str {
...
+    fn null() -> CassBorrowedPtr<Self> {
+        CassBorrowedPtr::null()
+    }
+    fn is_null(ptr: CassBorrowedPtr<Self>) -> bool {
+        ptr.is_null()
+    }
+}
+
+/// This traits should be implement for types that are passed between
+/// C and Rust API. We currently distinguish 3 kinds of implementors,
```
wprzytula: :wrench: typos: `This trait should be implemented`

muzarski:
> Waiting with review for a rebase on #207.

Everything is up to date

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +72,398 @@ macro_rules! make_c_str {
...
+impl Mutability for Const {}
+
+pub struct Mut;
+impl sealed::Sealed for Mut {}
+impl Mutability for Mut {}
+
+pub trait Properties {
+    type Ownership: Ownership;
+    type Mutability: Mutability;
+}
```
wprzytula: :question: Do you think these items can be left without docstrings? I think I'd add at least some basic ones.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +72,398 @@ macro_rules! make_c_str {
...
+    fn into_box(self) -> Option<Box<T>> {
+        // SAFETY: We are guaranteed that if ptr is Some, then it was obtained
+        // from a valid Box allocation (see new() implementation).
+        unsafe {
+            self.ptr.map(|p| {
+                #[allow(clippy::disallowed_methods)]
+                Box::from_raw(p.as_ptr())
+            })
+        }
+    }
```
wprzytula: :warning: :wrench:  Contrary to what I previously said about this function not being `unsafe`, there exists a scenario that triggers UB. Namely, `Self`, being a `CassPtr`, implements `Clone`; thus, when having two copies of the same pointer, one can call `into_box` on both, causing a _double free_.

Actually, even though this method accept `self` by value, it does not consume it, because `Self: Copy`.

wprzytula (changes requested):

scylla-rust-wrapper/src/argconv.rs
muzarski: Wow, I'm glad you noticed that. This is obviously a bug. So my proposition would be:
- implement Copy and Clone with `#[cfg(test)]` - in unit tests we get the pointers from API functions, and we need to be able to `Copy` them (same as C API user)
- `[Box/Arc/Ref]FFI::as_ref()` should accept `&CassPtr<_, _>` instead of `CassPtr<_, _>` - this would represent an aliasing to a valid non-mutable reference.
- remaining methods, e.g. `into_box`, `into_arc`, `into_mut_ref` (aliasing ^ mutability) should consume the pointer

WDYT?

wprzytula: > - implement Copy and Clone with #[cfg(test)] - in unit tests we get the pointers from API functions, and we need to be able to Copy them (same as C API user)

:warning: This unfortunately is still not sound. The `test` code must uphold to the same rules as non-`test` code, which means that safe code there must not trigger UB in any case.

> - [Box/Arc/Ref]FFI::as_ref() should accept &CassPtr<_, _> instead of CassPtr<_, _> - this would represent an aliasing to a valid non-mutable reference. The lifetime of reference would be inherited from the lifetime of the pointer
> - remaining methods, e.g. into_box, into_arc, into_mut_ref (aliasing ^ mutability) should consume the pointer

:+1: I'm fine with these ideas.

muzarski: > > * [Box/Arc/Ref]FFI::as_ref() should accept &CassPtr<_, _> instead of CassPtr<_, _> - this would represent an aliasing to a valid non-mutable reference. The lifetime of reference would be inherited from the lifetime of the pointer
> > * remaining methods, e.g. into_box, into_arc, into_mut_ref (aliasing ^ mutability) should consume the pointer
> 
> 👍 I'm fine with these ideas.

There are a couple things that need to be addressed (and discussed)

## API functions accept owned pointers
This is a problem, when some API function calls other API function. For example `cass_result_paging_state_token` does something like this:
```rust
    if cass_result_has_more_pages(result) == cass_false {
        return CassError::CASS_ERROR_LIB_NO_PAGING_STATE;
    }

    ... keep on using result, but it is consumed by cass_result_has_more_pages...
```
A workaround, which in my opinion makes sense is to extract the `cass_result_has_more_pages` to some function that accepts a reference to the pointer. This should be easy - there are only a couple of places where something like this occurs.

## Sometimes, we need to "erase" the lifetime of the reference
In some cases, we do not want to inherit the lifetime from the pointer provided to the API call. In other words, the reference should be valid when function call ends. From my investigation, there are two cases:
1. When we "erase" the lifetime in iterators (i.e. cast to `'static`). This one probably will not be a problem once replace `'static` lifetimes in iterators with some named lifetimes.
2. When we pass the `CassSession` reference to some future which is required to be `'static`. For example, see the comment to `CassSessionInner::connect()`:
```rust
    fn connect(
        // This reference is 'static because this is the only was of assuring the borrow checker
        // that holding it in our returned future is sound. Ideally, we would prefer to have
        // the returned future's lifetime constrained by real lifetime of the session's RwLock,
        // but this is impossible to be guaranteed due to C/Rust cross-language barrier.
        session_opt: &'static RwLock<Option<CassSessionInner>>,
        cluster: &CassCluster,
        keyspace: Option<String>,
    )
```

The solution: Introduce `[...]FFI::into_ref()` methods which will **CONSUME** the pointer, in favor of reference not inheriting the pointer's lifetime. Since we consume, the borrow-checker will work as expected with other FFI API functions.

As I mentioned, I believe that issue `1.` will disappear once we replace `'static` lifetimes in iterators with some named lifetimes (I might be too optimistic, though).
I'm not sure whether issue `2.` can be addressed some other way.

## CassFuture set_callback case...

This one is really tricky. This is the only place in current code, where we need to operate on an owned pointer, by passing it to the method called on a reference that inherits the pointer's lifetime. I'm not really sure how to address it. I believe this is the main reason I implemented `Copy` for `CassPtr`.

Here is a code snippet:
```rust
#[no_mangle]
pub unsafe extern "C" fn cass_future_set_callback(
    future_raw: CassSharedPtr<CassFuture>,
    callback: CassFutureCallback,
    data: *mut ::std::os::raw::c_void,
) -> CassError {
    ArcFFI::as_ref(&future_raw)
        .unwrap()
        .set_callback(future_raw, callback, data)
}
```

Compiler error:
```rust
error[E0505]: cannot move out of `future_raw` because it is borrowed
   --> src/future.rs:304:23
    |
298 |     future_raw: CassSharedPtr<CassFuture>,
    |     ---------- binding `future_raw` declared here
...
302 |     ArcFFI::as_ref(&future_raw)
    |                    ----------- borrow of `future_raw` occurs here
303 |         .unwrap()
304 |         .set_callback(future_raw, callback, data)
    |          ------------ ^^^^^^^^^^ move out of `future_raw` occurs here
    |          |
    |          borrow later used by call
    |
note: if `CassPtr<CassFuture, (argconv::Shared, Const)>` implemented `Clone`, you could clone the value
   --> src/argconv.rs:154:1
    |
154 | pub struct CassPtr<T: Sized, P: Properties> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ consider implementing `Clone` for this type
    |
   ::: src/future.rs:302:21
    |
302 |     ArcFFI::as_ref(&future_raw)
    |                     ---------- you could clone this value

```

wprzytula: > ## CassFuture set_callback case...
> 
> This one is really tricky. This is the only place in current code, where we need to operate on an owned pointer, by passing it to the method called on a reference that inherits the pointer's lifetime. I'm not really sure how to address it. I believe this is the main reason I implemented `Copy` for `CassPtr`.

From what I see, the callback is used as an opaque pointer; what's important about it is merely its address. Then, we could make `set_callback` accept `*const CassFuture` and adjust `CassFutureCallback` accordingly. The exact FFI is irrelevant in this specific case, because no allocations/deallocations happen there anyway.
Then, we could have a safe API function `to_raw(&self)` on `CassPtr` that would return the raw pointer. The Rust code as such does nothing with that particular pointer, only passes it down to the callback function (which is going to be implemented in C/C++). WDYT?

muzarski: I love it! Thank you.

scylla-rust-wrapper/src/argconv.rs
muzarski: Added basic docstrings

muzarski:
v2: Improved pointer API based on @wprzytula suggestions. Thank you!
- removed `Copy` and `Clone` implementations for `CassPtr` (currently, it's only defined for test purposes. I need to think how to address it some other way)
- Left `CassPtr::into_ref()` as is. As discussed in the comment, it is required to "erase" the lifetime of the returned reference. This method consumes the pointer.
- Introduced `CassPtr::as_ref()` method. It accepts a reference to the pointer. The lifetime of reference is inherited from pointer's lifetime.
- Changed `CassExclusiveMutPtr::into_mut_ref()` to `[...]::as_mut_ref()`. It accepts `&mut CassExclusiveMutPtr`. The lifetime of returned mutable reference is inherited from mutable borrow of the pointer. This is really cool, because this ensures `aliasing ^ mutability` at borrow-checker level.
- `CassPtr::is_null()` now borrows the pointer, instead of consuming it
- `CassFuture::set_callback()`, and `CassFutureCallback` now expect a raw pointer. Safe `CassPtr::as_raw()` method is introduced for this use-case

I still need to think how to address the issue in tests. It's not possible to reuse the pointer, which is not `Copy`, while passing it to multiple API functions that consume the pointer.

wprzytula:
> I still need to think how to address the issue in tests. It's not possible to reuse the pointer, which is not `Copy`, while passing it to multiple API functions that consume the pointer.

You could have an `unsafe` `clone()` method for tests only, then this would not break soundness guarantees of safe code.

muzarski:
> > I still need to think how to address the issue in tests. It's not possible to reuse the pointer, which is not `Copy`, while passing it to multiple API functions that consume the pointer.
> 
> You could have an `unsafe` `clone()` method for tests only, then this would not break soundness guarantees of safe code.

Applied the suggestion. At first, I did not understand how is this different from implementing `Copy` and `Clone` for `test` only. But this part: "this would not break soundness guarantees of **safe** code" convinced me.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +72,439 @@ macro_rules! make_c_str {
...
+/// Unsafe implementation of clone for tests.
+#[cfg(test)]
+impl<T: Sized, P: Properties> CassPtr<T, P> {
+    pub unsafe fn clone(&self) -> Self {
+        Self {
+            ptr: self.ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
```
wprzytula: :+1: Now it looks correct. Let's put motivation for this in a comment (why we even need to `clone` in tests).

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -97,3 +72,439 @@ macro_rules! make_c_str {
...
+    }
+    fn null() -> CassBorrowedPtr<Self> {
+        CassBorrowedPtr::null()
+    }
+    fn is_null(ptr: &CassBorrowedPtr<Self>) -> bool {
+        ptr.is_null()
+    }
+}
+
+/// This trait should be implement for types that are passed between
```
wprzytula: :pick:  typo: `implement` -> `implemented`

scylla-rust-wrapper/src/batch.rs
```diff
@@ -156,12 +163,12 @@ pub unsafe extern "C" fn cass_batch_set_tracing(
...
+    mut batch: CassExclusiveMutPtr<CassBatch>,
+    statement: CassExclusiveConstPtr<CassStatement>,
 ) -> CassError {
-    let batch = ptr_to_ref_mut(batch);
+    let batch = BoxFFI::as_mut_ref(&mut batch).unwrap();
     let state = Arc::make_mut(&mut batch.state);
-    let statement = ptr_to_ref(statement);
+    let statement = BoxFFI::as_ref(&statement).unwrap();
 
     match &statement.statement {
```
wprzytula: :question: :pushpin: Is this function still correct in case `Arc::make_mut(&mut batch.state)` returns an `Arc` to a new allocation, because there already have been multiple owners of `state`?

wprzytula (approved):

scylla-rust-wrapper/src/batch.rs
muzarski: This is as valid, as it was before the changes - we did not change the logic. But to answer your question (is it valid), I'd need to investigate it. TBH, I haven't touched batches in cpp-rust-driver yet. However, all these `Arc::make_mut` calls are a bit suspicious to me.

wprzytula: Yep, this is out of scope of this PR. Just wanted to raise your awareness about the potential bug here.

muzarski:
Rebased on master

muzarski:
v2.1: improved docstrings as suggested by @wprzytula

scylla-rust-wrapper/src/batch.rs
muzarski: I investigated this. It looks fine to me. It looks like a small optimization.

The reason is that we need to clone fields of `CassBatch` in `cass_session_execute_batch` - we should not make any assumptions about the provided `CassBatch` object in the functions - it could have been modified, or even freed:
```rust
    let session_opt = ArcFFI::into_ref(session_raw).unwrap();
    let batch_from_raw = BoxFFI::as_ref(&batch_raw).unwrap();
    let mut state = batch_from_raw.state.clone();
    let request_timeout_ms = batch_from_raw.batch_request_timeout_ms;

    // DO NOT refer to `batch_from_raw` inside the async block, as I've done just to face a segfault.
    let batch_exec_profile = batch_from_raw.exec_profile.clone();
```

So how the optimization works?
This statement `let mut state = batch_from_raw.state.clone();` is cheap, because it clones an Arc.

When it works?
Well, `Arc::make_mut` can potentially allocate, but this will not be the case if there is no concurrent execution of `cass_session_execute_batch` in the background - i.e., there is no other owner of the state.

Notice that this is all sound, because we do not expose the state to the users in any way.

Lorak-mmk:
I've read the first few commits, and it talks about `cass_data_type_new_from_existing` expecting a Arc-allocated CassDataType, but I don't see why.
For context, the code of this function:
```
#[no_mangle]
pub unsafe extern "C" fn cass_data_type_new_from_existing(
    data_type: *const CassDataType,
) -> *const CassDataType {
    let data_type = ptr_to_ref(data_type);
    Arc::into_raw(Arc::new(data_type.clone()))
}
```

As far as I can tell it copies the whole object, not Arc, and does not assume it is Arc-allocated.

muzarski:
> I've read the first few commits, and it talks about `cass_data_type_new_from_existing` expecting a Arc-allocated CassDataType, but I don't see why. For context, the code of this function:
> 
> ```
> #[no_mangle]
> pub unsafe extern "C" fn cass_data_type_new_from_existing(
>     data_type: *const CassDataType,
> ) -> *const CassDataType {
>     let data_type = ptr_to_ref(data_type);
>     Arc::into_raw(Arc::new(data_type.clone()))
> }
> ```
> 
> As far as I can tell it copies the whole object, not Arc, and does not assume it is Arc-allocated.

Sorry for the confusion. I remember that I made a mistake in some commit message, but forgot to fix it. I meant `cass_data_type_add_sub_type` function:
```rust
#[no_mangle]
pub unsafe extern "C" fn cass_data_type_add_sub_type(
    data_type: CassSharedPtr<CassDataType>,
    sub_data_type: CassSharedPtr<CassDataType>,
) -> CassError {
    let data_type = ArcFFI::as_ref(&data_type).unwrap();
    match data_type
        .get_mut_unchecked()
        .add_sub_data_type(ArcFFI::cloned_from_ptr(sub_data_type).unwrap())
    {
        Ok(()) => CassError::CASS_OK,
        Err(e) => e,
    }
}
```

muzarski:
Pushed fixed commit message (`collection: allocate static empty collections`).

Lorak-mmk:
The same mistake is present in `metadata: store CassDataType behind an Arc ` commit message.

scylla-rust-wrapper/src/batch.rs
wprzytula: :+1:  Please put the above findings in a comment there, not to confuse future readers.

muzarski:
Rebased on master

scylla-rust-wrapper/src/metadata.rs
Lorak-mmk: Is it possible for this upgrade to fail? Or does it indicate a bug in the driver?
If it is possible, can we have a test for that?

scylla-rust-wrapper/src/future.rs
```diff
@@ -40,9 +40,9 @@ struct BoundCallback {
...
 
 impl BoundCallback {
-    fn invoke(self, fut: &CassFuture) {
+    fn invoke(self, fut_ptr: *const CassFuture) {
         unsafe {
-            self.cb.unwrap()(fut as *const CassFuture, self.data);
+            self.cb.unwrap()(fut_ptr, self.data);
         }
     }
 }
```
Lorak-mmk: Shouldn't this `invoke` method be unsafe? It does have a requirement for its argument to be a correct (alingment, lifetime etc) pointer.

scylla-rust-wrapper/src/future.rs
```diff
@@ -258,7 +259,12 @@ impl CassFuture {
...
-    pub fn set_callback(&self, cb: CassFutureCallback, data: *mut c_void) -> CassError {
+    pub fn set_callback(
+        &self,
+        self_ptr: *const CassFuture,
+        cb: CassFutureCallback,
+        data: *mut c_void,
+    ) -> CassError {
         let mut lock = self.state.lock().unwrap();
         if lock.callback.is_some() {
             // Another callback has been already set
```
Lorak-mmk: Here `self_ptr` should refer to the same `CassFuture` as `self`, right?
Can you instead make the receiver of this function an arc? `self: &Arc<CassFuture>`

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+
+/// Represents immutable pointer.
+pub struct Const;
+impl sealed::Sealed for Const {}
+impl Mutability for Const {}
+
+/// Represents mutable pointer.
+pub struct Mut;
+impl sealed::Sealed for Mut {}
+impl Mutability for Mut {}
```
Lorak-mmk: 🔧 Could you describe more what mutability means here? It is not obvious in ffi-related code:
- It could be Rust mutability (where mutable reference = exclusive reference)
- It could be C++ mutability: logical mutability (so if object is const, its logical state won't change).

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+
+// repr(transparent), so the struct has the same layout as underlying Option<NonNull<T>>.
+// Thanks to https://doc.rust-lang.org/std/option/#representation optimization,
+// we are guaranteed, that for T: Sized, our struct has the same layout
+// and function call ABI as simply NonNull<T>.
+#[repr(transparent)]
+pub struct CassPtr<T: Sized, P: Properties> {
+    ptr: Option<NonNull<T>>,
+    _phantom: PhantomData<P>,
+}
```
Lorak-mmk: ⛏ Could you move it above the aliases? It is a bit confusing when reading the code: you see aliases that use `CassPtr`, but the `CassPtr` was not yet defined.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+/// For Arc-allocated pointers.
+pub struct Shared;
+impl sealed::Sealed for Shared {}
+impl Ownership for Shared {}
+
+/// Represents a borrowed pointer.
+/// For pointers created from some valid reference.
+pub struct Borrowed;
+impl sealed::Sealed for Borrowed {}
+impl Ownership for Borrowed {}
```
Lorak-mmk: 🔧 This also could use a bit more extensive doc comments for future contributors.
If a function receives `Shared` pointer, what it means? Is it guaranteed to be Arc-allocated, or is it logical sharing (and Arc is a job of ArcFFI, a higher layer)? Is the pointer borrowed to us, or given to us? It is important because it determines who (we or caller) is responsible for pointee destruction.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+        self.ptr.is_none()
+    }
+
+    /// Converts a pointer to an optional valid reference.
+    fn as_ref(&self) -> Option<&T> {
+        // SAFETY: We know that our pointers either come from a valid allocation (Box or Arc),
+        // or were created based on the reference to the field of some allocated object.
+        // Thus, if the pointer is non-null, we are guaranteed that it's valid.
+        unsafe { self.ptr.map(|p| p.as_ref()) }
+    }
```
Lorak-mmk: 💯 CassPtr is an object passed as function argument, so its lifetime is tied to the function scope, so this lifetime won't we static. I like that.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+
+    /// Converts a pointer to an optional valid reference, and consumes the pointer.
+    fn into_ref<'a>(self) -> Option<&'a T> {
+        // SAFETY: We know that our pointers either come from a valid allocation (Box or Arc),
+        // or were created based on the reference to the field of some allocated object.
+        // Thus, if the pointer is non-null, we are guaranteed that it's valid.
+        unsafe { self.ptr.map(|p| p.as_ref()) }
+    }
```
Lorak-mmk: OTOH here we can easily create a 'static reference from an object that was just borrowed to us for the function call!

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+
+    /// Converts self to raw pointer.
+    fn as_raw(&self) -> *const T {
+        self.ptr
+            .map(|ptr| ptr.as_ptr() as *const _)
+            .unwrap_or(std::ptr::null())
+    }
+}
```
Lorak-mmk: 🤔 It will probably become clear later, but after reading this I'm wondering why do we need this - isn't the goal to not use raw pointers?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+/// We introduce unsafe `clone` method for this reason.
+#[cfg(test)]
+impl<T: Sized, P: Properties> CassPtr<T, P> {
+    pub unsafe fn clone(&self) -> Self {
+        Self {
+            ptr: self.ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
```
Lorak-mmk: If it is unsafe, could you please describe safety requirements?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
+        // both of which accept an Arc.
+        // It means that the pointer comes from a valid allocation.
+        unsafe {
+            self.ptr.map(|p| {
+                #[allow(clippy::disallowed_methods)]
+                Arc::from_raw(p.as_ptr())
+            })
+        }
+    }
+
```
Lorak-mmk:
```rust
let obj: Arc<()> = Arc::new(());
let x: CassSharedPtr<()> = CassSharedPtr::from_ref(&obj);
std::mem::drop(obj);
let obj2: Arc<()> = CassSharedPtr::into_arc(x);
```

Isn't that use-after-free with safe code?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,35 +73,307 @@ macro_rules! make_c_str {
...
-        #[allow(clippy::disallowed_methods)]
+    fn as_ref<M: Mutability>(ptr: &CassExclusivePtr<Self, M>) -> Option<&Self> {
         ptr.as_ref()
     }
-    unsafe fn as_ref<'a>(ptr: *const Self) -> &'a Self {
-        #[allow(clippy::disallowed_methods)]
-        ptr.as_ref().unwrap()
+    fn into_ref<'a, M: Mutability>(ptr: CassExclusivePtr<Self, M>) -> Option<&'a Self> {
+        ptr.into_ref()
     }
```
Lorak-mmk: Do we need to expose `into_ref`? It is quite easy to cause a memory leak with it.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -108,36 +382,37 @@ pub trait BoxFFI {
...
+    fn free(ptr: CassSharedPtr<Self>) {
         std::mem::drop(ArcFFI::from_ptr(ptr));
     }
+    fn null() -> CassSharedPtr<Self> {
+        CassSharedPtr::null()
+    }
+    fn is_null(ptr: &CassSharedPtr<Self>) -> bool {
+        ptr.is_null()
+    }
 }
```
Lorak-mmk:
```rust
let obj: Arc<()> = Arc::new(());
let ptr = ArcFFI::as_ptr(&obj);
std::mem::drop(obj);
let obj2 = ArcFFI::from_ptr(ptr);
```

Another use after free :(

Lorak-mmk (changes requested):
I did not review the whole thing, because I spotted some important issues in argconv.rs. In general, please describe with more detail in precision what guarantees / requirements the types / methods require and provide. Without it it is a bit difficult to reason about the code.

scylla-rust-wrapper/src/argconv.rs
muzarski: That's exactly the use-case - to erase the lifetime. Otherwise, we wouldn't be able to return the types that borrow and hold references. For example, iterators hold some `'static` references.

scylla-rust-wrapper/src/argconv.rs
muzarski: The only use case is `cass_future_set_callback` function. See the discussion: https://github.com/scylladb/cpp-rust-driver/pull/208#discussion_r1869828743

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: We did discuss the possibility of such structs being lifetime-generic iirc.

For now, you could hold a pointer in such struct instead of a static reference.

muzarski: Still, the issue is not solved. I have a branch with `lifetime-generic` iterators, and erasing the lifetime is still necessary.

Why?
The function accepts an owned `CassPtr`, and we want to return something with the lifetime of this pointer (which is a lifetime of function body). So in the end, we want to return a type that references something that has a lifetime of function call.

Example:
```rust
pub struct CassRowsResultIterator<'result> {
    iterator: TypedRowIterator<'result, 'result, CassRawRow<'result>>,
    result_metadata: &'result CassResultMetadata,
    current_row: Option<CassRow<'result>>,
}

#[no_mangle]
pub unsafe extern "C" fn cass_iterator_from_result<'result>(
    result: CassSharedPtr<CassResult>,
) -> CassExclusiveMutPtr<CassIterator<'result>> {
    let result_from_raw = ArcFFI::as_ref(&result).unwrap();

    let iterator = match &result_from_raw.kind {
        CassResultKind::NonRows => CassResultIterator::NonRows,
        CassResultKind::Rows(cass_rows_result) => {
            CassResultIterator::Rows(CassRowsResultIterator {
                iterator: cass_rows_result.raw_rows.rows_iter().unwrap(),
                result_metadata: &cass_rows_result.metadata,
                current_row: None,
            })
        }
    };

    BoxFFI::into_ptr(Box::new(CassIterator::CassResultIterator(iterator)))
}
```

Compiler error:
```rust
error[E0515]: cannot return value referencing function parameter `result`
    --> src/query_result.rs:1047:5
     |
1034 |     let result_from_raw = ArcFFI::as_ref(&result).unwrap();
     |                                          ------- `result` is borrowed here
...
1047 |     BoxFFI::into_ptr(Box::new(CassIterator::CassResultIterator(iterator)))
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

scylla-rust-wrapper/src/argconv.rs
muzarski: Yeah.. Analogous issue appears with `CassBorrowedPtr`:
```rust
struct Bar;

impl RefFFI for Bar { }

struct Owner {
   owned: Bar,
}

let owner = Owner { owned: Bar };
let borrowed_ptr: CassBorrowedPtr<Bar> = RefFFI::as_ptr(&owner.owned);
std::mem::drop(owner);
// A reference from invalid pointer
let bar = RefFFI::as_ref(&borrowed_ptr);
```

I don't think there is a safe way to prevent such scenarios. I'll document that this API can be misused (and so, should be `unsafe`) and where we expect valid pointers etc.

muzarski:
v3:
## New `Origin` trait and modified `Ownership` trait
Introduced new `Origin` trait which represents an origin of the pointer, and three implementors:
- `FromBox` - box allocated, exclusive (potentially mutable) pointers
- `FromArc` - arc-allocated, shared (either owned, or borrowed) const pointers
- `FromRef` - pointers created from a valid rust reference, always borrowed and const

Modified implementors of `Ownership` trait. Now we distinguish two ownership kinds:
- `Owned` - owned pointer. Data needs to be freed
- `Borrowed` - borrowed pointers. Conversions to reference/Arcs from such pointers are `unsafe` - caller needs to ensure that owner of the data is alive

## Additional commit
Moved the changes that introduce `CassPtr` (and traits) to a separate commit. I hope this aids readability and reduces the noise during review.

Thoroughly documented each type/trait and its methods. Explained what each type guarantees (and what it does not guarantee). Justified the choice of the methods being safe/`unsafe`.

scylla-rust-wrapper/src/argconv.rs
wprzytula: > ```rust
> let obj: Arc<()> = Arc::new(());
> let x: CassSharedPtr<()> = CassSharedPtr::from_ref(&obj);
> std::mem::drop(obj);
> let obj2: Arc<()> = CassSharedPtr::into_arc(x);
> ```
> 
> Isn't that use-after-free with safe code?

Wouldn't [Arc::increment_strong_count](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.increment_strong_count) called upon `CassSharedPtr::from_ref` solve this particular case? By the way, I believe we should be doing that shared ownership with pointers. I remember that in the past, cpp-rust-driver used to call that function.

Lorak-mmk: The way I understand it, `from_ref` is for borrowing data to the user. If you increment strong count, then user is a co-owner of that data, not a borrower. This could be used only in cases where user is supposed to call some `_free` function later.

muzarski: > > ```rust
> > let obj: Arc<()> = Arc::new(());
> > let x: CassSharedPtr<()> = CassSharedPtr::from_ref(&obj);
> > std::mem::drop(obj);
> > let obj2: Arc<()> = CassSharedPtr::into_arc(x);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Isn't that use-after-free with safe code?
> 
> Wouldn't [Arc::increment_strong_count](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.increment_strong_count) called upon `CassSharedPtr::from_ref` solve this particular case? By the way, I believe we should be doing that shared ownership with pointers. I remember that in the past, cpp-rust-driver used to call that function.

Sure, it solves this particular case. However, it will almost certainly cause memory leaks in some scenarios. 

Consider `CassDataType`. There are multiple functions that return pointer to `CassDataType`. However, the semantics differ based on the context. For example:
- `cass_data_type_new` -> returns a pointer with a shared **ownership**. The pointee needs to be freed with `cass_data_type_free`.
- `cass_result_column_data_type` -> returns a pointer originated from Arc-allocation, but it **lends** the data. User is not expected (and he should not!!!) free the data type. This is the exact use case for `CassSharedPtr::from_ref()` (`ArcFFI::as_ptr()`). If we increased the reference count, we have a memory leak.

The distinctions between two semantics are better visible after my latest push, where I decoupled `Ownership` trait and implementors into `Origin` (`FromBox`, `FromArc`, `FromRef`) and `Ownership` (`Owned`, `Borrowed`).
There, `cass_data_type_new` returns a pointer that is `FromArc` and `Owned` (and `cass_data_type_free` accepts only `FromArc`, `Owned` pointers as well).  OTOH, `cass_result_column_data_type` returns `FromArc`, `Borrowed` pointer.

wprzytula: > * `cass_result_column_data_type` -> returns a pointer originated from Arc-allocation, but it **lends** the data. User is not expected (and he should not!!!) free the data type. This is the exact use case for `CassSharedPtr::from_ref()` (`ArcFFI::as_ptr()`). If we increased the reference count, we have a memory leak.

Then it's clear that the UAF performed by the above code by @Lorak-mmk is the user's fault (as understood by the C API). Only in Rust do we have to make the API `unsafe`, because Rust understands the mere possibility of introducing UAF condition as a bug in API. Do I get this correctly?

muzarski: > Only in Rust do we have to make the API `unsafe`, because Rust understands the mere possibility of introducing UAF condition as a bug in API. Do I get this correctly?

Yes, for me this is `unsafe`, because we can copy and paste the exact code snippet @Lorak-mmk provided and cause UAF. In general, in my new push all of the conversions `Owned` ptr -> `reference/Arc/Box` are safe, while `Borrowed` ptr -> `reference/Arc` are `unsafe`, because we are not guaranteed that the owner of borrowed data is alive when we dereference the pointer.

BTW, I was wondering if there is a way for a pointer to somehow inherit the lifetime, when the pointer is being borrowed (i.e. created from some immutable borrow &). I'm not really sure if this is possible.

Lorak-mmk: Yes this is true. I'm currently reading this PR and will experiment with it a little.
What I'm currently thinking:
- CassPtr should be lifetime-generic instead of having `Ownership` AT. OwnedPtr would be static, BorrowedPtr could be created from it (and borrow from it).
- `Origin` should be tied to the type (and so, the FFI trait), not the pointer.

scylla-rust-wrapper/src/argconv.rs
muzarski: It's removed in recent push

scylla-rust-wrapper/src/argconv.rs
muzarski: I investigated where we previously needed to erase the lifetime. There were two cases:
- iterators with `'static` references. I replaced them with generic lifetimes
- CassSession `'static` reference passed to an async block. I changed it, so we pass an `Arc` to the async block. This is a good change (and a fix), because cpp-driver assumes that session object can be prematurely free'd by the user. It should not cause segfaults in executions that haven't been awaited yet. What's more, there is a test that expects that prematurely freed session awaits all the futures before closing. We are not doing that yet, thus the test is not enabled.

After adjusting these two things, I could get rid of `into_ref` method.

muzarski:
Rebased on master

scylla-rust-wrapper/src/metadata.rs
muzarski: AFAIU, it should not be possible, and it indicates a bug in the driver. Why?

`CassSchemaMeta` contains and **owns** all metadata. It's obtained from `cass_session_get_schema_meta` - allocated as Box. So we have following relationship between the metadata structs:
- `CassSchemaMeta` owns `CassKeyspaceMeta`
- `CassKeyspaceMeta` owns both `CassTableMeta` and `CassMaterializedViewMeta`
- `CassMaterializedViewMeta` holds a weak reference to `CassTableMeta` that is owned by `CassKeyspaceMeta`

An idea on how to obtain the state where this upgrade() returns `None` (and write analogous test):
1. get schema metadata `CassSchemaMeta`
2. obtain a pointer to some view (`CassMaterializedViewMeta`) - easily done using API methods (`cass_schema_meta_keyspace_by_name` then `cass_keyspace_meta_materialized_view_by_name`)
3. free schema metadata (`cass_schema_meta_free`)
4. Call `cass_materialized_view_meta_base_table` (where upgrade happens) with the pointer obtained from bullet 2. <- this will panic, because the pointer is invalid. The owner (`CassSchemaMeta`) has already been freed.

scylla-rust-wrapper/src/future.rs
muzarski: I don't see a good reason behind it. We get a pointer in `cass_future_set_callback`, then we would need to convert it to `Arc` (increase a refcount), just to then convert it back to pointer (decrease refcount) and pass the ponter to callback.

muzarski:
v4:
- Introduced `CMutability` property - it simply tells whether the pointer was `T*` or `const T*` - as defined in C API. It tells about **logical mutability** of the pointer.
- Rebased on main
- Updated cover letter

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,6 +73,261 @@ macro_rules! make_c_str {
...
+    /// Converts a pointer to an optional valid mutable reference.
+    /// The reference inherits the lifetime of the pointer.
+    #[allow(clippy::wrong_self_convention)]
+    fn as_mut_ref(self) -> Option<&'a mut T> {
+        // SAFETY: Thanks to the validity and aliasing ^ mutability guarantees,
+        // we can safely convert the pointer to valid mutable (and exclusive) reference with
+        // correct lifetime.
+        unsafe { self.ptr.map(|mut p| p.as_mut()) }
+    }
+}
```
wprzytula: :thought_balloon: Wouldn't it make more sense in naming to call these `into_ref`, `into_mut_ref` ?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -289,7 +289,7 @@ impl<'a, T: Sized, P: Properties> CassPtr<'a, T, P> {
     }
 }
 
-impl<'a, T: Sized> CassPtr<'a, T, (Mut, CMut)> {
+impl<'a, T: Sized, CM: CMutability> CassPtr<'a, T, (Mut, CM)> {
```
wprzytula: :wrench: This should be likely part of the previous commit.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -333,30 +333,71 @@ impl<T: Sized> CassPtr<'_, T, (Mut, CMut)> {
...
+
+    /// Frees the pointee.
+    fn free<CM: CMutability>(ptr: CassPtr<'static, Self, (Mut, CM)>) {
         std::mem::drop(BoxFFI::from_ptr(ptr));
     }
+
+    #[cfg(test)]
+    fn null<'a, CM: CMutability>() -> CassPtr<'a, Self, (Const, CConst)> {
+        CassPtr::null()
+    }
```
wprzytula: :question: :pick: Why is this only for tests? Even if it's not currently used in the code, I'd rather silence the unused warning, yet made it discoverable by the IntelliSense for future devs of this lib.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -365,36 +406,86 @@ pub trait BoxFFI {
...
-    unsafe fn from_ptr(ptr: *const Self) -> Arc<Self> {
-        #[allow(clippy::disallowed_methods)]
-        Arc::from_raw(ptr)
+
+    /// Converts shared owned pointer back to owned Arc.
+    fn from_ptr<CM: CMutability>(ptr: CassPtr<'static, Self, (Const, CM)>) -> Option<Arc<Self>> {
+        // SAFETY:
+        // The only way to obtain a pointer with shared ownership ('static lifetime) is
+        // ArcFFI::into_ptr(). It converts an owned Arc into the pointer. It is thus safe,
+        // to convert such pointer back to owned Arc.
```
wprzytula: :pick:

wprzytula:
```suggestion
        // ArcFFI::into_ptr(). It converts an owned Arc into the pointer. It is thus safe
        // to convert such pointer back to owned Arc.
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -365,36 +406,86 @@ pub trait BoxFFI {
...
-    fn into_ptr(self: Arc<Self>) -> *const Self {
+
+    /// Creates a pointer from a valid Arc allocation.
+    fn into_ptr<CM: CMutability>(self: Arc<Self>) -> CassPtr<'static, Self, (Const, CM)> {
         #[allow(clippy::disallowed_methods)]
-        Arc::into_raw(self)
+        let ptr = Arc::into_raw(self);
+
+        // SAFETY:
+        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation
```
wprzytula, line 436: :pick:

wprzytula:
```suggestion
        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation.
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -365,36 +406,86 @@ pub trait BoxFFI {
...
+        // 2. pointer's lifetime - returned pointer has a 'static lifetime. It is a shared
+        //    owner of the pointee. User has to decrease the RC of the pointer (and potentially free the memory)
+        //    via ArcFFI::free().
```
wprzytula: :pick: to be more specific

wprzytula:
```suggestion
        // 2. pointer's lifetime - returned pointer has a 'static lifetime. It is a shared
        //    owner of the pointee. User has to decrement the RC of the pointer (and potentially free the memory)
        //    via ArcFFI::free().
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -365,36 +406,86 @@ pub trait BoxFFI {
...
-        #[allow(clippy::disallowed_methods)]
-        Arc::increment_strong_count(ptr);
-        #[allow(clippy::disallowed_methods)]
-        Arc::from_raw(ptr)
+
+    /// Increases the reference count of the pointer, and returns an owned Arc.
+    fn cloned_from_ptr<CM: CMutability>(ptr: CassPtr<'_, Self, (Const, CM)>) -> Option<Arc<Self>> {
+        // SAFETY:
+        // All pointers created via ArcFFI API are originated from Arc allocation.
+        // It is thus safe, to increase the reference count of the pointer, and convert
```
wprzytula: :pick:

wprzytula:
```suggestion
        // It is thus safe to increase the reference count of the pointer, and convert
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -365,36 +406,86 @@ pub trait BoxFFI {
...
+        // it to Arc. Because of the borrow-checker, it is not possible for the user
+        // to provide the pointer that points to already deallocated memory.
```
wprzytula: :pick:

wprzytula:
```suggestion
        // it to Arc. Because of the borrow-checker, it is not possible for the user
        // to provide a pointer that points to already deallocated memory.
```

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -405,12 +496,117 @@ pub trait ArcFFI {
...
+        match w.upgrade() {
+            Some(a) => {
+                #[allow(clippy::disallowed_methods)]
+                let ptr = Arc::as_ptr(&a);
+                unsafe { CassPtr::from_raw(ptr) }
+            }
+            None => CassPtr::null(),
+        }
+    }
+
```
wprzytula: :thought_balloon: I'm wondering why the function needs to be `unsafe`. Is the following scenario one valid reason for that?
```rust
let data = ...;
let arc = Arc::new(data);
let weak = Arc::downgrade(&arc);
let ptr = weak_as_ptr(&weak);
std::mem::drop(arc);

// Now `ptr` is dangling...
```

wprzytula: If I'm correct, maybe let's put this short example in the docstring of this function? Just to convince readers that `unsafe` is needed here.

scylla-rust-wrapper/src/batch.rs
```diff
@@ -158,12 +158,12 @@ pub unsafe extern "C" fn cass_batch_set_tracing(
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_batch_add_statement(
-    batch: *mut CassBatch,
-    statement: *const CassStatement,
+    batch: CassBorrowedMutPtr<CassBatch, CMut>,
+    statement: CassBorrowedPtr<CassStatement, CMut>,
```
wprzytula: :thought_balloon: I'm wondering why it's `CassBorrowedPtr` and not `CassBorrowedConstPtr`. After all, it's `*const` and `*mut` in Rust, so if want to be in line with this naming, we should include the `Const` part. Also, the constness guarantee would stand out more in code.

scylla-rust-wrapper/src/binding.rs
```diff
@@ -104,7 +104,7 @@ macro_rules! make_name_n_binder {
             use crate::value::CassCqlValue::*;
             let name = ptr_to_cstr_n(name, name_length).unwrap();
             match ($e)($($arg), *) {
-                Ok(v) => $consume_v(BoxFFI::as_mut_ref(this), name, v),
+                Ok(v) => $consume_v(BoxFFI::as_mut_ref( this).unwrap(), name, v),
```
wprzytula: :pick:

wprzytula:
```suggestion
                Ok(v) => $consume_v(BoxFFI::as_mut_ref(this).unwrap(), name, v),
```

scylla-rust-wrapper/src/binding.rs
```diff
@@ -116,14 +116,14 @@ macro_rules! make_appender {
...
-            this: *mut $this,
+            this: CassBorrowedMutPtr<$this, CMut>,
             $($arg: $t), *
         ) -> CassError {
             // For some reason detected as unused, which is not true
             #[allow(unused_imports)]
             use crate::value::CassCqlValue::*;
             match ($e)($($arg), *) {
-                Ok(v) => $consume_v(BoxFFI::as_mut_ref(this), v),
+                Ok(v) => $consume_v(BoxFFI::as_mut_ref( this).unwrap(), v),
```
wprzytula: :pick:

wprzytula:
```suggestion
                Ok(v) => $consume_v(BoxFFI::as_mut_ref(this).unwrap(), v),
```

scylla-rust-wrapper/src/lib.rs
```diff
@@ -8,7 +8,7 @@ use tokio::runtime::Runtime;
 
 #[macro_use]
 mod binding;
-mod argconv;
+pub mod argconv;
```
wprzytula, line 12: :question: Why is this needed?

wprzytula (changes requested):

scylla-rust-wrapper/src/argconv.rs
muzarski: TBH, I couldn't decide. According to https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv, there is no naming convention that matches our case.

I understand your reasoning - you suggest `into_` because methods accepts `self` (owned type). OTOH, how I understand this operation is that we convert one referential type (wrapper over raw pointer) to some other referential type - `as_` convention somehow sounds more natural to me in such case.

wprzytula: For me, the more important distinction is between `borrow()` (which borrows) and this method (which consumes) the caller. That's why my suggestion.
Normal references and pointers are `Copy`, whereas `CassPtr` is not even `Clone`. Therefore, it's different from usual Rust references and thus it should be closer to owned types wrt naming.

scylla-rust-wrapper/src/batch.rs
muzarski: I definitely agree. Let's make the naming consistent.

scylla-rust-wrapper/src/argconv.rs
muzarski: The compiler does not even complain if I remove the cfg - probably because this trait is `pub`... As a bonus, I'll change `pub` to `pub(crate)` in `argconv.rs` in a separate commit.

scylla-rust-wrapper/src/argconv.rs
muzarski: You convinced me. Renamed to `into_*`.

scylla-rust-wrapper/src/lib.rs
muzarski: It is not. I'm not sure when and why I introduced this change.

scylla-rust-wrapper/src/argconv.rs
muzarski: > 💭 I'm wondering why the function needs to be `unsafe`. Is the following scenario one valid reason for that?
> 
> ```rust
> let data = ...;
> let arc = Arc::new(data);
> let weak = Arc::downgrade(&arc);
> let ptr = weak_as_ptr(&weak);
> std::mem::drop(arc);
> 
> // Now `ptr` is dangling...
> ```

Correct.

> If I'm correct, maybe let's put this short example in the docstring of this function? Just to convince readers that `unsafe` is needed here.

Ok. Good idea.

scylla-rust-wrapper/src/lib.rs
muzarski: Nvm, it is needed. I forgot that doctests are compiled as a separate crate. Thus, they can access only `pub` items from our crate. I justified this change in the comment in code.

scylla-rust-wrapper/src/argconv.rs
muzarski: > As a bonus, I'll change `pub` to `pub(crate)` in `argconv.rs` in a separate commit.

Nvm, I can't do that because of doctests. The ffi traits need to be `pub`.

muzarski:
v4.1: addressed @wprzytula comments

wprzytula (approved):

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,6 +73,261 @@ macro_rules! make_c_str {
...
+/// Represents immutable pointer.
+pub struct Const;
+impl sealed::Sealed for Const {}
+impl Mutability for Const {}
+
+/// Represents mutable pointer.
+pub struct Mut;
+impl sealed::Sealed for Mut {}
+impl Mutability for Mut {}
```
Lorak-mmk: `Mutability` and `CMutability` are quite similar names. The names of specific structs are also similar.

WDYT about
`Mutability` -> `Ownership`
`Const` -> `Shared`
`Mut` -> `Exclusive`

I think it better reflects what this property is about, and reflects the imo better way to think about rust references.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,6 +73,261 @@ macro_rules! make_c_str {
...
+    /// The reference inherits the lifetime of the pointer.
+    #[allow(clippy::wrong_self_convention)]
+    fn into_ref(self) -> Option<&'a T> {
+        // SAFETY: Thanks to the validity and aliasing ^ mutability guarantees,
+        // we can safely convert the pointer to valid immutable reference with
+        // correct lifetime.
+        unsafe { self.ptr.map(|p| p.as_ref()) }
+    }
+}
+
```
Lorak-mmk: ❓ Why is this `allow` necessary? According to the table at https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention it should be fine.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -71,6 +73,261 @@ macro_rules! make_c_str {
...
+/// comes from [`Box::into_raw()`], user could create a [`CassPtr<'static, T, (Mut,)>`].
+/// `'static` lifetime represents that user is the exclusive **owner** of the pointee, and
+/// is responsible for freeing the memory (e.g. via [`Box::from_raw()`]).
+/// On the other hand, when pointer is created from some immutable reference `&'a T`,
+/// the correct choice of CassPtr would be [`CassPtr<'a, T, (Const,)>`]. It means that
+/// holder of the created pointer **borrows** the pointee (with some lifetime `'a`
+/// inherited from the immutable borrow `&'a T`).
+///
+/// Both [`CassPtr::as_ref()`] and [`CassPtr::as_mut_ref()`] consume the pointer.
+/// At first glance, it seems impossible to obtain multiple immutable reference from one pointer.
```
Lorak-mmk: I see no such methods. You meant `into_ref` and `into_mut_ref`, right?

Lorak-mmk (changes requested):
💯 Really great work! I no longer see any big problems, only left some minor comments.

I'm not a fan of the workaround in cass_future_set_callback, but I don't have any good ideas on how to improve it.

scylla-rust-wrapper/src/argconv.rs
muzarski: Yeah, I forgot to address the docstrings after renaming the methods.

scylla-rust-wrapper/src/argconv.rs
muzarski: It used to be called `as_*` (but @wprzytula eventually convinced me to "into_*"). Forgot to remove the allow.

scylla-rust-wrapper/src/argconv.rs
muzarski: I definitely agree with `Exclusive` and `Shared` names. Not sure about `Ownership`. For me, Ownership property is currently represented via lifetimes (Owned vs Borrowed).

OTOH, `Mutability` name is not perfect either. I'll try to come up with some reasonable name. If I fail to do so, I'll apply your suggestion and rename the trait to `Ownership`.

scylla-rust-wrapper/src/argconv.rs
muzarski: Ref: https://github.com/scylladb/cpp-rust-driver/pull/208#discussion_r2005329145

scylla-rust-wrapper/src/argconv.rs
muzarski: I couldn't come up with a better name. Renamed the trait and corresponding struct as you suggested.

muzarski:
v4.2: Addressed @Lorak-mmk comments.

wprzytula:
Hmm, why did cassandra tests fail?

wprzytula:
The cover letter needs to be adjusted for the recent `Mutability` -> `Ownership` change.

muzarski:
> Hmm, why did cassandra tests fail?

Due to the recent change on scylla-ccm's master. I'll be investivating this, but the failure is not related to this PR.

muzarski:
> > Hmm, why did cassandra tests fail?
> 
> Due to the recent change on scylla-ccm's master. I'll be investivating this, but the failure is not related to this PR.

@wprzytula @Lorak-mmk I opened an issue on scylla-ccm: https://github.com/scylladb/scylla-ccm/issues/646. As a temporary fix, should I set the ccm version in CI to the version before the change that introduced the bug?

Lorak-mmk:
> > > Hmm, why did cassandra tests fail?
> > 
> > Due to the recent change on scylla-ccm's master. I'll be investivating this, but the failure is not related to this PR.
> 
> @wprzytula @Lorak-mmk I opened an issue on scylla-ccm: https://github.com/scylladb/scylla-ccm/issues/646. As a temporary fix, should I set the ccm version in CI to the version before the change that introduced the bug?

Yes

muzarski:
CI passed - I am merging

PR #211, opened 2024-11-28
clippy: fix clippy lints for 1.83

muzarski:
- derive(FromPrimitive) does not work with new clippy. Need to implement it manually We can maybe revert it in the future. However, this change is not a big deal.
- Converted all manually created c-strings to c"" literals
- Removed some unnecessary return statements detected by clippy

## Pre-review checklist


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1070,6 +1070,8 @@ mod tests {
                     );
 
                     // empty strings
+                    // Allow it, since c"" somehow clashes with #[ntest] proc macro...
+                    #[allow(clippy::manual_c_str_literals)]
```
Lorak-mmk, line 1074: Eh, why is ntest so problematic...

Lorak-mmk (approved):
> Converted all manually created c-strings to c"" literals

Love it! But I'm surprised there is that few of them, I thought there was much more. Isn't there more in error stringifying function etc?

muzarski:
> > Converted all manually created c-strings to c"" literals
> 
> Isn't there more in error stringifying function etc?

You are right. I'm not sure why clippy did not detect them. I'll fix those in a separate commit.

muzarski:
v2: replaced all occurrences of `"...\0"` with `c"..."` string literals

scylla-rust-wrapper/src/cluster.rs
muzarski: I opened an issue: https://github.com/becheran/ntest/issues/30

wprzytula (approved):

Lorak-mmk (approved):

PR #213, opened 2024-12-03
deser: lazy CassRow construction

muzarski:
Depends on: https://github.com/scylladb/cpp-rust-driver/pull/208

## Motivation
Currently, we unnecessarily construct `CassRow`s eagerly. What we can do (and what cpp-driver does) is to deserialize rows lazily, during iteration using iterator created via `cass_iterator_from_result`.

Why?
Notice that `CassResult` has no methods that request a row at given index. The only exception is `cass_result_first_row` - but we can simply deserialize only first row, and cache it in `CassResult` so the call to `cass_result_first_row` is always fast.

## Changes
- Adjusted `CassRowsResult`, so now it only holds deserialized first row (instead of all rows), and payload required for deserialization (i.e. `DeserializedMetadataAndRawRows` and `CassResultMetadata`).
- Adjusted `CassResultIterator`, so now it is an enum - to distinguish whether it's an iterator created from either non-rows result or rows result.
- Introduced `CassRowsResultIterator,` which now uses `TypedRowIterator` under the hood. The rows are now deserialized on the fly, during `cass_iterator_next` call.
- Adjusted `CassRow`, and newly introduced `CassRowsResultIterator`, so they now borrow `CassResultMetadata` from `CassResult` (instead of sharing ownership via `Arc`). Their lifetime is already bound by the lifetime of `CassResult`, so there is no point in cloning an Arc.

## Pre-review checklist


muzarski:
This one is ready for review. https://github.com/scylladb/cpp-rust-driver/pull/208 needs to be merged first, though.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -265,17 +271,13 @@ pub unsafe extern "C" fn cass_iterator_get_row<'result>(
     let iter = BoxFFI::as_ref(iterator).unwrap();
 
     // Defined only for result iterator, for other types should return null
-    if let CassIterator::Result(result_iterator) = iter {
-        let iter_position = match result_iterator.position {
+    if let CassIterator::Result(CassResultIterator::Rows(rows_result_iterator)) = iter {
+        let iter_position = match rows_result_iterator.position {
```
wprzytula: :recycle: I suggest replacing this `if let` with `let else` for better code flow.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -167,13 +172,25 @@ pub unsafe extern "C" fn cass_iterator_next(
...
+                })
+                .map(|row| {
+                    CassRow::from_row_and_metadata(row, &rows_result_iterator.result_metadata)
+                });
+
+            rows_result_iterator.current_row = new_row;
+
+            rows_result_iterator.current_row.is_some() as cass_bool_t
         }
         CassIterator::Row(row_iterator) => {
```
wprzytula: :camping: I believe this method should only delegate to `next()` implementations for all of: `CassRowIterator`, `CassResultIterator`, etc. Those implementations should be extracted from this single method into separate methods; this would aid readability.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -266,23 +283,17 @@ pub unsafe extern "C" fn cass_iterator_next(
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_iterator_get_row<'result>(
-    iterator: CassBorrowedSharedPtr<CassIterator<'result>, CConst>,
+    iterator: CassBorrowedSharedPtr<'result, CassIterator<'result>, CConst>,
```
wprzytula, line 303: :thinking: Hmm, why does it syntactically type check? Before this commit, `CassBorrowedSharedPtr` had no lifetime specified and now it takes one. Is it allowed to omit lifetime arguments?

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -266,23 +283,17 @@ pub unsafe extern "C" fn cass_iterator_next(
...
-
-        return RefFFI::as_ptr(row);
+        return rows_result_iterator
+            .current_row
+            .as_ref()
+            .map(RefFFI::as_ptr)
+            .unwrap_or(RefFFI::null());
     }
 
     RefFFI::null()
```
wprzytula: :recycle: `let-else` for better logic linearity.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -9,17 +12,19 @@ use crate::metadata::{
...
 use std::os::raw::c_char;
+use std::sync::Arc;
 
 pub struct CassRowsResultIterator<'result> {
-    result: &'result CassRowsResult,
-    position: Option<usize>,
+    iterator: TypedRowIterator<'result, 'result, Row>,
+    result_metadata: Arc<CassResultMetadata>,
+    current_row: Option<CassRow>,
 }
```
wprzytula: :thinking: As the whole iterator is parametrized by a lifetime `'result`, why do we need to have a shared ownership over `CassResultMetadata`? Can't we borrow it instead from the same source that we borrow raw frame contents from?

wprzytula:
```rust
pub struct CassRowsResultIterator<'frame, 'metadata> {
    iterator: TypedRowIterator<'frame, 'metadata, Row>,
    result_metadata: &'metadata CassResultMetadata,
    current_row: Option<CassRow>,
}
```

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -167,13 +172,25 @@ pub unsafe extern "C" fn cass_iterator_next(
...
+                .iterator
+                .next()
+                .and_then(|res| match res {
+                    Ok(row) => Some(row),
+                    Err(e) => {
+                        // We have no way to propagate the error (return type is bool).
+                        // Let's at least log the deserialization error.
+                        tracing::error!("Failed to deserialize next row: {e}");
+                        None
+                    }
```
wprzytula: :disappointed: So sad that the API is so bad...

scylla-rust-wrapper/src/query_error.rs
```diff
@@ -15,7 +15,7 @@ pub enum CassErrorResult {
     Query(#[from] ExecutionError),
     #[error(transparent)]
     ResultMetadataLazyDeserialization(#[from] ResultMetadataAndRowsCountParseError),
-    #[error("Failed to deserialize rows: {0}")]
+    #[error("Failed to deserialize first row: {0}")]
     Deserialization(#[from] DeserializationError),
```
wprzytula: :camping: Shall we rename this module to `execution_error`, together with renaming `Query` to `Statement`?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -55,21 +57,21 @@ impl CassResult {
...
                 let cass_result = CassResult {
-                    tracing_id: rows_result.tracing_id(),
+                    tracing_id,
                     paging_state_response,
                     kind: CassResultKind::Rows(CassRowsResult {
-                        rows: cass_rows,
+                        raw_rows,
+                        first_row,
                         metadata,
                     }),
```
wprzytula: :question: Just to make sure it's supposed to work: we're creating a `CassRowsResult` with `raw_rows` still containing the first row, even though the first row is already deserialized. Won't this make the first row deserialized/returned twice?

scylla-rust-wrapper/src/iterator.rs
wprzytula: Or, if we need an `Arc` to be available for cloning due to some APIs, then we could maybe store `&'metadata Arc<ResultMetadata>`?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -26,10 +26,30 @@ pub enum CassResultKind {
...
+        raw_rows: &DeserializedMetadataAndRawRows,
+        metadata: &CassResultMetadata,
+    ) -> Result<Option<CassRow<'static>>, CassErrorResult> {
+        let first_row = raw_rows
+            .rows_iter::<Row>()
+            // unwrap: Row always passes the typecheck.
+            .unwrap()
+            .next()
+            .transpose()?
+            .map(|row| std::mem::transmute(CassRow::from_row_and_metadata(row, metadata)));
```
wprzytula: :wrench: `mem::transmute` is a particularly wounding footgun. IIUC, the only thing you achieve by using it is lifetime extension (from `'a` to `'static`). Let's extract a local function that does only that (encapsulating and limiting the destructive potential of `mem::transmute`) and call it.

```suggestion
        fn extend_lifetime<T>(v: &'_ T) -> &'static T {
            std::mem::transmute(v)
        }
        
        let first_row = raw_rows
            .rows_iter::<Row>()
            // unwrap: Row always passes the typecheck.
            .unwrap()
            .next()
            .transpose()?
            .map(|row| extend_lifetime(CassRow::from_row_and_metadata(row, metadata)));
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -58,13 +78,9 @@ impl CassResult {
...
                 let (raw_rows, tracing_id, _) = rows_result.into_inner();
-                let first_row = raw_rows
-                    .rows_iter::<Row>()
-                    // unwrap: Row always passes the typecheck.
-                    .unwrap()
-                    .next()
-                    .transpose()?
-                    .map(|row| CassRow::from_row_and_metadata(row, &metadata));
+                // SAFETY: We make sure that `CassRow` will live no longer than `CassRowsResult` that it borrows from.
+                // This is becase, first_row is a part of constructed CassRowsResult.
```
wprzytula: :wrench:

```suggestion
                // SAFETY: We make sure that `CassRow` will live no longer than `CassRowsResult` that it borrows from.
                // This is because `first_row` is a part of constructed CassRowsResult.
```

wprzytula: :memo:

> The fields of a [struct](https://doc.rust-lang.org/reference/types/struct.html) are dropped in declaration order.

`metadata` is declared below `first_row`, so `first_row` will indeed live no longer than `metadata` :+1:.

scylla-rust-wrapper/.cargo/config.toml
Lorak-mmk, line 1: Commit: cargo/cmake: compile with cpp_rust_unstable cfg 

Could you explain in the commit message why are changes on both files necessary? It is not clear to me why changes in config.toml are not enough.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -648,10 +659,14 @@ pub unsafe extern "C" fn cass_iterator_from_result<'result>(
...
-            position: None,
-        }),
+        CassResultKind::Rows(cass_rows_result) => {
+            CassResultIterator::Rows(CassRowsResultIterator {
+                // unwrap: Row always passes the typecheck.
+                iterator: cass_rows_result.raw_rows.rows_iter().unwrap(),
+                result_metadata: Arc::clone(&cass_rows_result.metadata),
+                current_row: None,
+            })
+        }
```
Lorak-mmk: Nit: For functions like `rows_iter` it is imo more readable to always specify the type (`.rows_iter::<Row>()`)

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -26,10 +26,30 @@ pub enum CassResultKind {
...
+            // unwrap: Row always passes the typecheck.
+            .unwrap()
+            .next()
+            .transpose()?
+            .map(|row| std::mem::transmute(CassRow::from_row_and_metadata(row, metadata)));
+
+        Ok(first_row)
+    }
+}
+
```
Lorak-mmk: Commit: "row: borrow metadata from result, instead of cloning an Arc "

`CassRowsResult` borrows from `CassResultMetadata`, not `CassRowsResult`, so the safety comment here seems to be imprecise.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -58,13 +78,9 @@ impl CassResult {
...
-                let first_row = raw_rows
-                    .rows_iter::<Row>()
-                    // unwrap: Row always passes the typecheck.
-                    .unwrap()
-                    .next()
-                    .transpose()?
-                    .map(|row| CassRow::from_row_and_metadata(row, &metadata));
+                // SAFETY: We make sure that `CassRow` will live no longer than `CassRowsResult` that it borrows from.
+                // This is becase, first_row is a part of constructed CassRowsResult.
+                let first_row = unsafe { CassRowsResult::create_first_row(&raw_rows, &metadata)? };
```
Lorak-mmk: This comment also fails to mention `CassResultMetadata`.

Why is it important? Apart from liveness there is one more important requirement: You are not allowed to move the borrowed `CassResultMetadata`, even if it keeps living.

Until one notices that it is stored behind and Arc here it is not obvious that it is not moved during, or after, creation of `CassRowsResult`

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk: Also I think we should use `unsafe` blocks for unsafe operations even in unsafe functions. 
The lint is warn-by-default in 2024 edition of Rust, which we can migrate to. We could also enable it even before migrating.

Lorak-mmk (changes requested):

scylla-rust-wrapper/.cargo/config.toml
muzarski: From my experience, the `CMakeRust` does not pick up the config file. This might be related: https://github.com/Devolutions/CMakeRust/issues/14.

muzarski: I'll adjust the cmake files, so they pick up the config file. Thanks to that, in the future, if we modify the `.cargo/config.toml` we won't need to adjust the cmake files.

Lorak-mmk: Ok, please explain it in the commit message.

muzarski: > I'll adjust the cmake files, so they pick up the config file. Thanks to that, in the future, if we modify the `.cargo/config.toml` we won't need to adjust the cmake files.

It's annoying, but I won't be able to do that. According to the documentation (https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags), the `RUSTFLAGS` env variable overrides the rustflags defined in the config file. Cmake sets some RUSTFLAGS. At first, I thought that I might be able to add these to `config.toml`, but the RUSTFLAGS defined by the cmake are not static - they are dependent on the platform.

So I'll leave the current version and add a comment explaining why it has to be like that.

Lorak-mmk: Ok. Out of curiosity: what RUSTFLAGS are set there? It is a bit surprising that we need any (other than cpp_rust_unstable).

muzarski: On my machine these are: `RUSTFLAGS=-Clink-arg=-Wl,-soname=libscylla-cpp-driver.so.2`

These look like some linker options

scylla-rust-wrapper/src/query_result.rs
muzarski: > Also I think we should use `unsafe` blocks for unsafe operations even in unsafe functions. The lint is warn-by-default in 2024 edition of Rust, which we can migrate to. We could also enable it even before migrating.

Opened a PR: https://github.com/scylladb/cpp-rust-driver/pull/232

muzarski: > Commit: "row: borrow metadata from result, instead of cloning an Arc "
> 
> `CassRowsResult` borrows from `CassResultMetadata`, not `CassRowsResult`, so the safety comment here seems to be imprecise.

What I had in mind, is that `CassRow` borrows the metadata (`CassResultMetadata`) which is owned by `CassRowsResult`. But I agree, that the safety comment is not perfect. I'll adjust it.

scylla-rust-wrapper/src/query_result.rs
muzarski: Good catch. I took for granted the fact that most of (if not all) the objects created by our API are heap-allocated and cannot be moved out of the allocated memory unless free'd.

muzarski:
Rebased on master (to include https://github.com/scylladb/cpp-rust-driver/pull/232)

muzarski:
v1.1: Adjusted safety comments and commit messages as @Lorak-mmk suggested.

The CI will fail - still waiting for https://github.com/scylladb/cpp-rust-driver/pull/233 to finish (cmake version bump)

muzarski:
WTF, `Build / Build, lint and run unit tests` passed. It looks like not all GH runners have CMake updated to 4.0 yet.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -26,10 +26,35 @@ pub enum CassResultKind {
...
+    // CassRow borrows `metadata` field.
+    pub first_row: Option<CassRow<'static>>,
     pub metadata: Arc<CassResultMetadata>,
 }
 
+impl CassRowsResult {
+    /// SAFETY:
+    /// Caller needs to ensure that `metadata`:
+    /// 1. outlives the created row
+    /// 2. is not moved while the row lives (will not change its place in memory)
```
Lorak-mmk:
```
3. No mutable reference to this `CassResultMetadata` is created during the lifetime of returned CassRow.
In other words - Rust borrowing rules are upheld.
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -58,13 +83,11 @@ impl CassResult {
...
-                    // unwrap: Row always passes the typecheck.
-                    .unwrap()
-                    .next()
-                    .transpose()?
-                    .map(|row| CassRow::from_row_and_metadata(row, &metadata));
+                // SAFETY:
+                // 1. created row will live no longer than metadata, because they are both owned by `CassRowsResult`.
+                // 2. metadata object will not change its place in memory,
+                //    because it's obtained from an Arc allocation. Arc will not be freed as long as `CassRowsResult` lives.
+                let first_row = unsafe { CassRowsResult::create_first_row(&raw_rows, &metadata)? };
```
Lorak-mmk: And we do not create mut refs to this metadata, right?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/query_result.rs
muzarski: Correct, `CassResultMetadata` is immutable.

muzarski:
v1.2: Rebased on master (cmake version bump) and addressed @Lorak-mmk comments

Lorak-mmk (approved):

wprzytula (changes requested):

Lorak-mmk:
Regarding the self-reference problem: maybe we could use `yoke` as we did in Rust Driver?

scylla-rust-wrapper/src/query_result.rs
muzarski: If someone creates the `cass_iterator_from_result` which iterates over `CassRow`s, it is expected that the first row is still there. So you are right, the first row will be deserialized again during iteration (first call to `cass_iterator_next`).

scylla-rust-wrapper/src/iterator.rs
muzarski: It is addressed in last commit. `CassRowsResultIterator` looks exactly as you suggested.

scylla-rust-wrapper/src/query_result.rs
wprzytula: > If someone creates the `cass_iterator_from_result` which iterates over `CassRow`s, it is expected that the first row is still there. So you are right, the first row will be deserialized again during iteration (first call to `cass_iterator_next`).

Can we somehow skip the second deserialization, as we already have the first row deserialized?

muzarski: I don't think it's possible with current API. `TypedRowIterator` performs deserialization in `next()`. Even if we wanted to skip the first row, we would need to call `next()`. So in the end, we would still need to deserialize the row.

Maybe if `TypedRowIterator` was `Clone` (but it's not), then we could:
1. create TypedRowIterator
2. deserialize first row (call `next()` on the iterator)
3. store the iterator without the first row in `CassRowsResult` (self-borrowing yet again)
4. during the call to `cass_iterator_from_result`, we could pass the first row (deserialized) and iterator (without first row) to the `CassRowsResultIterator` by cloning.

I don't believe that such micro-optimizations are worth introducing at the cost of code simplicity and readability.

Notice that I am working towards the lazy deserialization, so there won't be any allocations during the deserialization except the allocation for vector of values, which is necessary

Lorak-mmk: Side question: currently you are always deserializing to `Row`, right? Is that the final approach? I don't thing it will allow implementing `cass_value_get_bytes`, for that you need full lazy deserialization.

muzarski: It's not the final approach. The follow-up PR will have two new structs defined:
```rust
pub struct CassRawValue<'result> {
    pub typ: &'result ColumnType<'result>,
    pub slice: Option<FrameSlice<'result>>,
}
impl DeserializeValue for CassRawValue { ... }

pub struct CassRawRow<'result> {
    columns: Vec<CassRawValue<'result>>,
}
impl DeserializeRow for CassRawRow { ... }
```

The `CassRawValue` is going to be stored directly in `CassValue`. Let's say someone calls `cass_value_get_float`, then we can easily deserialize the value using `f32 as DeserializeValue` because we store the `FrameSlice` and `ColumnType`. It also allows us to implement `cass_value_get_bytes` (because we store the serialized data).

muzarski: So in the end, we replace `Row` with `CassRawRow` and `CqlValue` with `CassRawValue`.

muzarski:
> Regarding the self-reference problem: maybe we could use `yoke` as we did in Rust Driver?

I considered this. I'd need to investigate how easy it would be to implement it. I'm not that familiar with `yoke` traits, but I always have the rust-driver deserialization code as a reference.

scylla-rust-wrapper/src/iterator.rs
muzarski: It's not possible. The compiler complains that it does not know which of the two `iterator` lifetimes to choose.

`CassRowsResultIterator` before this commit:
```rust
pub struct CassRowsResultIterator<'result> {
    result: &'result CassRowsResult,
    position: Option<usize>,
}
```
I'll try to explain in pseudo code what happened before this commit.
```rust
fn cass_iterator_get_row<'ptr, 'result>(
    iterator: CassPtr<'ptr, CassIterator<'result>>
) -> CassPtr<'result, CassRow> {
     let iter: &'ptr CassIterator<'result> = BoxFFI::as_ref(iterator);
     // ... Some manipulation to obtain the reference to `CassRow`. Assume that we have get_row() method which returns the reference that iterator holds.
     // The lifetime is 'result, because iterator holds &'result CassRowsResult
     let row: &'result CassRow = iter.get_row(); 
     let result_ptr: CassPtr<'result, CassRow> = RefFFI::as_ptr(row);
     result_ptr
}
```
So in this case, the lifetime of the pointer `'ptr` is arbitrary. No constraints are needed.


`CassRowsResultIterator` after this commit:
```rust
pub struct CassRowsResultIterator<'result> {
    iterator: TypedRowIterator<'result, 'result, Row>,
    result_metadata: Arc<CassResultMetadata>,
    current_row: Option<CassRow>,
}
```

pseudo-code with lifetimes after this commit:
```rust
fn cass_iterator_get_row<'ptr, 'result>(
    iterator: CassPtr<'ptr, CassIterator<'result>>
) -> CassPtr<'result, CassRow> {
     let iter: &'ptr CassIterator<'result> = BoxFFI::as_ref(iterator);
     // Let's ignore the fact that this is an Option.
     // The lifetime is `'ptr` because the iterator OWNS this row. The lifetime of 'iter' is 'ptr
     let row: &'ptr CassRow = &iter.current_row; 
     // The pointer inherits the lifetime of the reference
     let result_ptr: CassPtr<'ptr, CassRow> = RefFFI::as_ptr(row);
     result_ptr

    // Oh no! But we asked for CassPtr<'result, CassRow>, and not CassPtr<'ptr, CassRow>.
    // The constraint 'ptr: 'result is required!
}
```

So the difference is that in the first case, the iterator **borrows** the `CassRow` with some lifetime `'result`, while in the second case, the iterator **owns** the `CassRow`.

scylla-rust-wrapper/src/iterator.rs
muzarski: Done in a separate commit.

scylla-rust-wrapper/src/query_error.rs
muzarski: Done in a separate commit

muzarski:
v2.0:
- implemented self-borrowing using `yoke`, as @Lorak-mmk suggested

Lorak-mmk (approved):

scylla-rust-wrapper/src/iterator.rs
wprzytula: Thanks for so detailed explanation! I followed it carefully and understood a lot about practical implications of our new FFI model around lifetime-parameterised `CassPtr`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -121,20 +121,87 @@ impl CassResultMetadata {
...
+        }
+
+        #[cfg(test)]
+        pub fn mock_from_row_and_metadata(row: Row, metadata: Arc<CassResultMetadata>) -> Self {
+            let yoke = Yoke::attach_to_cart(metadata, |metadata_ref| {
+                CassRowWrapper(CassRow::from_row_and_metadata(row, metadata_ref))
+            });
+
+            Self(yoke)
         }
```
wprzytula: :pick: I'd use this one in `new_from_raw_rows_and_metadata` for two reasons:
1. Not to duplicate the code around `Yoke`'s usage.
2. To separate and encapsulate `Yoke`-related code.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -121,20 +121,87 @@ impl CassResultMetadata {
...
+                .unwrap()
+                .next()
+                .transpose()?
+                .map(|row: Row| {
+                    let yoke = Yoke::attach_to_cart(metadata, |metadata_ref| {
+                        CassRowWrapper(CassRow::from_row_and_metadata(row, metadata_ref))
+                    });
+
+                    Self(yoke)
+                });
```
wprzytula: :100: Good job with employing `Yoke`!

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -2,7 +2,7 @@ use scylla::errors::*;
 
 // Re-export error types.
 pub(crate) use crate::cass_error_types::{CassError, CassErrorSource};
-use crate::query_error::CassErrorResult;
+use crate::execution_error::CassErrorResult;
```
wprzytula, line 5: :wrench:
`CassErrorResult::Query` variant still needs rename. Same for variables: `query_error` -> `execution_error`.

wprzytula (changes requested):

muzarski:
v2.1: Addressed @wprzytula comments

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -119,25 +121,82 @@ impl CassResultMetadata {
...
+            let row = raw_rows
+                .rows_iter::<Row>()
+                // unwrap: Row always passes the typecheck.
+                .unwrap()
+                .next()
+                .transpose()?
+                .map(|row: Row| Self::new_from_row_and_metadata(row, metadata));
+
+            Ok(row)
+        }
```
wprzytula: :wrench: IIUC, this constructor creates the `RowWithSelfBorrowedMetadata` based on **the first row** of `DeserializedMetadataAndRawRows`. I believe this property is nonobvious and thus should be highlighted in the constructor's name, or at least in its docstring.

wprzytula (changes requested):

muzarski:
v2.2: Renamed the constructor so it indicates that it constructs first row only.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -119,25 +121,83 @@ impl CassResultMetadata {
...
+
+        pub fn new_from_row_and_metadata(row: Row, metadata: Arc<CassResultMetadata>) -> Self {
+            let yoke = Yoke::attach_to_cart(metadata, |metadata_ref| {
+                CassRowWrapper(CassRow::from_row_and_metadata(row, metadata_ref))
+            });
+
+            Self(yoke)
+        }
+    }
```
wprzytula: :question: Does this need to be `pub`? Isn't `pub(self)` enough?

wprzytula (approved):
LGTM

scylla-rust-wrapper/src/query_result.rs
muzarski: No - apart from the internal usage, the test is using it as well.

wprzytula: I thought (perhaps wrongly) that the test is put in the same module (or in a submodule).

muzarski: The structure is following:
```rust
mod query_result {
    mod row_with_self_borrowed_metadata {
        impl RowWithSelfBorrowedMetadata { ... }
    }

    mod tests {
         use super::row_with_self_borrowed_metadata::RowWithSelfBorrowedMetadata;
         ...
    }
}
```

The method cannot be `pub(self)`. But, it could be `pub(super)`. I'll do it.

muzarski:
v2.3: Reduced visibility of `RowWithSelfBorrowedMetadata` methods to `pub(super)`.

wprzytula (approved):

Lorak-mmk (approved):

PR #221, opened 2025-02-14
Make sure all queries to system.local uses WHERE clause

### LINKED ISSUE #220: Make sure driver uses `where key='local'` for every `select` from `system.local`
Go over driver code and make sure there is no full scan and every query is done with `where key='local'` for `select`s from `system.local`

dkropachev:
Full scan requests are executed slower even if there is only one record

Fixes: https://github.com/scylladb/cpp-rust-driver/issues/220
## Pre-review checklist
- [ ] ~~I have enabled appropriate tests in `.github/workflows/build.yml` in `gtest_filter`.~~
- [ ] ~~I have enabled appropriate tests in `.github/workflows/cassandra.yml` in `gtest_filter`.~~


Lorak-mmk:
Why do we need this? It's only tests / examples.

dkropachev:
> Why do we need this? It's only tests / examples.

Sure, why not fix them too?

wprzytula:
> Why do we need this? It's only tests / examples.

It makes sense to have proper use of ScyllaDB in all places, including tests and especially examples.

muzarski (approved):

Issue #133, opened 2024-07-02
Make `QueryError` -> `CassError` conversion more accurate


wprzytula:
## Problem

At the moment, the error conversion is much limited due to stringified errors in Rust driver.
The biggest offender in my opinion is:
```rust
    QueryError::InvalidMessage(_string) => CassError::CASS_ERROR_SERVER_INVALID_QUERY,
```
which is totally inaccurate, because `SERVER_INVALID_QUERY` strictly corresponds to `DbError::Invalid`. At the same time, `QueryError::InvalidMessage` is abused so much by converting all `ParseError`s and `FrameError`s to it.

## Solution

Once the errors refactor lands in the Rust driver, a complex review of existing error conversion should be done to make it more accurate. Or else users moving from cpp-driver to cpp-rust-driver will get much confused by error messages. After all, failure deserializing a row may be displayed at the moment as "server invalid query".

wprzytula:
Another finding:
`NewSessionError::ProtocolError` is converted to a different error than `QueryError::ProtocolError`.
```rust
    NewSessionError::ProtocolError(_str) => {
        CassError::CASS_ERROR_LIB_UNABLE_TO_DETERMINE_PROTOCOL
    }
```
```rust
    DbError::ProtocolError => CassError::CASS_ERROR_SERVER_PROTOCOL_ERROR,
```
This is a mess.

Lorak-mmk:
Yeah, when writing the initial version during the hachathon we didn't put much care in error conversions. Getting something that mostly works was more important.
After error refactor in Rust Driver is finished will be a good time to review the conversions to see which errors cpp-driver returns under which conditions and convert accordingly.

PR #223, opened 2025-03-05
cargo: bump scylla dependency to 1.0.0

muzarski:
First two commits are related to bumping the dependency versions. Rust-driver that needed to be addressed:
- errors refactor - some variants were moved, some new error types were introduced. I tried to adjust the rust-driver to cpp-driver errors conversion as accurate as I could. However, this still needs to be addressed later. There is an issue: https://github.com/scylladb/cpp-rust-driver/issues/133
- modules refactor - some paths needed to be adjusted on cpp-rust-driver side
- &[ColumnSpec] vs ColumnSpecs
- Unification of CqlType and ColumnType

## &[ColumnSpec] vs ColumnSpecs
In later commits, we remove the CassResultMetadata constructor from &[ColumnSpec] in favor of ColumnSpecs.

## ColumnType vs CqlType
Remaining commits improve/simplify the code related to constructing CassDataType from ColumnType. There were two methods: `get_column_type()` (converts ColumnType to CassDataType) and `get_column_type_from_cql_type()` (used to convert CqlType to CassDataType). As of now, having these two methods is redundant.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -10,8 +10,8 @@ categories = ["database"]
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v0.15.0", features = [
-    "ssl",
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.0.0", features = [
+    "openssl-010",
 ] }
 tokio = { version = "1.27.0", features = ["full"] }
 uuid = "1.1.2"
@@ -32,7 +32,7 @@ bindgen = "0.65"
 chrono = "0.4.20"
 
 [dev-dependencies]
-scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v0.15.0" }
+scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.0.0" }
 
 assert_matches = "1.5.0"
 ntest = "0.9.3"
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -1,4 +1,4 @@
-use scylla::transport::errors::*;
+use scylla::errors::*;
 
 // Re-export error types.
 pub(crate) use crate::cass_error_types::{CassError, CassErrorSource};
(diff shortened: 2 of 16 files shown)


scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -27,30 +27,84 @@ impl ToCassError for CassErrorResult {
...
+            RequestAttemptError::UnableToAllocStreamId => CassError::CASS_ERROR_LIB_NO_STREAMS,
+            RequestAttemptError::BrokenConnectionError(_) => {
+                CassError::CASS_ERROR_LIB_UNABLE_TO_CONNECT
+            }
+            RequestAttemptError::BodyExtensionsParseError(_) => {
+                CassError::CASS_ERROR_LIB_MESSAGE_ENCODE
+            }
+            RequestAttemptError::CqlResultParseError(_) => {
+                CassError::CASS_ERROR_LIB_UNEXPECTED_RESPONSE
+            }
```
wprzytula, line 89: :rocket: :100:  Wonderful! The error refactor has just proven to be very helpful for cpp-rust-driver error parity with cpp-driver.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -82,27 +136,18 @@ impl ToCassError for DbError {
...
-            BadQuery::BadKeyspaceName(_bad_keyspace_name) => CassError::CASS_ERROR_LAST_ENTRY,
-            BadQuery::Other(_other_query) => CassError::CASS_ERROR_LAST_ENTRY,
-            BadQuery::SerializationError(e) => {
-                if e.downcast_ref::<UnknownNamedParameterError>().is_some() {
-                    // It means that our custom `UnknownNamedParameterError` was returned.
-                    CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
-                } else {
-                    CassError::CASS_ERROR_LAST_ENTRY
-                }
-            }
```
wprzytula, line 105: :question: Is this logic gone or am I missing something?

wprzytula: Indeed, it's present below. Nvm. Leaving this for other reviewers.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -170,7 +207,13 @@ impl CassErrorMessage for CassErrorResult {
     }
 }
 
-impl CassErrorMessage for QueryError {
+impl CassErrorMessage for ExecutionError {
+    fn msg(&self) -> String {
+        self.to_string()
+    }
+}
```
wprzytula, line 214: :question: Is `CassErrorMessage::msg` anywhere different from `Display` impl for the given type?

:thought_balloon: :pushpin: If not, maybe let's just require `Display` as a supertrait and provide default method impl that calls `ToString::to_string`?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -374,18 +380,18 @@ pub fn get_column_type_from_cql_type(
...
+            let name = &definition.name;
             CassDataTypeInner::UDT(UDTDataType::create_with_params(
                 user_defined_types,
                 keyspace_name,
                 name,
                 *frozen,
             ))
         }
+
+        // Column type is non_ehxaustive.
```
wprzytula: :pick: typo: `ehxaustive`

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -487,43 +493,78 @@ impl CassDataTypeInner {
...
-        ColumnType::Float => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_FLOAT),
-        ColumnType::Int => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_INT),
-        ColumnType::BigInt => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_BIGINT),
-        ColumnType::Text => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_TEXT),
-        ColumnType::Timestamp => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_TIMESTAMP),
-        ColumnType::Inet => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_INET),
-        ColumnType::Duration => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_DURATION),
-        ColumnType::List(boxed_type) => CassDataTypeInner::List {
+        ColumnType::Native(Ascii) => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_ASCII),
+        ColumnType::Native(Boolean) => {
```
wprzytula: :recycle: If you already glob import other enums' variants, why not import `ColumnType`'s, too? This will reduce the (quite massive) amount of code bloat here.

wprzytula: Same with `CassValueType`.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -532,21 +573,18 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
...
         ColumnType::Tuple(v) => CassDataTypeInner::Tuple(
             v.iter()
                 .map(|col_type| Arc::new(get_column_type(col_type)))
                 .collect(),
         ),
-        ColumnType::Uuid => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_UUID),
-        ColumnType::Varint => CassDataTypeInner::Value(CassValueType::CASS_VALUE_TYPE_VARINT),
+
+        // ColumnType is non_exhaustive.
+        _ => panic!("Unknown column type"),
```
wprzytula: :warning: :thinking: Is it reasonable to make cpp-rust-driver panic on an unknown type? Addition of a new type in Rust driver can be skipped upon bumping its version in cpp-rust-driver; then the Rust driver may return a type that is unknown here and OOOOPS...

scylla-rust-wrapper/src/query_error.rs
```diff
@@ -72,26 +72,26 @@ pub unsafe extern "C" fn cass_error_result_consistency(
...
+                }
+                RequestAttemptError::DbError(DbError::WriteTimeout { consistency, .. }, _) => {
+                    CassConsistency::from(*consistency)
+                }
+                RequestAttemptError::DbError(DbError::ReadFailure { consistency, .. }, _) => {
+                    CassConsistency::from(*consistency)
+                }
+                RequestAttemptError::DbError(DbError::WriteFailure { consistency, .. }, _) => {
+                    CassConsistency::from(*consistency)
+                }
```
wprzytula: :recycle: Maybe let's use `|` to group all variants that we extract `consistency` from, to unify all such branches?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -1645,17 +1621,19 @@ mod tests {
...
-            col_spec(
-                THIRD_COLUMN_NAME,
-                ColumnType::Collection {
-                    frozen: false,
-                    typ: CollectionType::List(Box::new(ColumnType::Native(NativeType::Double))),
-                },
-            ),
-        ]));
+        let metadata = Arc::new(CassResultMetadata::from_column_spec_views(
+            ColumnSpecs::new(&[
```
wprzytula: :wrench: I'd keep the constructor's name as `from_column_specs`, as there are no longer any `column_spec_views`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -103,7 +103,7 @@ pub struct CassResultMetadata {
 }
 
 impl CassResultMetadata {
-    pub fn from_column_spec_views(col_specs: ColumnSpecs<'_, '_>) -> CassResultMetadata {
+    pub fn from_column_specs(col_specs: ColumnSpecs<'_, '_>) -> CassResultMetadata {
```
wprzytula, line 106: Oh OK, it's been addressed in the next commit.

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -295,33 +295,31 @@ impl CassDataType {
...
-            NativeType::Time => CassValueType::CASS_VALUE_TYPE_TIME,
-            NativeType::Timeuuid => CassValueType::CASS_VALUE_TYPE_TIMEUUID,
-            NativeType::Uuid => CassValueType::CASS_VALUE_TYPE_UUID,
-            NativeType::Varint => CassValueType::CASS_VALUE_TYPE_VARINT,
-
-            // NativeType is non_exhaustive
-            _ => CassValueType::CASS_VALUE_TYPE_UNKNOWN,
-        }
+fn native_type_to_cass_value_type(native_type: &NativeType) -> CassValueType {
+    match native_type {
```
wprzytula: :100: I very much agree with this commit's far-sighted considerations. Good job!
```
 cass_types: replace From<NativeType> for CassValueType

From<> impl takes an owned NativeType. Because of that, we need to
clone the NativeType when we construct CassValueType (because we only
have access to the reference).

As of now, the clone() is not a problem, because NativeType could easily
derive Copy. It may not be the case in the future, so let's protect ourselves
from implicit, potentially expensive cloning.
```

wprzytula (changes requested):

scylla-rust-wrapper/src/cass_error.rs
muzarski: Yes, I extracted it to `impl ToCassError for SerializationError`.

scylla-rust-wrapper/src/cass_error.rs
muzarski: True! But it pains me that a lot of rust-driver error types/variants do not have their C API equivalent. I'm really leaning towards introducing new error variants in [cassandra.h](https://github.com/scylladb/cpp-rust-driver/blob/master/include/cassandra.h#L652-L712). This may done as a part of #133

scylla-rust-wrapper/src/cass_types.rs
muzarski: True. I think it's much better to return some error here. The caller can then handle it appropriately.

I'll adjust it in a separate commit(s).

scylla-rust-wrapper/src/cass_error.rs
Lorak-mmk: What does cpp-driver return in the cases where we return such errors that have no equivalent?

muzarski: I'd need to investigate.

scylla-rust-wrapper/src/cass_types.rs
muzarski: `CassValueType` is a struct (generated from bindings). It's not possible to `use CassValueType::*`.

scylla-rust-wrapper/src/cass_error.rs
muzarski: As of now what you are saying is true. I'm not sure, but maybe in the future we may want some error to implement `msg()` in other way. I think it's fine to leave it as is for now.

scylla-rust-wrapper/src/cass_types.rs
muzarski: Ok, I started doing that and error handling becomes quite messy and complicated (especially when constructing cluster metadata structures). I'm not sure if this should be a part of this PR. I'll open an issue about it.

muzarski:
v1.1:
- removed `unsafe` from `create_table_metadata` function

scylla-rust-wrapper/src/cass_types.rs
wprzytula: If you prefer not to clutter this PR with that work, let's just return a generic error kind. It's still much better than having this merged with a panic here.

muzarski: Right after I turned off my laptop yesterday, I got a realization - we can simply return CassValueType::CASS_VALUE_TYPE_UNKNOWN...

I'll do it this way.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -108,7 +108,7 @@ impl BoundPreparedStatement {
 
 #[derive(Clone)]
 pub struct BoundSimpleQuery {
-    pub query: Query,
+    pub query: Statement,
```
wprzytula, line 111: :thought_balloon: :pushpin: Shall we update the naming `query` -> `statement` in cpp-rust-driver, too?

wprzytula (approved):

scylla-rust-wrapper/src/statement.rs
muzarski: Yeah, I plan to do that. It's not crucial, though. The most important is naming of public API which we (almost) fully inherited from the cpp-driver.

muzarski: I opened a P3 issue https://github.com/scylladb/cpp-rust-driver/issues/224

muzarski:
v1.2: Returning `CassValueType::CASS_VALUE_TYPE_UNKNOWN` instead of panicking when matching against `_` (ColumnType is non_exhaustive).

wprzytula (approved):

scylla-rust-wrapper/Cargo.toml
```diff
@@ -21,7 +21,7 @@ num-traits = "0.2"
 num-derive = "0.3"
 libc = "0.2.108"
 openssl-sys = "0.9.75"
-openssl = "0.10.32"
+openssl = "0.10.70"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
 tracing = "0.1.37"
 futures = "0.3"
```
Lorak-mmk: "cargo: bump openssl to 0.10.70 "

I wonder why this change is necessary. It only narrows version set accepted by cpp-rust-driver, right?

Lorak-mmk (approved):

scylla-rust-wrapper/Cargo.toml
muzarski: True, this is unnecessary. I don't quite remember why I decided to bump this version. I think it was just a quick workaround without examining the conflict error. AFAIU, `cargo update` is enough - it will update the locked version of openssl dependency.

muzarski:
v1.3: Dropped the commit that bumps openssl version in Cargo.toml in favor of `cargo update` (which updates the lockfile).

PR #225, opened 2025-03-11
treewide: change *const to *mut based on cassandra.h

muzarski:
There are some types that are shared (`Arc`), but are mutable from cpp-driver perspective.

For some reason, there were some discrepancies between C API (defined in cassandra.h) and rust binding functions. On C side, the pointer would be `T*` (non-const), while on Rust side it would be `*const T`. This PR fixes those discrepancies, so we are consistent with cassandra.h

~While going over these types, I found out that for example `CassResult` unnecessarily implements `ArcFFI`. I'll address it in a separate PR (because there are other things to adjust as well).~ Nvm, I just didn't notice that `CassResult` is cloned in `cass_future_get_result` - this is because we use `.clone()` here and not `Arc::clone()`. I'll change it then, so it's easier to spot in the future.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -446,12 +446,8 @@ pub fn get_column_type(column_type: &ColumnType) -> CassDataType {
     CassDataType::new(inner)
 }
 
-// Changed return type to const ptr - ArcFFI::into_ptr is const.
-// It's probably not a good idea - but cppdriver doesn't guarantee
-// thread safety apart from CassSession and CassFuture.
-// This comment also applies to other functions that create CassDataType.
 #[no_mangle]
-pub unsafe extern "C" fn cass_data_type_new(value_type: CassValueType) -> *const CassDataType {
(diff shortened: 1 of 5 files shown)


muzarski:
I see that there are other types with this issue. I'll address them as well.

Lorak-mmk (approved):
I see no harm in this change.

I don't know in general what pointer types should we used. 
There are 2 competing semantics, Rust one (mut = exclusive, const = shared) and C++ one (mut = logically mutable, const = logically immutable). Afaik `*const` and `*mut` are not strictly bound to any of them.
IIUC, this PR aligns them with C++ semantics, right?

muzarski:
> IIUC, this PR aligns them with C++ semantics, right?

Correct. Take for example some type T, that is shared but can be mutated. In rust, we would represent it as e.g. `Arc<Mutex<T>>` or `Arc<UnsafeCell<T>>`. From C perspective (and cpp-driver API), this is just a `*` (not `const *`), because we can mutate it.

wprzytula (approved):

PR #227, opened 2025-03-11
cargo: add "lib" to crate-type list

muzarski:
It's required to enable the doctests for the crate. They will come in handy during further safety refactor.

I also marked the code snippets from documentation in binding.rs with `rust,ignore`. They should not be run as doc tests.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -371,13 +371,13 @@ macro_rules! invoke_binder_maker_macro_with_type {
 ///         There are also 2 helper variants, to accomodate scenarios often encountered in cppdriver (sets of 3 functions,
 ///         binding the same type by index, name and name_n):
 ///  * `make_binders!(type, fn_idx, fn_name, fn_name_n)` - is equivalent to:
-///     ```
+///     ```rust,ignore
 ///     make_binders!(@index type, fn_idx);
 ///     make_binders!(@name type, fn_name);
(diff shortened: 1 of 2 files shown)


scylla-rust-wrapper/src/binding.rs
```diff
@@ -371,13 +371,13 @@ macro_rules! invoke_binder_maker_macro_with_type {
...
+///     ```text
 ///     make_binders!(@index type, fn_idx);
 ///     make_binders!(@name type, fn_name);
 ///     make_binders!(@name_n type, fn_name_n);
 ///     ```
 ///  * `make_binders!(t1, fn_idx, t2, fn_name, t3, fn_name_n)` - is equivalent to:
-///     ```
+///     ```text
 ///     make_binders!(@index t1, fn_idx);
 ///     make_binders!(@name t2, fn_name);
```
Lorak-mmk: `rust,ignore` should work here, instead of `text`, right? It would preserve syntax highlighting.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/binding.rs
muzarski: True, it works. Thanks!

muzarski:
v1.1: changed `text` to `rust,ignore`, to preserve syntax highlighting

Lorak-mmk (approved):

PR #228, opened 2025-03-11
implement `cass_table_meta_column`

muzarski:
The order of table columns in cpp-driver (and in DESCRIBE KEYSPACE from cqlsh):
1. pk columns sorted by `position`
2. ck columns sorted by `position`
3. remaining columns sorted alphabetically

This is exactly how I implemented `cass_table_meta_column`.

My only concern is that every time the column is not a part of neither pk nor ck, we need to allocate a vector representing the remaining columns, and sort it. I wonder if it's worth adding such vector to `CassTableMeta`, so we do that only one time.

Advantage of other solution is obviously that we increase the performance of `cass_table_meta_column` - I'm not sure how often this method is called by the users.
Disadvantage is that we need to store another vector for each table metadata. It means more memory consumption by metadata structs, which is already big enough.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/metadata.rs
+++ b/scylla-rust-wrapper/src/metadata.rs
@@ -32,6 +32,8 @@ pub struct CassTableMeta {
     pub columns_metadata: HashMap<String, CassColumnMeta>,
     pub partition_keys: Vec<String>,
     pub clustering_keys: Vec<String>,
+    /// Non-key columns sorted alphabetically by name.
+    pub non_key_sorted_columns: Vec<String>,
     pub views: HashMap<String, Arc<CassMaterializedViewMeta>>,
 }
 
@@ -81,11 +83,25 @@ pub fn create_table_metadata(table_name: &str, table_metadata: &Table) -> CassTa
             columns_metadata.insert(column_name.clone(), cass_column_meta);
         });
 
+    let mut non_key_sorted_columns = columns_metadata
+        .iter()
+        .filter(|(_, column)| {
+            !matches!(
+                column.column_kind,
+                CassColumnType::CASS_COLUMN_TYPE_PARTITION_KEY
+                    | CassColumnType::CASS_COLUMN_TYPE_CLUSTERING_KEY,
+            )
+        })
+        .map(|(name, _column)| name.to_owned())
+        .collect::<Vec<_>>();
+    non_key_sorted_columns.sort_unstable();
+
     CassTableMeta {
         name: table_name.to_owned(),
         columns_metadata,
         partition_keys: table_metadata.partition_key.clone(),
         clustering_keys: table_metadata.clustering_key.clone(),
+        non_key_sorted_columns,
         views: HashMap::new(),
     }
 }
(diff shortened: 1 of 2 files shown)


muzarski:
v1.1: Changed `EXPECT_EQ` to `ASSERT_EQ` in test.

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -209,6 +209,58 @@ pub unsafe extern "C" fn cass_table_meta_column_count(table_meta: *const CassTab
...
+    // 3. remaining columns in alphabetical order <- this is something we need to guarantee.
+
+    let table_meta = RefFFI::as_ref(table_meta);
+    let index = index as usize;
+
+    // Check if the index lands in partition keys. If so, simply return the corresponding column.
+    if let Some(pk_name) = table_meta.partition_keys.get(index) {
+        // unwrap: partition key must exist in columns_metadata. This is ensured by rust-driver.
+        return RefFFI::as_ptr(table_meta.columns_metadata.get(pk_name).unwrap());
+    }
```
Lorak-mmk, line 258: 📝 It is ensured only since 1.0. But I see that PR bumping the driver version to 1.0 was already merged, so this is fine.

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -209,6 +209,58 @@ pub unsafe extern "C" fn cass_table_meta_column_count(table_meta: *const CassTab
...
+            )
+        })
+        .collect::<Vec<_>>();
+    // Sort remaining columns alphabetically.
+    non_pk_ck_columns.sort_unstable_by_key(|column| &column.name);
+
+    non_pk_ck_columns
+        .get(index)
+        .map_or(std::ptr::null(), |column| RefFFI::as_ptr(column))
+}
```
Lorak-mmk: I think it makes sense to precompute the vector of non-key sorted columns.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -209,6 +209,58 @@ pub unsafe extern "C" fn cass_table_meta_column_count(table_meta: *const CassTab
...
+#[no_mangle]
+pub unsafe extern "C" fn cass_table_meta_column(
+    table_meta: *const CassTableMeta,
+    index: size_t,
+) -> *const CassColumnMeta {
+    // The order of columns in cpp-driver (and in DESCRIBE TABLE in cqlsh):
+    // 1. partition keys sorted by position <- this is guaranteed by rust-driver.
+    //    Table::partition_keys is a Vector of pk names, sorted by position.
+    // 2. clustering keys sorted by position <- this is guaranteed by rust-driver (same reasoning as above).
+    // 3. remaining columns in alphabetical order <- this is something we need to guarantee.
```
wprzytula, line 237: :wrench: Could you provide examples here? Those from our PM on Slack seem to be a good fit.

muzarski: I added an example that we use in the integration test.

scylla-rust-wrapper/src/metadata.rs
muzarski: Done. Precomputed the `Vec<String>` that contains the names of non-key columns sorted by the column name.

muzarski:
v1.2: Addressed review comments

wprzytula (approved):

Lorak-mmk (approved):

PR #229, opened 2025-03-12
safety: pointer related bugfixes

muzarski:
This PR introduces multiple fixes around pointer manipulation in the driver. These bugs were found thanks to the `CassPtr` which will be introduced in a follow-up PR.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/metadata.rs
+++ b/scylla-rust-wrapper/src/metadata.rs
@@ -427,7 +427,14 @@ pub unsafe extern "C" fn cass_materialized_view_meta_base_table(
     view_meta: *const CassMaterializedViewMeta,
 ) -> *const CassTableMeta {
     let view_meta = RefFFI::as_ref(view_meta);
-    view_meta.base_table.as_ptr()
+
+    match view_meta.base_table.upgrade() {
+        Some(arc) => RefFFI::as_ptr(&arc),
+        None => {
+            tracing::error!("Failed to upgrade a weak reference to table metadata from materialized view metadata! This is a driver bug!");
+            std::ptr::null()
+        }
+    }
 }
 
 #[no_mangle]
--- a/scylla-rust-wrapper/src/metadata.rs
+++ b/scylla-rust-wrapper/src/metadata.rs
@@ -51,7 +51,7 @@ impl RefFFI for CassMaterializedViewMeta {}
 
 pub struct CassColumnMeta {
     pub name: String,
-    pub column_type: CassDataType,
+    pub column_type: Arc<CassDataType>,
     pub column_kind: CassColumnType,
 }
 
@@ -66,7 +66,7 @@ pub fn create_table_metadata(table_name: &str, table_metadata: &Table) -> CassTa
         .for_each(|(column_name, column_metadata)| {
             let cass_column_meta = CassColumnMeta {
                 name: column_name.clone(),
-                column_type: get_column_type(&column_metadata.typ),
+                column_type: Arc::new(get_column_type(&column_metadata.typ)),
                 column_kind: match column_metadata.kind {
                     ColumnKind::Regular => CassColumnType::CASS_COLUMN_TYPE_REGULAR,
                     ColumnKind::Static => CassColumnType::CASS_COLUMN_TYPE_STATIC,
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/future.rs
```diff
@@ -258,7 +260,12 @@ impl CassFuture {
         }
     }
 
-    pub fn set_callback(&self, cb: CassFutureCallback, data: *mut c_void) -> CassError {
+    pub unsafe fn set_callback(
+        &self,
+        self_ptr: *mut CassFuture,
+        cb: CassFutureCallback,
+        data: *mut c_void,
+    ) -> CassError {
```
wprzytula: :question: What do we achieve by passing `*mut Self` alongside `&Self`? Can't we just cast `&Self` to `*mut Self` in `set_callback`'s body?

muzarski: This is explained in commit message. It's preparation for `CassPtr` introduction. This `*mut CassFuture` will be replaced by `CassPtr`.

I don't want expose an API to convert a reference to `CassPtr` via `ArcFFI`. I want to be sure that `ArcFFI` creates `CassPtr` which are backed by `Arc` allocation. We don't have such guarantee if we allow to create a pointer from just ref (`&`).

Here is a snippet from the follow-up PR. These are the only methods on `ArcFFI` that convert rust reference primitives to pointers:
```rust
pub trait ArcFFI: Sized {
    /// Creates a pointer from a valid reference to Arc-allocated data.
    /// Holder of the pointer borrows the pointee.
    fn as_ptr<'a>(self: &'a Arc<Self>) -> CassPtr<'a, Self, (Const,)> {
        let ptr = Arc::as_ptr(self);

        // SAFETY:
        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation
        // 2. pointer's lifetime - pointer inherits the lifetime of provided Arc's borrow.
        //    What's important is that the returned pointer borrows the data, and is not the
        //    shared owner. Thus, user cannot call ArcFFI::free() on such pointer.
        // 3. mutability - we always create a `Const` pointer.
        unsafe { CassPtr::from_raw(ptr) }
    }

    /// Creates a pointer from a valid Arc allocation.
    fn into_ptr(self: Arc<Self>) -> CassPtr<'static, Self, (Const,)> {
        let ptr = Arc::into_raw(self);

        // SAFETY:
        // 1. validity guarantee - pointer is valid, since it's obtained from Arc allocation
        // 2. pointer's lifetime - returned pointer has a 'static lifetime. It is a shared
        //    owner of the pointee. User has to decrease the RC of the pointer (and potentially free the memory)
        //    via ArcFFI::free().
        // 3. mutability - we always create a `Const` pointer.
        unsafe { CassPtr::from_raw(ptr) }
    }
}
```

Notice that if we expose a method like `fn as_ptr_from_ref(&self) -> CassPtr<'_, Self, (Const,)>`, we lose the guarantee that this reference originates from some `Arc` allocation.

Why do we need such guarantee? An example:
```rust
struct Foo;
impl ArcFFI for Foo {}

let foo = Foo;
let ptr: CassPtr = ArcFFI::as_ptr_from_ref(&foo);

// We try to increase RC of a pointer that is not Arc-allocated.
let arced: Arc<Foo> = ArcFFI::cloned_from_ptr(ptr);
```

muzarski: I can make this commit a part of follow-up PR, so there is more context why this change makes sense.

muzarski:
v1.1: Dropped `future: pass future pointer to callback directly` commit. It will be part of follow-up PR.

wprzytula (approved):

scylla-rust-wrapper/src/metadata.rs
```diff
@@ -427,7 +427,11 @@ pub unsafe extern "C" fn cass_materialized_view_meta_base_table(
     view_meta: *const CassMaterializedViewMeta,
 ) -> *const CassTableMeta {
     let view_meta = RefFFI::as_ref(view_meta);
-    view_meta.base_table.as_ptr()
+
+    match view_meta.base_table.upgrade() {
+        Some(arc) => RefFFI::as_ptr(&arc),
+        None => std::ptr::null(),
+    }
```
Lorak-mmk: Commit: "metadata: try to upgrade Weak before returning a pointer "

Commit message says that if the upgrade fails then there is a serious bug in the driver.
If that is the case, maybe we should panic when this happens, or at least print some warning message?

Lorak-mmk (approved):

scylla-rust-wrapper/src/metadata.rs
muzarski: Added `tracing::error!` log

Lorak-mmk (approved):

PR #230, opened 2025-03-12
iterator: cleanup iterator module and implement cass_iterator_type

### LINKED ISSUE #47: CassIterator missing functions
In cppdriver CassIterator let's you also iterate over maps/udts/tuples/collections, not only rows in dataset.

Missing functions:

- [ ] `cass_iterator_type`
- [x] `cass_iterator_from_row`
- [x] `cass_iterator_from_collection`
- [x] `cass_iterator_from_map`
- [x] `cass_iterator_from_tuple`
- [x] `cass_iterator_fields_from_user_type`
- [x] `cass_iterator_get_column`
- [x] `cass_iterator_get_value`
- [x] `cass_iterator_get_map_key`
- [x] `cass_iterator_get_map_value`
- [x] `cass_iterator_get_user_type_field_name`
- [x] `cass_iterator_get_user_type_field_value`
- [x] `cass_iterator_get_user_type`

muzarski:
This PR mostly does some cleanups/small refactors. In addition, it implements `cass_iterator_type` method.

Fixes: https://github.com/scylladb/cpp-rust-driver/issues/47

## Pre-review checklist


scylla-rust-wrapper/src/iterator.rs
```diff
@@ -83,17 +83,33 @@ pub enum CassMaterializedViewsMetaIterator<'schema> {
...
-    CassMapIterator(CassMapIterator<'result_or_schema>),
-    CassTupleIterator(CassCollectionIterator<'result_or_schema>),
-    CassUdtIterator(CassUdtIterator<'result_or_schema>),
-    CassSchemaMetaIterator(CassSchemaMetaIterator<'result_or_schema>),
-    CassKeyspaceMetaTableIterator(CassKeyspaceMetaIterator<'result_or_schema>),
-    CassKeyspaceMetaUserTypeIterator(CassKeyspaceMetaIterator<'result_or_schema>),
-    CassMaterializedViewsMetaIterator(CassMaterializedViewsMetaIterator<'result_or_schema>),
-    CassColumnsMetaIterator(CassColumnsMetaIterator<'result_or_schema>),
+    // Iterators derived from CassResult.
+    // Naming convetion of the variants: the name of the collection.
```
wprzytula: :pick: typo: `convetion` -> `convention`

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -83,17 +83,33 @@ pub enum CassMaterializedViewsMetaIterator<'schema> {
...
+    /// Iterator over rows in a result.
+    Result(CassResultIterator<'result_or_schema>),
+    /// Iterator over columns (values) in a row.
+    Row(CassRowIterator<'result_or_schema>),
+    /// Iterator over values in a collection.
+    Collection(CassCollectionIterator<'result_or_schema>),
+    /// Iterator over key-value pairs in a map.
+    Map(CassMapIterator<'result_or_schema>),
```
wprzytula, line 97: :question: :thinking: Is the iterator really over key-value pairs? I vaguely remember it iterates over keys and values interleaved.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -83,17 +83,33 @@ pub enum CassMaterializedViewsMetaIterator<'schema> {
...
+    /// Iterator over keyspaces in schema metadata.
+    KeyspacesMeta(CassSchemaMetaIterator<'result_or_schema>),
+    /// Iterator over tables in keyspace metadata.
+    TablesMeta(CassKeyspaceMetaIterator<'result_or_schema>),
+    /// Iterator over UDTs in keyspace metadata.
+    UserTypes(CassKeyspaceMetaIterator<'result_or_schema>),
+    /// Iterator over materialized views in either keyspace or table metadata.
+    MaterializedViewsMeta(CassMaterializedViewsMetaIterator<'result_or_schema>),
+    /// Iterator over columns metadata in either table or view metadata.
+    ColumnsMeta(CassColumnsMetaIterator<'result_or_schema>),
```
wprzytula: :tada: I love this commit. Everything suddenly so neat, readable and well-documented! Congrats!

wprzytula (changes requested):

scylla-rust-wrapper/src/iterator.rs
muzarski: There are two kinds of iterators that can be created from map:
- `cass_iterator_from_map` -> this creates an iterator over key-value pairs. There are methods to obtain the [key](https://github.com/scylladb/cpp-rust-driver/blob/master/scylla-rust-wrapper/src/query_result.rs#L542) and [value](https://github.com/scylladb/cpp-rust-driver/blob/master/scylla-rust-wrapper/src/query_result.rs#L569) at given position.
- `cass_iterator_from_collection` -> this iterates over keys and values interleaved, as you mentioned. See [cass_iterator_get_value](https://github.com/scylladb/cpp-rust-driver/blob/master/scylla-rust-wrapper/src/query_result.rs#L505)

muzarski:
v1.1: Fixed the typo in the comment

scylla-rust-wrapper/src/iterator.rs
wprzytula: Makes sense, thanks!

wprzytula (approved):

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -85,6 +87,34 @@ pub unsafe extern "C" fn cass_iterator_free(iterator: *mut CassIterator) {
...
+            CassIteratorType::CASS_ITERATOR_TYPE_TYPE_META
+        }
+        CassIterator::CassKeyspaceMetaViewIterator(_) => {
+            CassIteratorType::CASS_ITERATOR_TYPE_MATERIALIZED_VIEW_META
+        }
+        CassIterator::CassTableMetaIterator(_) => CassIteratorType::CASS_ITERATOR_TYPE_COLUMN_META,
+        CassIterator::CassViewMetaIterator(_) => CassIteratorType::CASS_ITERATOR_TYPE_COLUMN_META,
+    }
+}
+
```
Lorak-mmk: Commit: "iterator: implement cass_iterator_type"

What does it mean that it needs to be adjusted? I've read the commit message, I've read the code, and I don't know what it means.

scylla-rust-wrapper/src/iterator.rs
Lorak-mmk, line 1: "iterator: add CassTupleIterator variant "

Commit message says that we need a way to distinguish iterator over tuple from iterator over other collection, but does not explain why.  Please explain - it is not obvious to me at all.

Lorak-mmk, line 1: Commit: "iterator: introduce CassColumnMeta iterator "

Again, why should we do this? Is there some difference in behavior between those iterators? Does it allow some nice code cleanups?

Lorak-mmk: The name of this new variant seems inconsistent.
We have CassKeyspaceMetaTableIterator, which iterator over Tables in a Keyspace metadata.
There are 2 other similar variants (for udts and views).

We have CassTableMetaIterator, which iterates over stuff (columns) in a Table
We have CassViewMetaIterator, which iterates over stuff (columns) in a View

Now you introduce CassColumnsMetaIterator - interpreting the name the same as others, I would assume it iterates over some stuff in Column, but it iterates over columns.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -85,6 +93,7 @@ pub enum CassIterator<'result_or_schema> {
     CassKeyspaceMetaTableIterator(CassKeyspaceMetaIterator<'result_or_schema>),
     CassKeyspaceMetaUserTypeIterator(CassKeyspaceMetaIterator<'result_or_schema>),
     CassKeyspaceMetaViewIterator(CassKeyspaceMetaIterator<'result_or_schema>),
+    CassMaterializedViewsMetaIterator(CassMaterializedViewsMetaIterator<'result_or_schema>),
     CassColumnsMetaIterator(CassColumnsMetaIterator<'result_or_schema>),
     CassTableMetaIterator(CassTableMetaIterator<'result_or_schema>),
```
Lorak-mmk: Ditto about inconsistent naming.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -67,33 +67,51 @@ pub struct CassViewMetaIterator<'schema> {
...
+    UdtFields(CassUdtIterator<'result_or_schema>),
+    /// Iterator over keyspaces in schema metadata.
+    KeyspacesMeta(CassSchemaMetaIterator<'result_or_schema>),
+    /// Iterator over tables in keyspace metadata.
+    TablesMeta(CassKeyspaceMetaIterator<'result_or_schema>),
+    /// Iterator over UDTs in keyspace metadata.
+    UserTypes(CassKeyspaceMetaIterator<'result_or_schema>),
+    /// Iterator over materialized views in either keyspace or table metadata.
+    MaterializedViewsMeta(CassMaterializedViewsMetaIterator<'result_or_schema>),
+    /// Iterator over columns metadata in either table or view metadata.
```
Lorak-mmk, line 113: Hmm... ok so you try fix the consistency here, after introducing inconsistency before.

Previously we had one naming convention, now we have too. Why can't we have one?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/iterator.rs
muzarski: Because from user's perspective these are two different kinds of iterators. One created from `cass_iterator_from_collection`, and the other created from `cass_iterator_from_tuple`. In our case, the underlying structure is the same, but for the user these are not the same. Why? Because there is `CASS_ITERATOR_TYPE_COLLETION` and `CASS_ITERATOR_TYPE_TUPLE`. This is the exact reason why we need to distinguish them - to return corresponding `CassIteratorType` in `cass_iterator_type` method.

I'll put more context to the commit message.

scylla-rust-wrapper/src/iterator.rs
muzarski: The naming conventions are derived from `CassColumnType` enum variants. I don't know why, but this is the convention chosen by cpp-driver. For result-derived iterators, it's the name of the collection we iterate over. For schema-based iterators, it's the name of the items in the collection we iterate over.

As a side note, it took me a while to notice the difference. When I first saw that iterator returned from `cass_iterator_from_result` is `CASS_ITERATOR_TYPE_RESULT`, I assumed that iterator returned from `cass_iterator_keyspaces_from_schema_meta`, will be `CASS_ITERATOR_TYPE_SCHEMA_META` or sth like that. But no, its type is `CASS_ITERATOR_TYPE_KEYSPACE_META` (because we iterate over keyspaces metadata from schema metadata).

scylla-rust-wrapper/src/iterator.rs
muzarski: > Commit: "iterator: introduce CassColumnMeta iterator "
> 
> Again, why should we do this? Is there some difference in behavior between those iterators? Does it allow some nice code cleanups?

Notice that before this PR we used `CassIterator::CassTableMetaIterator` for two things:
1. as you noticed, to iterate over columns metadata (`cass_iterator_columns_from_table_meta`)
2. to iterate over materialized views meta (`cass_iterator_materialized_views_from_table_meta`)

But there is other way to create an iterator over columns. Namely `cass_iterator_columns_from_materialized_view_meta`.

And now, both iterators `cass_iterator_columns_from_table_meta` and `cass_iterator_columns_from_materialized_view_meta` should return `CASS_ITERATOR_TYPE_COLUMN_META` when `cass_iterator_type` is called on them.

This is why I introduce the additional layer (`CassIterator::CassColumnsMetaIterator`). I want `CassIterator` top-level hierarchy directly correspond to `CassIteratorType` enum variants.

scylla-rust-wrapper/src/iterator.rs
muzarski: As I mentioned in my other comment, I want to adjust `CassIterator` enum hierarchy so each variant corresponds to the variant in `CassIteratorType` enum. I'll put this information in the code comment and commit message.

Lorak-mmk: Later commits explained that it is about naming of the variants.
I don't think this commit needs to mention that. Naming change is unrelated.

muzarski: > Later commits explained that it is about naming of the variants. I don't think this commit needs to mention that. Naming change is unrelated.

Not only the naming, but structure as well.

Lorak-mmk: ok

muzarski:
v1.2: Adjusted commit messages so they better explain why `CassIterator` hierarchy is changed in this PR.

Lorak-mmk (approved):

PR #236, opened 2025-04-04
Lazy deser

### LINKED ISSUE #50: CassValue missing functions
Function list:

- [x] `cass_value_data_type`
- [x] `cass_value_get_bytes`
- [x] `cass_value_get_decimal`
- [x] `cass_value_get_duration`
- [x] `cass_value_type`
- [x] `cass_value_is_collection`
- [x] `cass_value_is_duration`
- [x] `cass_value_item_count`
- [x] `cass_value_primary_sub_type`
- [x] `cass_value_secondary_sub_type`

### LINKED ISSUE #94: Add support for `cass_value_get_decimal` and `cass_value_get_bytes`

### LINKED ISSUE #216: Lazy deserialization
Currently, rows result are eagerly deserialized to `Vec<Row>`. We can definitely improve that.

1. We do not need to store `Vec<CassRow>` in `CassRowsResult`. This because there is no API function that returns a row at given index. We only need to deserialize the first row (`cass_result_first_row`) and store it in `CassRowsResult`.
2. We do not need to (and should not) deserialize values to `CqlValue`. Because of this, we are not able to implement `cass_value_get_bytes`, which should return serialized value. We should rather deserialize to some partially-serialized type, which contains the bytes and information about the CQL column type.
```[tasklist]
### Tasks
- [ ] #50 
- [ ] https://github.com/scylladb/cpp-rust-driver/pull/213
```

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/216, Fixes: https://github.com/scylladb/cpp-rust-driver/issues/50, Fixes: https://github.com/scylladb/cpp-rust-driver/issues/94

This PR introduces lazy deserialization to the driver!

## PR structure
1. ~Commits 1 and 2 - there are small fixes not really related to deserialization~
2. Commit 1 - introduction of main heroes, namely `CassRawRow` and `CassRawValue` 
3. Then there are commits which introduce changes that still compile, but allow to reduce the noise during review of `query_result: lazy deserialization` commit. These commits mostly introduce new types (e.g. iterators)
4. Commit `query_result: lazy deserialization ` - Replaces the old deserialization API with a new one. It adjust the implementation of C API. In particular, the `cass_value_get_bytes` is now properly implemented.
5. At the end there are commits introducing deserialization unit tests.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/iterator.rs
+++ b/scylla-rust-wrapper/src/iterator.rs
@@ -1,5 +1,5 @@
 use scylla::deserialize::result::TypedRowIterator;
-use scylla::deserialize::value::{ListlikeIterator, MapIterator};
+use scylla::deserialize::value::{DeserializeValue, ListlikeIterator, MapIterator};
 use scylla::value::Row;
 
 use crate::argconv::{
@@ -189,6 +189,213 @@ impl CassCollectionIterator<'_> {
             CassCollectionIterator::Listlike(listlike_iterator) => listlike_iterator.next(),
             CassCollectionIterator::Map(map_collection_iterator) => map_collection_iterator.next(),
         }
+    }
+}
+
+// TODO: consider introducing this to Rust driver.
+mod tuple_iterator {
+    use scylla::cluster::metadata::ColumnType;
+    use scylla::deserialize::value::{self, DeserializeValue};
+    use scylla::deserialize::FrameSlice;
+    use scylla::errors::{DeserializationError, TypeCheckError};
+    use scylla::frame::frame_errors::LowLevelDeserializationError;
+    use thiserror::Error;
+
+    pub(super) struct TupleIterator<'frame, 'metadata> {
+        all_metadata: &'metadata [ColumnType<'metadata>],
+        remaining_metadata: &'metadata [ColumnType<'metadata>],
+        iterator: BytesSequenceIterator<'frame>,
+    }
+
+    impl<'frame, 'metadata> TupleIterator<'frame, 'metadata> {
+        fn new(metadata: &'metadata [ColumnType<'metadata>], slice: FrameSlice<'frame>) -> Self {
+            Self {
+                all_metadata: metadata,
+                remaining_metadata: metadata,
+                iterator: BytesSequenceIterator::new(slice),
+            }
+        }
+    }
+
+    impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for TupleIterator<'frame, 'metadata> {
+        fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+            match typ {
+                ColumnType::Tuple { .. } => Ok(()),
+                _ => Err(TypeCheckError::new(value::BuiltinTypeCheckError {
+                    rust_name: std::any::type_name::<Self>(),
+                    cql_type: typ.clone().into_owned(),
+                    kind: value::BuiltinTypeCheckErrorKind::TupleError(
+                        value::TupleTypeCheckErrorKind::NotTuple,
+                    ),
+                })),
+            }
+        }
+
+        fn deserialize(
+            typ: &'metadata ColumnType<'metadata>,
+            v: Option<FrameSlice<'frame>>,
+        ) -> Result<Self, DeserializationError> {
+            let slice: FrameSlice<'frame> = v.ok_or_else(|| {
+                DeserializationError::new(value::BuiltinDeserializationError {
+                    rust_name: std::any::type_name::<Self>(),
+                    cql_type: typ.clone().into_owned(),
+                    kind: value::BuiltinDeserializationErrorKind::ExpectedNonNull,
+                })
+            })?;
+
+            let metadata = match typ {
+                ColumnType::Tuple(types) => types.as_slice(),
+                _ => {
+                    unreachable!("Typecheck should have prevented this scenario!")
+                }
+            };
+
+            Ok(Self::new(metadata, slice))
+        }
+    }
+
+    impl<'frame, 'metadata> Iterator for TupleIterator<'frame, 'metadata> {
+        type Item = Result<
+            (&'metadata ColumnType<'metadata>, Option<FrameSlice<'frame>>),
+            DeserializationError,
+        >;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            let pos = self.all_metadata.len() - self.remaining_metadata.len();
+
+            // We don't fail if there are more serialized fields than declared in the metadata.
+            // This is what we do for static-size tuple deserialization.
+            // For example, (a, b, c) does not fail if slice contains 4 serialized values (in Rust-driver).
+            let (head, metadata) = self.remaining_metadata.split_first()?;
+            self.remaining_metadata = metadata;
+
+            let raw_res = match self.iterator.next() {
+                Some(Ok(raw)) => Ok((head, raw)),
+                Some(Err(e)) => Err(DeserializationError::new(
+                    value::BuiltinDeserializationError {
+                        rust_name: std::any::type_name::<Self>(),
+                        cql_type: ColumnType::Tuple(self.all_metadata.to_owned()).into_owned(),
+                        kind: value::BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
+                    },
+                )),
+                // Value is missing.
+                None => Err(DeserializationError::new(TupleMissingValue(pos))),
+            };
+
+            Some(raw_res)
+        }
+    }
+
+    // This would be a variant of TupleDeserializationErrorKind in rust-driver.
+    #[derive(Error, Debug)]
+    #[error("Failed to deserialize tuple: serialized value is missing as position {0}")]
+    struct TupleMissingValue(usize);
+
+    // --------------------------
+    // COPIED FROM RUST-DRIVER!!!
+    // --------------------------
+    #[derive(Clone, Copy, Debug)]
+    struct BytesSequenceIterator<'frame> {
+        slice: FrameSlice<'frame>,
+    }
+
+    impl<'frame> BytesSequenceIterator<'frame> {
+        fn new(slice: FrameSlice<'frame>) -> Self {
+            Self { slice }
+        }
+    }
+
+    impl<'frame> From<FrameSlice<'frame>> for BytesSequenceIterator<'frame> {
+        #[inline]
+        fn from(slice: FrameSlice<'frame>) -> Self {
+            Self::new(slice)
+        }
+    }
+
+    impl<'frame> Iterator for BytesSequenceIterator<'frame> {
+        type Item = Result<Option<FrameSlice<'frame>>, LowLevelDeserializationError>;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            if self.slice.as_slice().is_empty() {
+                None
+            } else {
+                Some(self.slice.read_cql_bytes())
+            }
+        }
+    }
+}
+
+/// Iterator created from [`cass_iterator_from_tuple()`].
+pub struct CassTupleIterator<'result> {
+    iterator: tuple_iterator::TupleIterator<'result, 'result>,
+    metadata: &'result [Arc<CassDataType>],
+    current_entry: Option<CassTupleIteratorEntry<'result>>,
+}
+
+pub struct CassTupleIteratorEntry<'result> {
+    field_value: CassValue<'result>,
+    metadata_types_index: usize,
+}
+
+impl<'result> CassTupleIterator<'result> {
+    fn new_from_value(
+        value: &'result CassValue<'result>,
+    ) -> Result<Self, NonNullDeserializationError> {
+        let tuple_iterator = value.get_non_null::<tuple_iterator::TupleIterator>()?;
+
+        // SAFETY: `CassDataType` is obtained from `CassResultMetadata`, which is immutable.
+        let metadata = match unsafe { value.value_type.get_unchecked() } {
+            CassDataTypeInner::Tuple(typ) => typ.as_slice(),
+            _ => panic!("Expected tuple type. Typecheck should have prevented such scenario!"),
+        };
+
+        Ok(Self {
+            iterator: tuple_iterator,
+            metadata,
+            current_entry: None,
+        })
+    }
+
+    fn next(&mut self) -> bool {
+        // Handle scenario where underlying iterator is exhausted.
+        let Some(deser_result) = self.iterator.next() else {
+            return false;
+        };
+
+        // Handle the deserialization error from underlying iterator.
+        let Ok((column_type, slice)) = deser_result else {
+            tracing::error!("Failed to deserialize next tuple value: {deser_result:?}");
+            return false;
+        };
+
+        // Deserialize to CassRawValue.
+        let raw_value: CassRawValue = match <_ as DeserializeValue>::deserialize(column_type, slice)
+        {
+            Ok(v) => v,
+            Err(e) => {
+                tracing::error!("Failed to deserialize next tuple value: {e}");
+                return false;
+            }
+        };
+
+        // Update the current entry.
+        let new_metadata_types_index = self
+            .current_entry
+            .as_ref()
+            .map(|entry| entry.metadata_types_index + 1)
+            .unwrap_or(0);
+
+        let new_value = CassValue {
+            value: raw_value,
+            value_type: &self.metadata[new_metadata_types_index],
+        };
+
+        self.current_entry = Some(CassTupleIteratorEntry {
+            field_value: new_value,
+            metadata_types_index: new_metadata_types_index,
+        });
+
+        true
     }
 }
 
--- a/scylla-rust-wrapper/src/iterator.rs
+++ b/scylla-rust-wrapper/src/iterator.rs
@@ -72,13 +72,13 @@ impl CassRowIterator<'_> {
     }
 }
 
-pub struct CassCollectionIterator<'result> {
+pub struct LegacyCassCollectionIterator<'result> {
     value: &'result CassValue,
     count: u64,
     position: Option<usize>,
 }
 
-impl CassCollectionIterator<'_> {
+impl LegacyCassCollectionIterator<'_> {
     fn next(&mut self) -> bool {
         let new_pos: usize = self.position.map_or(0, |prev_pos| prev_pos + 1);
 
@@ -88,13 +88,13 @@ impl CassCollectionIterator<'_> {
     }
 }
 
-pub struct CassMapIterator<'result> {
+pub struct LegacyCassMapIterator<'result> {
     value: &'result CassValue,
     count: u64,
     position: Option<usize>,
 }
 
-impl CassMapIterator<'_> {
+impl LegacyCassMapIterator<'_> {
     fn next(&mut self) -> bool {
         let new_pos: usize = self.position.map_or(0, |prev_pos| prev_pos + 1);
 
@@ -104,13 +104,13 @@ impl CassMapIterator<'_> {
     }
 }
 
-pub struct CassUdtIterator<'result> {
+pub struct LegacyCassUdtIterator<'result> {
     value: &'result CassValue,
     count: u64,
     position: Option<usize>,
 }
 
-impl CassUdtIterator<'_> {
+impl LegacyCassUdtIterator<'_> {
     fn next(&mut self) -> bool {
         let new_pos: usize = self.position.map_or(0, |prev_pos| prev_pos + 1);
 
@@ -210,13 +210,13 @@ pub enum CassIterator<'result_or_schema> {
     /// Iterator over columns (values) in a row.
     Row(CassRowIterator<'result_or_schema>),
     /// Iterator over values in a collection.
-    Collection(CassCollectionIterator<'result_or_schema>),
+    Collection(LegacyCassCollectionIterator<'result_or_schema>),
     /// Iterator over key-value pairs in a map.
-    Map(CassMapIterator<'result_or_schema>),
+    Map(LegacyCassMapIterator<'result_or_schema>),
     /// Iterator over values in a tuple.
-    Tuple(CassCollectionIterator<'result_or_schema>),
+    Tuple(LegacyCassCollectionIterator<'result_or_schema>),
     /// Iterator over fields (values) in UDT.
-    Udt(CassUdtIterator<'result_or_schema>),
+    Udt(LegacyCassUdtIterator<'result_or_schema>),
 
     // Iterators derived from CassSchemaMeta.
     // Naming convention of the variants: name of item in the collection (plural).
@@ -726,7 +726,7 @@ pub unsafe extern "C" fn cass_iterator_from_collection<'result>(
     };
     let val = RefFFI::as_ref(value).unwrap();
 
-    let iterator = CassCollectionIterator {
+    let iterator = LegacyCassCollectionIterator {
         value: val,
         count: item_count,
         position: None,
@@ -744,7 +744,7 @@ pub unsafe extern "C" fn cass_iterator_from_tuple<'result>(
 
     if let Some(Value::CollectionValue(Collection::Tuple(val))) = &tuple.value {
         let item_count = val.len();
-        let iterator = CassCollectionIterator {
+        let iterator = LegacyCassCollectionIterator {
             value: tuple,
             count: item_count as u64,
             position: None,
@@ -765,7 +765,7 @@ pub unsafe extern "C" fn cass_iterator_from_map<'result>(
 
     if let Some(Value::CollectionValue(Collection::Map(val))) = &map.value {
         let item_count = val.len();
-        let iterator = CassMapIterator {
+        let iterator = LegacyCassMapIterator {
             value: map,
             count: item_count as u64,
             position: None,
@@ -786,7 +786,7 @@ pub unsafe extern "C" fn cass_iterator_fields_from_user_type<'result>(
 
     if let Some(Value::CollectionValue(Collection::UserDefinedType { fields, .. })) = &udt.value {
         let item_count = fields.len();
-        let iterator = CassUdtIterator {
+        let iterator = LegacyCassUdtIterator {
             value: udt,
             count: item_count as u64,
             position: None,
--- a/scylla-rust-wrapper/src/iterator.rs
+++ b/scylla-rust-wrapper/src/iterator.rs
@@ -1,5 +1,5 @@
 use scylla::deserialize::result::TypedRowIterator;
-use scylla::deserialize::value::{DeserializeValue, ListlikeIterator, MapIterator};
+use scylla::deserialize::value::{DeserializeValue, ListlikeIterator, MapIterator, UdtIterator};
 use scylla::value::Row;
 
 use crate::argconv::{
@@ -463,6 +463,85 @@ impl<'result> CassMapIterator<'result> {
         self.current_entry = new_entry;
 
         self.current_entry.is_some()
+    }
+}
+
+pub struct CassUdtIterator<'result> {
+    iterator: UdtIterator<'result, 'result>,
+    metadata: &'result [(String, Arc<CassDataType>)],
+    current_entry: Option<CassUdtIteratorEntry<'result>>,
+}
+
+struct CassUdtIteratorEntry<'result> {
+    field_value: CassValue<'result>,
+    metadata_types_index: usize,
+}
+
+impl<'result> CassUdtIterator<'result> {
+    fn new_from_value(
+        value: &'result CassValue<'result>,
+    ) -> Result<Self, NonNullDeserializationError> {
+        let udt_iterator = value.get_non_null::<UdtIterator>()?;
+
+        // SAFETY: `CassDataType` is obtained from `CassResultMetadata`, which is immutable.
+        let metadata = match unsafe { value.value_type.get_unchecked() } {
+            CassDataTypeInner::UDT(udt) => udt.field_types.as_slice(),
+            _ => panic!("Expected UDT type. Typecheck should have prevented such scenario!"),
+        };
+
+        Ok(Self {
+            iterator: udt_iterator,
+            metadata,
+            current_entry: None,
+        })
+    }
+
+    fn next(&mut self) -> bool {
+        // Handle scenario where underlying iterator is exhausted.
+        let Some(((_field_name, field_type), deser_result)) = self.iterator.next() else {
+            return false;
+        };
+
+        // Handle the deserialization error from underlying iterator.
+        let Ok(field_value) = deser_result else {
+            tracing::error!("Failed to deserialize next UDT field: {deser_result:?}");
+            return false;
+        };
+        // Flatten: Treat missing UDT field as null.
+        // This is a bit different from what cpp-driver does.
+        // cpp-driver fails in case when field metadata is provided, but the serialized value is missing.
+        // After the discussion, we concluded that this is not the correct behaviour. CQL protocol
+        // clearly states that this is a valid scenario: https://github.com/apache/cassandra/blob/4a80daf32eb4226d9870b914779a1fc007479da6/doc/native_protocol_v4.spec#L1003.
+        let field_slice = field_value.flatten();
+
+        // Deserialize to CassRawValue.
+        let raw_value: CassRawValue =
+            match <_ as DeserializeValue>::deserialize(field_type, field_slice) {
+                Ok(v) => v,
+                Err(e) => {
+                    tracing::error!("Failed to deserialize UDT field value: {e}");
+                    return false;
+                }
+            };
+
+        // Update the current entry.
+        let new_metadata_types_index = self
+            .current_entry
+            .as_ref()
+            .map(|entry| entry.metadata_types_index + 1)
+            .unwrap_or(0);
+
+        let new_value = CassValue {
+            value: raw_value,
+            value_type: &self.metadata[new_metadata_types_index].1,
+        };
+
+        self.current_entry = Some(CassUdtIteratorEntry {
+            field_value: new_value,
+            metadata_types_index: new_metadata_types_index,
+        });
+
+        true
     }
 }
 
--- a/scylla-rust-wrapper/src/iterator.rs
+++ b/scylla-rust-wrapper/src/iterator.rs
@@ -1,6 +1,5 @@
 use scylla::deserialize::result::TypedRowIterator;
 use scylla::deserialize::value::{DeserializeValue, ListlikeIterator, MapIterator, UdtIterator};
-use scylla::value::Row;
 
 use crate::argconv::{
     write_str_to_c, ArcFFI, BoxFFI, CConst, CMut, CassBorrowedExclusivePtr, CassBorrowedSharedPtr,
@@ -13,9 +12,8 @@ use crate::metadata::{
 };
 use crate::query_result::cass_raw_value::CassRawValue;
 use crate::query_result::{
-    cass_value_is_collection, cass_value_item_count, cass_value_type, CassResult, CassResultKind,
-    CassResultMetadata, CassRow, CassValue, Collection, LegacyCassValue,
-    NonNullDeserializationError, Value,
+    cass_value_type, CassRawRow, CassResult, CassResultKind, CassResultMetadata, CassRow,
+    CassValue, LegacyCassValue, NonNullDeserializationError,
 };
 use crate::types::{cass_bool_t, size_t};
 
@@ -25,7 +23,7 @@ use std::os::raw::c_char;
 use std::sync::Arc;
 
 pub struct CassRowsResultIterator<'result> {
-    iterator: TypedRowIterator<'result, 'result, Row>,
+    iterator: TypedRowIterator<'result, 'result, CassRawRow<'result, 'result>>,
     result_metadata: &'result CassResultMetadata,
     current_row: Option<CassRow<'result>>,
 }
@@ -53,7 +51,9 @@ impl CassResultIterator<'_> {
                     None
                 }
             })
-            .map(|row| CassRow::from_row_and_metadata(row, rows_result_iterator.result_metadata));
+            .map(|row: CassRawRow| {
+                CassRow::from_row_and_metadata(row.columns, rows_result_iterator.result_metadata)
+            });
 
         rows_result_iterator.current_row = new_row;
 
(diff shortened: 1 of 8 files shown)


muzarski:
Rebased on master - updated cover letter (previous first two commits were merged as a part of #237 )

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -119,6 +126,41 @@ impl CassResultMetadata {
     }
 }
 
+pub(crate) struct CassRawRow<'result> {
+    pub(crate) columns: Vec<CassRawValue<'result>>,
+}
+
```
Lorak-mmk: I'm curious, is that how cpp-drivier represents row too?
The alternative would be to have it in some (even more) deserialized form.
It would get rid of Vec (allocation), but instead force deserialization again when calling e.g. cass_row_get_column_by_name

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -203,6 +245,74 @@ mod row_with_self_borrowed_metadata {
...
+impl<'frame, 'metadata, 'result> DeserializeValue<'frame, 'metadata> for CassRawValue<'result>
+where
+    'metadata: 'result,
+    'frame: 'result,
+{
+    fn type_check(_typ: &ColumnType) -> Result<(), TypeCheckError> {
+        Ok(())
+    }
```
Lorak-mmk: Why did you choose to have those bounds instead of having 2 lifetime parameters on CassRawValue?
I assume there is some reason, but it is not explained in any comment or in commit message.

The same applies to `CassRawRow`

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -203,6 +245,74 @@ mod row_with_self_borrowed_metadata {
...
+                }
+                ColumnType::Tuple(types) => Ok(Some(types.len())),
+                ColumnType::UserDefinedType { definition, .. } => {
+                    Ok(Some(definition.field_types.len()))
+                }
+                _ => Ok(None),
+            })
+            .transpose()?
+            .flatten();
+
```
Lorak-mmk: I think this piece of code deserves a proper comment.
- Remind the reader how is empty FrameSlice different from None
- Describe cpp-driver semantics of functions that use `item_count`
- Do we always receive all fields of UDT? Or is it possible for some to be omitted (null, empty etc)? If there may be a divergence, which item_count (in the type vs in the value) does cpp-driver use?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/query_result.rs
muzarski: I'm not sure why I introduced this many (3) lifetime paramterers for this `impl`. This should just be `impl<'result> DeserializeValue<'result, 'result> for CassRawValue<'result>` (same for CassRawRow).

I don't see a point in distinguishing between `'frame` and `'metadata` lifetimes. Both frame and metadata are borrowed from `CassResult`, thus `'result` lifetime.

scylla-rust-wrapper/src/query_result.rs
muzarski: cpp-driver holds `Vector<Value>` as well.

So your suggestion is:
- represent row as e.g. `ColumnIterator` (or something that can be cheaply converted to `ColumnIterator`)
- in `cass_row_get_column_by_name` we would iterate over the `ColumnIterator` and search for column with given name
- analogously for `cass_row_get_column`, just that we return the column at given index (Iterator::nth()).

At first glance, I see no API that would clash with such solution. Should I implement it?

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk: I agree it is not necessary to distinguish them when looking at the whole codebase and how those types are used.

I'm not sure how I feel about that as a reason for not distinguishing them.
Does `CassRawValue` care if `typ` and `slice` come from the same place? Probably not. So maybe it would be a better abstraction if it did not require them to be from the same place.
Would it complicate some other code or have other downsides?

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk: I'm not saying that this alternative is better. It was more of a way to show that this is not the only possible solution, and so there should be a comment / commit message explaining why this solution was chosen.

scylla-rust-wrapper/src/query_result.rs
muzarski: > I think this piece of code deserves a proper comment.
> 
>     * Remind the reader how is empty FrameSlice different from None

Thanks for mentioning it. I totally forgot to check how `cpp-driver` handles the empty values (empty FrameSlice for `Emptiable` types). I found this: https://github.com/scylladb/cpp-driver/blob/master/src/decoder.cpp#L152-L161. So if the type is `Emptiable`, and the `FrameSlice` is empty, we should convert it to `None` (i.e. treat is as null).

I'll adjust the code, and add the TODOs regarding adding support for empty values - e.g. `cass_value_is_empty`, `cass_data_type_is_emptiable`, `cass_collection_append_empty_value`, `cass_statement_add_empty_value` (with proper typechecking) etc. This can be introduces as an extension to cpp-driver's API.

>     * Describe cpp-driver semantics of functions that use `item_count`

Ok.

>     * Do we always receive all fields of UDT? Or is it possible for some to be omitted (null, empty etc)? 

Let's consider 4 states the serialized UDT value can be in (the type that `UdtIterator` returns is `Option<Option<FrameSlice>>`:
1. value is present, non-null, non-empty - `Some(Some(frame_slice))` where `!frame_slice::is_empty()`.
2. value is present, null - `Some(None)`
3. value is present, empty - `Some(Some(frame_slice))` where `frame_slice::is_empty()`.
4. value is missing from the serialized form (but corresponding field was provided in the type - `None`.

Cases 1 and 2 are straightforward.
Case 3 is addressed by the logic I linked above (treating empty value of `Emptiable` type as null).
Case 4, from what I understand, is handled [here](https://github.com/scylladb/cpp-driver/blob/master/src/decoder.cpp#L143) by calling [decode_int32](https://github.com/scylladb/cpp-driver/blob/master/src/decoder.hpp#L169-L175). If there are missing bytes, `decode_int32` emits an error log, and returns `false` (`CHECK_REMAINING` does just that). In result, the `Decoder::decode_value` fallbacks to default constructor for `Value`, ~which constructs a null `Value`~ which constructs a `value` such that `!value.is_valid()`. Then, the corresponding iterator on `next()` returns falls.

I need to adjust my current `CassUdtIterator` code to case 4, because currently I treat missing values as nulls (instead of failing). 

> If there may be a divergence, which item_count (in the type vs in the value) does cpp-driver use?

https://github.com/scylladb/cpp-driver/blob/master/src/value.cpp#L201-L208 <- for tuples and UDTs we shall use the count defined in the type.

muzarski: And I'll obviously add some comments to the code, most likely with the links to the corresponding cpp-driver parts.

muzarski:
v1.1: Addressed @Lorak-mmk comments. Added "empty" values handling during deserialization.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -119,6 +124,37 @@ impl CassResultMetadata {
...
+        Ok(())
+    }
+
+    fn deserialize(
+        mut row: ColumnIterator<'frame, 'metadata>,
+    ) -> Result<Self, DeserializationError> {
+        let mut columns = Vec::with_capacity(row.size_hint().0);
+        while let Some(column) = row.next().transpose()? {
+            columns.push(
+                <CassRawValue>::deserialize(column.spec.typ(), column.slice).map_err(|err| {
```
wprzytula: :question: What is this syntax? It's not an ordinary _full qualification_ (`<Typ as Trait>::method`). Then what?
Is this associated function's call? Then why not simply `CassRawValue::deserialize(...)`?

muzarski:
Update: I am currently working on implementing Rust unit tests for deserialization.

muzarski:
v1.2: Introduced deserialization unit tests for native, collection and tuple types. I'll implement UDT tests tomorrow.

muzarski:
v1.3: Added deserialization test for udt iterator - it covers the scenario where there is more fields metadata than in serialized form.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -203,6 +239,127 @@ mod row_with_self_borrowed_metadata {
...
+        fn type_check(_typ: &ColumnType) -> Result<(), TypeCheckError> {
+            Ok(())
+        }
+
+        fn deserialize(
+            typ: &'metadata ColumnType<'metadata>,
+            v: Option<FrameSlice<'frame>>,
+        ) -> Result<Self, DeserializationError> {
+            // Handle the "empty" values. cpp-driver fallbacks to null if the value is "empty", and the type
+            // is non-string/byte and emptiable.
```
wprzytula: :pick: readability:

```suggestion
            // Handle the "empty" values. cpp-driver fallbacks to null if:
            // 1. the value is "empty", AND
            // 2. the type non-string/byte and emptiable.
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -203,6 +239,127 @@ mod row_with_self_borrowed_metadata {
...
+                        Ok(Some(length))
+                    }
+                    ColumnType::Tuple(types) => Ok(Some(types.len())),
+                    ColumnType::UserDefinedType { definition, .. } => {
+                        Ok(Some(definition.field_types.len()))
+                    }
+                    _ => Ok(None),
+                })
+                .transpose()?
+                .flatten();
```
wprzytula: :recycle: For readability, let's extract this to a helper function `determine_value_item_count(&ColumnType, FrameSlice)`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -383,42 +390,16 @@ pub(crate) mod cass_raw_value {
...
     pub(crate) value_type: Arc<CassDataType>,
 }
 
-impl FFI for NewCassValue<'_> {
+impl FFI for CassValue<'_> {
     type Origin = FromRef;
 }
 
-impl<'result> NewCassValue<'result> {
+impl<'result> CassValue<'result> {
```
wprzytula: :recycle: I strongly advice to prefix old names with `Legacy` and introduce new entities under the target name at the first place. You took the other way (introduce new entities with a `New` prefix and only later rename it). The difference is important when tracking back changes - `git blame` shows the rename in your approach and real changes in my approach as the last change.

wprzytula: Ah, only now have I seen this:
> Note: I cannot rename current CassValue to LegacyCassValue because it is used in the public C API.

Hmm, can't anything be done about this?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -31,11 +31,15 @@ pub enum CassResultKind {
...
+    // Arc: shared with first_row (yoke).
+    pub(crate) shared_data: Arc<CassRowsResultSharedData>,
+    pub(crate) first_row: Option<RowWithSelfBorrowedResultData>,
+}
+
+pub(crate) struct CassRowsResultSharedData {
+    pub(crate) raw_rows: DeserializedMetadataAndRawRows,
+    // Arc: shared with CassPrepared
+    pub(crate) metadata: Arc<CassResultMetadata>,
 }
```
wprzytula, line 47: :thought_balloon: I dislike that we introduce another allocation & indirection here.
IIUC, the reason for that is to allow owned sharing the frame bytes (technically, the whole `DeserializedMetadataAndRawRows`) between `CassRowsResult` (borrowed by `CassResultIterator`, created in `cass_iterator_from_result()`) and `RowWithSelfBorrowedResultData` (representing the first row, which can be examined using `cass_result_first_row()`). I understand (correct me if I'm wrong) that there are therefore only going to be two clones of `Arc<CassRowsResultSharedData>`.

Assuming that, we could just make `DeserializedMetadataAndRawRows` clonable in Rust driver. Actually, both variants of `ResultMetadataHolder`, that is, `SelfBorrowedMetadataContainer` and `Arc`, implement `Clone`, so it's as simple as adding missing `#[derive(Clone)]` for `ResultMetadataHolder` and `DeserializedMetadataAndRawRows`. What we get is that we can simply clone `DeserializedMetadataAndRawRows` into `RowWithSelfBorrowedResultData`, avoiding the allocation of `CassRowsResultSharedData`. WDYT?

wprzytula (changes requested):
Posting the first portion of review comments.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -197,45 +200,64 @@ mod row_with_self_borrowed_metadata {
...
-                .next()
-                .transpose()?
-                .map(|row: Row| Self::new_from_row_and_metadata(row, metadata));
+            #[derive(Error, Debug)]
+            enum AttachError {
+                #[error(transparent)]
+                CassErrorResult(#[from] CassErrorResult),
+                #[error("No rows in the result set")]
+                NoRows,
+            }
```
wprzytula: :question: What's the point of introducing such a fully-fledged error type, together with `Error` impl? It doesn't seem to be displayed anywhere. Instead, it's destructured shortly after creation. Is the approach warranted by the `Yoke::try_attach_to_cart` requirements?
:thought_balloon:  If not, I'd rather return `Option<Result<_, DeserializationError>>` from `Yoke::try_attach_to_cart` in the very first place. If it requires `Result<T>` as its return type, then `Result<_, Option<_>, DeserializationError>` - it's easily `transpose()`able.

scylla-rust-wrapper/src/query_result.rs
muzarski: I definitely wanted to write `<_ as DeserializeValue>`. Will fix that.

scylla-rust-wrapper/src/query_result.rs
muzarski: >  I understand (correct me if I'm wrong) that there are therefore only going to be two clones of `Arc<CassRowsResultSharedData>`.

That's correct. One clone owned by `CassRowsResult`, the other owned by `RowWithSelfBorrowedResultData`.

> Assuming that, we could just make `DeserializedMetadataAndRawRows` clonable in Rust driver. Actually, both variants of `ResultMetadataHolder`, that is, `SelfBorrowedMetadataContainer` and `Arc`, implement `Clone`, so it's as simple as adding missing `#[derive(Clone)]` for `ResultMetadataHolder` and `DeserializedMetadataAndRawRows`. What we get is that we can simply clone `DeserializedMetadataAndRawRows` into `RowWithSelfBorrowedResultData`, avoiding the allocation of `CassRowsResultSharedData`. WDYT?

Well, true but it's not that straightforward. I already considered that.

Notice that `Yoke<Y, C>` requires that `C: StableDeref` which is an unsafe [trait](https://docs.rs/stable_deref_trait/1.2.0/stable_deref_trait/trait.StableDeref.html). We would need to come up with some wrapper type e.g.:
```rust
struct RawRowsAndMetadataWrapper {
     raw_rows: DeserializedMetadataAndRawRows,
     metadata: Arc<CassResultMetadata>,
}
```

This type would need to implement (unsafely `StableDeref`). So then, we need to come up with (and carefully design) a `Deref` implementation for such type. My first idea is `Target` to be sth like (TypedRowIterator, ResultMetadata) but I am not sure whether it would be possible and (most importantly!!!) safe. So I decided to go with the safe option for now. I can try playing around with it if you want.

scylla-rust-wrapper/src/query_result.rs
muzarski: Well, I couldn't come up with anything. I certainly can't just replace `CassValue` with `NewCassValue` as it would result in one huge un-reviewable commit. We could always squash the changes after the review, so the replacement appears in one commit.

scylla-rust-wrapper/src/query_result.rs
muzarski: > 💭 If not, I'd rather return `Option<Result<_, DeserializationError>>` from `Yoke::try_attach_to_cart` in the very first place.

It requires `Result`.

> If it requires `Result<T>` as its return type, then `Result<_, Option<_>, DeserializationError>` - it's easily `transpose()`able.

Ok, so after error handling we end up with `Yoke<Option<T>, C>` and I don't like this. I prefer `Option<Yoke<T, C>>`. Why? In the first case, the backing cart (of type `C`) is held in memory for no reason if the yoke'd value is `None`. Maybe I missed something, and there is an API to map `Yoke<Option<T>, C>` to `Option<Yoke<T, C>>`.

scylla-rust-wrapper/src/query_result.rs
muzarski: > My first idea is `Target` to be sth like (TypedRowIterator, ResultMetadata) 

This won't work. `as_deref()` would need to return `&(TypedRowITerator, CassResultMetadata)`, but `DeserializedMetadataAndRawRows::rows_iter()` returns **owned** `TypedRowIterator`. We would need to come up with something else.

Notice that `as_deref()` cannot return `&(DeserializedMetadataAndRawRows, CassResultMetadata)` neither. This would be unsafe - if our type wrapper is moved, the address of `raw_rows: DeserializedMetadataAndRawRows` can be changed. This breaks the assumption of `StableDeref`. Unless we assume that our wrapper is never moved (which may actually be the case, but I'm not entirely sure as of now).

wprzytula: Keep in mind that `CassResult` is already always stored behind an `Arc`. Knowing that, we can leverage that stable deref to either:
1) employ `Yoke`, or
2) use thorough `unsafe`
and get rid of all needless allocations and even `Arc` clones.

C API already gives us strong enough lifetime guarantees; what we only have to do is to limit the Rust inner API not to introduce bugs accidentally.

scylla-rust-wrapper/src/query_result.rs
muzarski: I removed the `derive(Error)` in favor of just `impl From<CassErrorResult> for AttachError`.

muzarski:
v1.4: Addressed @wprzytula comments

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -203,6 +240,136 @@ mod row_with_self_borrowed_metadata {
...
+        }
+
+        fn deserialize(
+            typ: &'metadata ColumnType<'metadata>,
+            v: Option<FrameSlice<'frame>>,
+        ) -> Result<Self, DeserializationError> {
+            // Handle the "empty" values. cpp-driver fallbacks to null if:
+            // 1. the value is "empty", AND
+            // 2. the type non-string/byte and emptiable.
+            // TODO: Extend the C API and support "empty" values.
```
Lorak-mmk: Missing `is`? `the type non-string` -> `the type is non-string`.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -464,6 +464,84 @@ impl<'result> CassMapIterator<'result> {
     }
 }
 
+pub struct CassUdtIterator<'result> {
+    iterator: UdtIterator<'result, 'result>,
+    metadata: &'result [(String, Arc<CassDataType>)],
+    current_entry: Option<CassUdtIteratorEntry<'result>>,
+}
```
Lorak-mmk: Here we can see that CassDataType is borrowed from 'result.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -419,6 +420,63 @@ impl FFI for CassValue {
     type Origin = FromRef;
 }
 
+pub struct NewCassValue<'result> {
+    pub(crate) value: CassRawValue<'result, 'result>,
+    pub(crate) value_type: Arc<CassDataType>,
+}
```
Lorak-mmk: And here the Arc is cloned. It should not be a problem to make it an Arc reference.
Is it possible to make it just a reference? It depends: is there an API that lets user get CassDataType from CassValue? If so, then it is not possible.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/query_result.rs
muzarski: Great observation. This can definitely be changed to `&'result Arc<CassDataType>`. This way we simply borrow the `Arc` and save one `Arc::clone()` call.

scylla-rust-wrapper/src/query_result.rs
muzarski: Done as you suggested - firstly renamed `CassValue` -> `LegacyCassValue`, and then re-introduced `CassValue`. Hope this helps during the review.

muzarski:
v1.5: Addressed comments

`CassValue` now borrows the data_type from result (i.e. holds `&'result Arc<CassDataType>`).

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -811,95 +811,77 @@ pub unsafe extern "C" fn cass_iterator_get_column<'result>(
...
+                .and_then(|state| {
+                    let current_entry = map_collection_iterator.iterator.current_entry.as_ref();
+                    match state {
+                        CassMapCollectionIteratorState::Key => {
+                            current_entry.map(|entry| RefFFI::as_ptr(&entry.0))
+                        }
+                        CassMapCollectionIteratorState::Value => {
+                            current_entry.map(|entry| RefFFI::as_ptr(&entry.1))
+                        }
+                    }
```
Lorak-mmk: Nit: Maybe a slightly better way (less duplication) would be:
```
let v = match ....;
current_entry.map(|entry| RefFFI::as_ptr(&v))
```

Lorak-mmk (approved):

Lorak-mmk (approved):

scylla-rust-wrapper/src/query_result.rs
wprzytula: I don't want to hold back merging this PR, as it's correct and its introduced overhead is not large. However, knowing that the C API gives us possibility not to introduce that overhead at all, I suggest opening an issue that mentions this possible optimisation.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -72,6 +76,56 @@ impl CassRowIterator<'_> {
...
+            CassDataTypeInner::List { typ, .. } | CassDataTypeInner::Set { typ, .. } => {
+                // Expect: `typ` is an `Option<Arc<CassDataType>>`. It is None, for untyped set or list.
+                // There is no such thing as untyped list/set in CQL protocol, thus we do not expect it in result metadata.
+                typ.as_ref()
+                    .expect("List or set type provided from result metadata should be fully typed!")
+            }
+            _ => {
+                panic!("Expected list or set type. Typecheck should have prevented such scenario!")
+            }
+        };
```
wprzytula: :warning: :question: I can see `unsafe`  without explanation why it's safe to do so. Moreover, it's not clear to me why, at least without diving into `CassRawValue` and its invariants.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -126,6 +126,72 @@ impl<'result> CassListlikeIterator<'result> {
...
+
+        let key_value_types = match unsafe { value.value_type.get_unchecked() } {
+            CassDataTypeInner::Map { typ, .. } => match typ {
+                MapDataType::KeyAndValue(key_type, val_type) => {
+                    (key_type, val_type)
+                }
+                _ => panic!("Untyped or half-typed map received in result metadata. Something really bad happened!"),
+            },
+            _ => panic!("Expected map type. Typecheck should have prevented such scenario!"),
+        };
```
wprzytula, line 434: ditto: nontrivial `unsafe`.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -191,6 +191,212 @@ impl CassCollectionIterator<'_> {
...
+                        kind: value::BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
+                    },
+                )),
+                // Value is missing.
+                None => Err(DeserializationError::new(TupleMissingValue(pos))),
+            };
+
+            Some(raw_res)
+        }
+    }
```
wprzytula: :thought_balloon: Rust driver currently allows deserializing CQL Tuples to Rust tuples (unless you deserialize to the type-erased CqlValue - then it's a `Vec<CqlValue>`). This is a limitation that may hurt those who don't know the size of the tuple in advance.
_Well, the question (as always) is, how do people actually use CQL - what do they do with their tuples?_
Anyway, Rust driver's users already can partly deserialize CQL collections to the half-raw form of `<Collection>Iterator`, and then perform any desired way of custom deserialization (including deserialization to their own, perhaps dynamic, structures in a way that is less wasteful than just throwing everything into `Vec<CqlValue>`). I believe **we should allow the same for tuples**. To reiterate, the viable options are:
1. [present] Rust tuple of concrete (known a priori, at compile time) types - efficient yet limiting.
3. [present] `CqlValue::Tuple(Vec<CqlValue>)` - inefficient and requires matching even if there is some more compile-time knowledge on the underlying types/length.
4. [my proposed addition] `TupleIterator` - manual yet flexible and efficient.

:thought_balloon: Now I have yet more ideas. We could allow:
1. deserializing collections and/or tuples into fixed-size arrays. The type check would fail if the length of the collection is not equal to the expected length.
2. deserializing collections into `SmallVec` - why not allow such optimisation for our users? `smallvec` is already API-stable BTW.
3. deserializing CQL Map into `Vec<(Key, Value)>`. I'm afraid this may collide with the general `Vec<T>` trait impl, though.
4. I also considered buffer/allocation re-use for collections, that is, deserializing collections into a pre-exising `Vec` (like into `&mut Vec`). However, here we are limited by the `DeserializeValue::deserialize()` signature, which does not accept any kind of a buffer. A possible workaround would be to deserialize to an immediate `CollectionDeserializer`, which would expose a `deserialize_to_buffer(&mut Vec)` method.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -463,6 +463,84 @@ impl<'result> CassMapIterator<'result> {
...
+impl<'result> CassUdtIterator<'result> {
+    fn new_from_value(
+        value: &'result CassValue<'result>,
+    ) -> Result<Self, NonNullDeserializationError> {
+        let udt_iterator = value.get_non_null::<UdtIterator>()?;
+
+        let metadata = match unsafe { value.value_type.get_unchecked() } {
+            CassDataTypeInner::UDT(udt) => udt.field_types.as_slice(),
+            _ => panic!("Expected UDT type. Typecheck should have prevented such scenario!"),
+        };
```
wprzytula, line 494: ditto: `unsafe`

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -463,6 +463,84 @@ impl<'result> CassMapIterator<'result> {
...
+        })
+    }
+
+    fn next(&mut self) -> bool {
+        // Handle scenario where underlying iterator is exhausted.
+        let Some(((_field_name, field_type), deser_result)) = self.iterator.next() else {
+            return false;
+        };
+
+        // Handle the deserialization error from underyling iterator.
```
wprzytula: :wrench:

wprzytula:
```suggestion
        // Handle the deserialization error from underlying iterator.
```

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -76,7 +76,7 @@ impl CassRowIterator<'_> {
     }
 }
 
-/// An iterator created from [`cass_iterator_from_collection()`] with list or set provided as a value.
+/// Iterator created from [`cass_iterator_from_collection()`] with list or set provided as a value.
```
wprzytula: :pick: This change looks redundant.

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -788,7 +788,7 @@ pub unsafe extern "C" fn cass_iterator_get_row<'result>(
 #[no_mangle]
 pub unsafe extern "C" fn cass_iterator_get_column<'result>(
     iterator: CassBorrowedSharedPtr<CassIterator<'result>, CConst>,
-) -> CassBorrowedSharedPtr<'result, LegacyCassValue, CConst> {
+) -> CassBorrowedSharedPtr<'result, CassValue<'result>, CConst> {
```
wprzytula: :question: Why in this function `iterator` does not get an explicit `'result` lifetime, whereas...

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -811,95 +811,77 @@ pub unsafe extern "C" fn cass_iterator_get_column<'result>(
 
 #[no_mangle]
 pub unsafe extern "C" fn cass_iterator_get_value<'result>(
-    iterator: CassBorrowedSharedPtr<CassIterator<'result>, CConst>,
-) -> CassBorrowedSharedPtr<'result, LegacyCassValue, CConst> {
+    iterator: CassBorrowedSharedPtr<'result, CassIterator<'result>, CConst>,
+) -> CassBorrowedSharedPtr<'result, CassValue<'result>, CConst> {
```
wprzytula: :question: ...in this function `iterator` does get an explicit `'result` lifetime?

scylla-rust-wrapper/src/iterator.rs
```diff
@@ -1180,91 +1137,85 @@ pub unsafe extern "C" fn cass_iterator_from_row<'result>(
...
+    if RefFFI::is_null(&value) {
         return BoxFFI::null_mut();
     }
 
-    let item_count = unsafe { cass_value_item_count(value.borrow()) };
-    let item_count = match unsafe { cass_value_type(value.borrow()) } {
-        CassValueType::CASS_VALUE_TYPE_MAP => item_count * 2,
-        _ => item_count,
-    };
+    let value_type = unsafe { cass_value_type(value.borrow()) };
```
wprzytula, line 1109: :wrench: Justify `unsafe` and prove correctness.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -477,21 +484,21 @@ impl ToCassError for NonNullDeserializationError {
...
-            LegacyCassValue {
-                value: val.map(|col_val| get_column_value(col_val, &column_type)),
-                value_type: column_type,
-            }
+        .map(|(value, col_spec)| {
+            let value_type = &col_spec.data_type;
+            CassValue { value, value_type }
         })
         .collect()
 }
```
wprzytula: :thought_balloon: We're accepting a `Vec<CassRawValue>` only to produce `Vec<CassValue>`. This got me thinking if we really have to materialize `CassRawValue` in a `Vec`. So I learned how `CassRawRow` differs from `CassRow` and my findings are the following: `CassRow` holds `CassValue` and `CassRawRow` holds `CassRawValue`, with the former containing `CassDataType` and the latter not. As a side note, I believe these observations could be put in docstrings / comments.
Perhaps we could make `CassRawRow` contain a `ColumnIterator` instead of `Vec<CassRawValue>`? Then, we would construct `CassValue` straight during manual iteration over `ColumnIterator`, in `create_cass_row_columns()`.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -950,91 +993,89 @@ pub unsafe extern "C" fn cass_value_get_decimal(
...
-        Some(Value::RegularValue(CqlValue::Text(s))) => unsafe {
-            write_str_to_c(s.as_str(), output, output_size)
-        },
-        Some(_) => return CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE,
-        None => return CassError::CASS_ERROR_LIB_NULL_VALUE,
-    }
+    let val: &CassValue = val_ptr_to_ref_ensure_non_null!(value);
+
+    // It seems that cpp driver doesn't check the type - you can call _get_string
+    // on any type and get internal represenation. I don't see how to do it easily in
```
wprzytula: :camping:

```suggestion
    // on any type and get internal representation. I don't see how to do it easily in
```

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -0,0 +1,368 @@
...
+        value: do_deserialize::<CassRawValue>(&typ, &bytes),
+        value_type: &data_type,
+    };
+    let value_ptr = RefFFI::as_ptr(&cass_value);
+
+    // Check that we retrieved the original value.
+    let actual = cass_value_ptr_to_rust_type(value_ptr.borrow());
+    assert_eq!(to_serialize_non_null, actual);
+
+    if unsafe { cass_value_is_null(value_ptr.borrow()) == 0 } {
```
wprzytula, line 248: :wrench: Justify `unsafe` and prove correctness.

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -0,0 +1,368 @@
...
+        // Compare the serialized bytes with the bytes from `cass_value_get_bytes`.
+        let actual_bytes = unsafe {
```
wprzytula, line 250: ditto: `unsafe`

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -0,0 +1,368 @@
...
+    test_deserialize::<CqlValue, _>(
+        ColumnType::Native(NativeType::Boolean),
+        CqlValue::Empty,
+        |value_ptr| unsafe {
+            if cass_value_is_null(value_ptr) != 0 {
+                CqlValue::Empty
+            } else {
+                CqlValue::Boolean(true)
+            }
+        },
```
wprzytula, line 385: ditto: `unsafe`

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -0,0 +1,368 @@
...
+    <T as DeserializeValue<'frame, 'metadata>>::type_check(typ).unwrap();
+    let mut frame_slice = FrameSlice::new(bytes);
+    let value = frame_slice.read_cql_bytes().unwrap();
+    <T as DeserializeValue<'frame, 'metadata>>::deserialize(typ, value).unwrap()
+}
+
+/// Given a pointer to a `CassValue`, this trait allows you to convert it to a Rust type.
+trait FromCassValuePtr {
+    fn from_cass_value_ptr(value: CassBorrowedSharedPtr<CassValue, CConst>) -> Self;
+}
```
wprzytula, line 71: :100: :rocket:  Wow, this is a truly impressive unit test suite! Congrats!

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -366,3 +377,377 @@ fn test_deserialize_value_native() {
...
 }
+
+#[test]
+fn test_deserialize_tuple_iterator() {
+    let typ = ColumnType::Tuple(vec![
+        ColumnType::Native(NativeType::Int),
+        ColumnType::Native(NativeType::Text),
+        ColumnType::Native(NativeType::Float),
+    ]);
+    let to_serialize = (42i32, String::from("Ala ma kota"), 42.42f32);
```
wprzytula: :pick: readability nit: you can insert underscores for cleaner separation in numeric literals. See:

wprzytula:
```suggestion
    let to_serialize = (42_i32, String::from("Ala ma kota"), 42.42_f32);
```

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -366,3 +377,377 @@ fn test_deserialize_value_native() {
...
+    let bytes = Bytes::from(do_serialize(&to_serialize, &typ));
+    let data_type = Arc::new(get_column_type(&typ));
+    let cass_value = CassValue {
+        value: do_deserialize::<CassRawValue>(&typ, &bytes),
+        value_type: &data_type,
+    };
+    let value_ptr = RefFFI::as_ptr(&cass_value);
+
+    unsafe {
+        assert_eq!(cass_value_item_count(value_ptr.borrow()), 3);
```
wprzytula, line 406: ditto: justify `unsafe`.

scylla-rust-wrapper/src/ser_de_tests.rs
```diff
@@ -366,3 +377,377 @@ fn test_deserialize_value_native() {
...
+    let bytes = Bytes::from(do_serialize(&to_serialize, &typ));
+    let data_type = Arc::new(get_column_type(&typ));
+    let cass_value = CassValue {
+        value: do_deserialize::<CassRawValue>(&typ, &bytes),
+        value_type: &data_type,
+    };
+    let value_ptr = RefFFI::as_ptr(&cass_value);
+
+    unsafe {
+        assert_eq!(cass_value_item_count(value_ptr.borrow()), 4);
```
wprzytula, line 459: ditto: justify `unsafe`

scylla-rust-wrapper/src/iterator.rs
wprzytula: cc @Lorak-mmk

scylla-rust-wrapper/src/query_result.rs
wprzytula: Opened https://github.com/muzarski/cpp-rust-driver/pull/1.

wprzytula (changes requested):

scylla-rust-wrapper/src/iterator.rs
muzarski: `CassRowIterator` borrows the `CassValue` from `CassRow` (`&'result CassValue<'result>`). No need to introduce a constraint on the iterator.

scylla-rust-wrapper/src/iterator.rs
muzarski: And here the iterator types **own** the `CassValue` (e.g. `current_value: Option<CassValue<'result>>`), and so we need an additional constraint on the iterator itself.

scylla-rust-wrapper/src/query_result.rs
muzarski: Ok, I'll open an issue regarding this

scylla-rust-wrapper/src/query_result.rs
muzarski: Thanks! I cherry-picked the commit on top of existing commits. I hope that's fine.

scylla-rust-wrapper/src/iterator.rs
muzarski: In other words: `CassIterator` that is passed to this method must **outlive** the returned `CassValue`.

scylla-rust-wrapper/src/iterator.rs
muzarski: Here, we do not require the `CassIterator` to outlive the `CassValue` - the value is already owned by `CassRow` (in `Vec<>`).

muzarski:
v2:
- cherry-picked @wprzytula commit which gets rid of additional `Vec<>` allocation per-row. Much thanks!
- sprinkled some `SAFETY` comments in `iterator.rs` and `query_result.rs`. Added a module-level `SAFETY` comment in deserialization test module.
- addressed remaining minor comments from @wprzytula

scylla-rust-wrapper/src/query_result.rs
muzarski: > Ok, I'll open an issue regarding this

https://github.com/scylladb/cpp-rust-driver/issues/244

scylla-rust-wrapper/src/iterator.rs
wprzytula: Thanks!

wprzytula (approved):
LGTM! Good job!

We can merge this as-is. However, let's not forget about the `TupleIterator` that is introduced here, which I deem better aligned to `scylla-cql`. Especially that it may be useful for `scylla` users, too, for manual deserialization with custom optimisations.
Unless @Lorak-mmk disagrees, I encourage you @muzarski to open a PR to `scylla-cql` with `TupleIterator`.

PR #254, opened 2025-04-15
Implement local address and local port range configuration

### LINKED ISSUE #239: Implement local bind address and port range configuration
This includes following methods:
- `cass_cluster_set_local_address`
- `cass_cluster_set_local_address_n`
- `cass_cluster_set_local_port_range`

It was recently implemented on rust-driver side (https://github.com/scylladb/scylla-rust-driver/pull/1300 and https://github.com/scylladb/scylla-rust-driver/pull/1303) - low hanging fruit.

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/239

This pr implements `cass_cluster_set_local_addres[_n]` and `cass_cluster_set_local_port_range`.

## Integration tests:
There will be a follow-up PR which enables the integration tests that we "unlocked" by implementing these. The reason is that it requires to adjust some test logic (e.g. logs filtering) and error handling. Don't want to introduce too much noise to this PR.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -24,8 +24,10 @@ use scylla::statement::{Consistency, SerialConsistency};
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::future::Future;
+use std::net::IpAddr;
 use std::num::NonZero;
 use std::os::raw::{c_char, c_int, c_uint};
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -56,6 +58,8 @@ const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_millis(5000);
 const DEFAULT_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(30);
 // - keepalive timeout is 60 secs
 const DEFAULT_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(60);
+// - default local ip address is arbitrary
+const DEFAULT_LOCAL_IP_ADDRESS: Option<IpAddr> = None;
 
 const DRIVER_NAME: &str = "ScyllaDB Cpp-Rust Driver";
 const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -227,6 +231,7 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
             .write_coalescing_delay(DEFAULT_WRITE_COALESCING_DELAY)
             .keepalive_interval(DEFAULT_KEEPALIVE_INTERVAL)
             .keepalive_timeout(DEFAULT_KEEPALIVE_TIMEOUT)
+            .local_ip_address(DEFAULT_LOCAL_IP_ADDRESS)
     };
 
     BoxFFI::into_ptr(Box::new(CassCluster {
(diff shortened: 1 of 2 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -482,6 +487,52 @@ pub unsafe extern "C" fn cass_cluster_set_port(
...
+        None
+    } else {
+        // SAFETY: We assume that user provides valid pointer and length.
+        match unsafe { ptr_to_cstr_n(ip, ip_length) } {
+            Some(ip_str) => match IpAddr::from_str(ip_str) {
+                Ok(addr) => Some(addr),
+                Err(err) => {
+                    tracing::error!("Failed to resolve ip address <{}>: {}", ip_str, err);
+                    return CassError::CASS_ERROR_LIB_HOST_RESOLUTION;
+                }
```
Lorak-mmk: Is there really some resolving going on in `IpAddr::from_str`? I thought it only parses the address from string, so HOST_RESOLUTION is not expected here imo.

wprzytula: +1, `LIB_BAD_PARAMS` makes more sense.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -975,12 +1026,89 @@ mod tests {
...
+
+            // non-numeric address
+            {
+                assert_cass_error_eq!(
+                    cass_cluster_set_local_address(cluster_raw.borrow_mut(), c"foo".as_ptr()),
+                    CassError::CASS_ERROR_LIB_HOST_RESOLUTION
+                );
+            }
+
+            cass_cluster_free(cluster_raw);
```
wprzytula: :wrench: Let's add a test case for a non-utf8 string passed.

wprzytula (changes requested):

scylla-rust-wrapper/src/cluster.rs
muzarski: It's not about what makes sense here, but what cpp-driver does (I should have put a comment denoting that).

I was confused myself, because:
1. The documentation says:
```
 * @param[in] name IP address to bind, or empty string for no binding.
 * Only numeric addresses are supported; no resolution is done.
 * @return CASS_OK if successful, otherwise an error occurred.
```
**no resolution is done**
2: And then there is this: https://github.com/scylladb/cpp-driver/blob/master/src/cluster_config.cpp#L504-L508.
The `Address` constructor indeed does no resolution - it simply translates the human readable ip address to bytes. So this inlines with the documentation. But then, the method "is_valid_and_**resolved**" is called an if it returns false, the `CASS_ERROR_LIB_HOST_RESOLUTION` is returned to the user.

So this terminology inconsistency comes from cpp-driver. Should we fix that? If so, I'll need to adjust an integration test that tests this scenario (in a follow-up PR).

wprzytula: IIUC, cpp-driver returns a misleading error kind. Right? If so, let's fix that, both in the code and in integration tests.

muzarski:
v2: `cass_cluster_set_local_address` now returns `LIB_BAD_PARAMS` if user provided unparsable ip address string. Added the test case for non-utf8 bytes slice.

wprzytula (approved):

muzarski:
v2.1: Rebased on `master` (fixed some import conflict)

wprzytula (approved):

Lorak-mmk (approved):

wprzytula (approved):

Issue #132, opened 2024-06-30
cpp-rust-driver test parity with original cpp-driver 


roydahan:
We would like the cpp-rust-driver to have all the original cpp driver tests running as part of CI.
This is an umbrella issue (epic) to include all the current & future tasks to get a working running CI and update here about the status of it.
(e.g. How many tests we have that are not passing (if any)).
```[tasklist]
### Tasks
- [x] Make sure we have all tests from original cpp driver (let's write here the amount of tests)
- [x] Enable (if not already) all tests to run as part of CI --> how many failing compared to original driver
- [ ] Start openning issues for failing tests / classes / missing functionality
- [ ] https://github.com/scylladb/cpp-rust-driver/issues/144
```

https://docs.google.com/spreadsheets/d/1CuXBS9CkBpwIoLmPOXPPS3_V-mMffnt_EWGJjNPDKgA/edit?usp=sharing

Migrated to Jira: [DRIVER-138](https://scylladb.atlassian.net/browse/DRIVER-138)

[DRIVER-138]: https://scylladb.atlassian.net/browse/DRIVER-138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

wprzytula:
IIRC, some Datastax tests require some functionality besides the API defined by `cassandra.h`, whereas cpp-rust-driver only supports that API. Can you confirm that @Lorak-mmk?

Lorak-mmk:
There is additional testing API
https://github.com/scylladb/cpp-driver/blob/master/src/testing.hpp
https://github.com/scylladb/cpp-driver/blob/master/src/testing.cpp

but it should not be a problem - we can implement it or change the tests to avoid using it (or modify it).

Regarding the issue: We already run integration tests, with both Scylla and C* - or at least the subset that works.
You can find the list of used tests in CI yaml: https://github.com/scylladb/cpp-rust-driver/blob/master/.github/workflows/cassandra.yml

roydahan:
Yes, I know we run some.
First, we would like to run all and get a mapping of the amount of tests failing compared to original cpp driver and start mapping also missing functionality.

In this epic, we will collect the task items to enable/disable relevant tests and track the status till we reach parity.

roydahan:
@muzarski can you please update your progress here?

muzarski:
[...]

Issue #130, opened 2024-06-13
Compile on Windows


mykaul:
We should use GH action and make sure this compiles on Windows as well.

Lorak-mmk:
Our fork of original cpp-driver didn't support Windows afaik so it's not a compatibility issue. Is there a specific reason for this change / client request etc?
Drivers are usually used server-side which is dominated by linux. Is there any other driver apart from python-driver where we test Windows in CI?
I have very bad memories from trying to use windows in CI for C++ projetcs, and in general from trying to maintain Windows compatibility.
While this project is mostly Rust there is still a fair share of C++ code. I fear the moment when we would need to debug and fix some windows-specific issue...

mykaul:
Customer request. See PRs on the previous CPP driver.

Lorak-mmk:
Do you know the use case of this customer? Do they use Windows Server for Scylla clients?

Lorak-mmk:
Scylla itself doesn't support Windows (but that would be a "fun" hackathon project :D ), so it can only be an issue in environment where some servers run Linux and some run Windows Server - I'm curious why that would be the case.

mykaul:
> Scylla itself doesn't support Windows (but that would be a "fun" hackathon project :D ), so it can only be an issue in environment where some servers run Linux and some run Windows Server - I'm curious why that would be the case.

Why won't the client run on Windows?

mykaul:
See https://github.com/scylladb/cpp-driver/pull/83 for ref from the old driver.

Lorak-mmk:
> > Scylla itself doesn't support Windows (but that would be a "fun" hackathon project :D ), so it can only be an issue in environment where some servers run Linux and some run Windows Server - I'm curious why that would be the case.
> 
> Why won't the client run on Windows?

You mean the end-application that runs on some untrusted users PC? Because that would require such user to have access to the database.

mykaul:
> > > Scylla itself doesn't support Windows (but that would be a "fun" hackathon project :D ), so it can only be an issue in environment where some servers run Linux and some run Windows Server - I'm curious why that would be the case.
> > 
> > 
> > Why won't the client run on Windows?
> 
> You mean the end-application that runs on some untrusted users PC? Because that would require such user to have access to the database.

I have no idea if it's trusted or not. It's not different than a Linux client. It's a different OS, that some people actually use - and may need access to Scylla from it. Probably some Windows Server edition.

dkropachev:
[...]

PR #255, opened 2025-04-16
Adjust and enable some `ControlConnection` integration tests

muzarski:
Depends on: https://github.com/scylladb/cpp-rust-driver/pull/254. Start review from `cass_error: remove misleading comment`.

To enable the tests I firstly needed to do some adjustments:
- most importantly, the adjustments to error conversions - it turns out we misused the `UNABLE_TO_CONNECT` error in places where `LIB_NO_HOSTS_AVAILABLE` should be returned. This is adjusted.
- Integration tests logic:
   - one test expected `cass_cluster_set_local_address` to return `LIB_HOST_RESOLUTION` when unparsable ip address is provided. We decided that it should return `LIB_BAD_PARAMS` in such case - I adjusted the integration test case to this.
   - After error conversion adjustments, remaining tests only required to adjust the log criteria in the test - logs emitted by rust-driver are slightly different than the ones emitted by cpp-driver.

## Tests
Tests enabled:
- ConnectUsingInvalidIpAddress
- ConnectUsingInvalidPort
- ConnectUsingUnresolvableLocalIpAddress
- ConnectUsingUnbindableLocalIpAddress
- ConnectUsingValidLocalIpAddressButInvalidRemote
    
Tests that we keep disabled:
- TopologyChange <- some error during `ccm add` command (need to investigate)
- FullOutage <- requires cass_cluster_set_constant_reconnect
- TerminatedUsingMultipleIoThreadsWithError <- requires cass_cluster_set_num_threads_io

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -47,10 +47,6 @@ impl ToCassError for ExecutionError {
 
 impl ToCassError for ConnectionPoolError {
     fn to_cass_error(&self) -> CassError {
-        // I know that TranslationError (corresponding to CASS_ERROR_LIB_HOST_RESOLUTION)
-        // is hidden under the ConnectionPoolError.
-        // However, we still have a lot work to do when it comes to error conversion.
-        // I will address it, once we start resolving all issues related to error conversion.
         CassError::CASS_ERROR_LIB_UNABLE_TO_CONNECT
     }
 }
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -47,7 +47,7 @@ impl ToCassError for ExecutionError {
 
 impl ToCassError for ConnectionPoolError {
     fn to_cass_error(&self) -> CassError {
-        CassError::CASS_ERROR_LIB_UNABLE_TO_CONNECT
+        CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE
     }
 }
 
@@ -75,7 +75,7 @@ impl ToCassError for RequestAttemptError {
             }
             RequestAttemptError::UnableToAllocStreamId => CassError::CASS_ERROR_LIB_NO_STREAMS,
             RequestAttemptError::BrokenConnectionError(_) => {
-                CassError::CASS_ERROR_LIB_UNABLE_TO_CONNECT
+                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE
             }
             RequestAttemptError::BodyExtensionsParseError(_) => {
                 CassError::CASS_ERROR_LIB_MESSAGE_ENCODE
--- a/scylla-rust-wrapper/src/cass_error.rs
+++ b/scylla-rust-wrapper/src/cass_error.rs
@@ -33,7 +33,7 @@ impl ToCassError for ExecutionError {
             ExecutionError::BadQuery(bad_query) => bad_query.to_cass_error(),
             ExecutionError::RequestTimeout(_) => CassError::CASS_ERROR_LIB_REQUEST_TIMED_OUT,
             ExecutionError::EmptyPlan => CassError::CASS_ERROR_LIB_INVALID_STATE,
-            ExecutionError::MetadataError(_) => CassError::CASS_ERROR_LIB_INVALID_STATE,
+            ExecutionError::MetadataError(e) => e.to_cass_error(),
             ExecutionError::ConnectionPoolError(e) => e.to_cass_error(),
             ExecutionError::PrepareError(e) => e.to_cass_error(),
             ExecutionError::LastAttemptError(e) => e.to_cass_error(),
@@ -160,11 +160,41 @@ impl ToCassError for NewSessionError {
                 CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE
             }
             NewSessionError::EmptyKnownNodesList => CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-            NewSessionError::MetadataError(_) => CassError::CASS_ERROR_LIB_INVALID_STATE,
+            NewSessionError::MetadataError(e) => e.to_cass_error(),
             NewSessionError::UseKeyspaceError(_) => {
                 CassError::CASS_ERROR_LIB_UNABLE_TO_SET_KEYSPACE
             }
             // NS error is non_exhaustive
+            _ => CassError::CASS_ERROR_LAST_ENTRY,
+        }
+    }
+}
+
+impl ToCassError for MetadataError {
+    fn to_cass_error(&self) -> CassError {
+        match self {
+            MetadataError::ConnectionPoolError(e) => e.to_cass_error(),
+            MetadataError::FetchError(e) => match &e.error {
+                // Server bug - invalid CQL type in system table.
+                MetadataFetchErrorKind::InvalidColumnType(_) => {
+                    CassError::CASS_ERROR_LIB_UNEXPECTED_RESPONSE
+                }
+                MetadataFetchErrorKind::PrepareError(e) => e.to_cass_error(),
+                MetadataFetchErrorKind::SerializationError(e) => e.to_cass_error(),
+                MetadataFetchErrorKind::NextRowError(_) => {
+                    CassError::CASS_ERROR_LIB_UNEXPECTED_RESPONSE
+                }
+
+                // non_exhaustive
+                _ => CassError::CASS_ERROR_LAST_ENTRY,
+            },
+            // Remaining errors indicate a serious server bug - e.g. all peers have empty token lists.
+            MetadataError::Keyspaces(_)
+            | MetadataError::Peers(_)
+            | MetadataError::Tables(_)
+            | MetadataError::Udts(_) => CassError::CASS_ERROR_LIB_UNEXPECTED_RESPONSE,
+
+            // non_exhaustive
             _ => CassError::CASS_ERROR_LAST_ENTRY,
         }
     }
--- a/tests/src/integration/tests/test_control_connection.cpp
+++ b/tests/src/integration/tests/test_control_connection.cpp
@@ -132,9 +132,9 @@ CASSANDRA_INTEGRATION_TEST_F(ControlConnectionTests, ConnectUsingInvalidIpAddres
   CHECK_FAILURE;
 
   // Attempt to connect to the server using an invalid IP address
-  logger_.add_critera("Unable to establish a control connection to host "
-                      "1.1.1.1 because of the following error: Underlying "
-                      "connection error: Connection timeout");
+  logger_.add_critera("Could not fetch metadata, error: "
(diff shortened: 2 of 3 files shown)


Lorak-mmk (approved):
> Note: As we can see, rust-driver does not provide the node address
> either in the log message or in the error type.
> This is something we should consider adding to rust-driver.

Please open an issue about it

> - TopologyChange <- some error during `ccm add` command (need to investigate)

In a follow up or before merging this?

muzarski:
> > Note: As we can see, rust-driver does not provide the node address
> > either in the log message or in the error type.
> > This is something we should consider adding to rust-driver.
> 
> Please open an issue about it

Ok.

> > * TopologyChange <- some error during `ccm add` command (need to investigate)
> 
> In a follow up or before merging this?

I'll open separate issue regarding this. I'm not sure how much effort this would require - is it just a fix on cpp-rust-driver side? Or maybe there is some bug in ccm. Who knows...

muzarski:
Rebased on `master`.

Lorak-mmk (approved):

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -47,7 +47,7 @@ impl ToCassError for ExecutionError {
 
 impl ToCassError for ConnectionPoolError {
     fn to_cass_error(&self) -> CassError {
-        CassError::CASS_ERROR_LIB_UNABLE_TO_CONNECT
+        CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE
     }
 }
```
wprzytula: commit `cass_error: replace UNABLE_TO_CONNECT with NO_HOSTS_AVAILABLE`:
:100: Thanks for such a thorough explanation! This is what makes commits a pure pleasure to review.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -170,6 +170,30 @@ impl ToCassError for NewSessionError {
...
+            MetadataError::FetchError(e) => match &e.error {
+                // Server bug - invalid CQL type in system table.
+                MetadataFetchErrorKind::InvalidColumnType(_) => {
+                    CassError::CASS_ERROR_LIB_INVALID_STATE
+                }
+                MetadataFetchErrorKind::PrepareError(e) => e.to_cass_error(),
+                MetadataFetchErrorKind::SerializationError(e) => e.to_cass_error(),
+                MetadataFetchErrorKind::NextRowError(_) => {
+                    CassError::CASS_ERROR_LIB_UNEXPECTED_RESPONSE
+                }
```
wprzytula: :question: what makes the `InvalidColumnType` variant different from `NextRowError` that warrants returning `CASS_ERROR_LIB_INVALID_STATE` instead of `CASS_ERROR_LIB_UNEXPECTED_RESPONSE`? I'd guess that `InvalidColumnType` is as much result of an unexpected response (i.e., an invalid response) as `NextRowError`.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -170,6 +170,30 @@ impl ToCassError for NewSessionError {
...
+
+                // non_exhaustive
+                _ => CassError::CASS_ERROR_LAST_ENTRY,
+            },
+            // Remaining errors indicate a serious server bug - e.g. all peers have empty token lists.
+            _ => CassError::CASS_ERROR_LIB_INVALID_STATE,
```
wprzytula: :recycle: Could you please list all remaining variants instead of relying on the catch-all? This is prone to bugs when new variants are added in the Rust driver.

tests/src/integration/tests/test_control_connection.cpp
```diff
@@ -204,7 +204,9 @@ CASSANDRA_INTEGRATION_TEST_F(ControlConnectionTests, ConnectUsingUnbindableLocal
   CHECK_FAILURE;
 
   // Attempt to connect to the server using an unbindable local IP address
-  logger_.add_critera("Unable to bind local address: address not available");
+  logger_.add_critera("Could not fetch metadata, error: "
+                      "Control connection pool error: The pool is broken; "
+                      "Last connection failed with: Cannot assign requested address (os error 99)");
```
wprzytula: :question: :thinking: Is this going to work on Windows (and other OSes)? I'm afraid it's not. Especially the `os error 99` is specific to Linux, I suppose.
A possible fix would be not to rely on a particular formulation of the OS-level error. I'm not sure, though, how to determine that the underlying `IoError` was truly about being unable to assign the address.

tests/src/integration/tests/test_control_connection.cpp
```diff
@@ -236,8 +236,9 @@ CASSANDRA_INTEGRATION_TEST_F(ControlConnectionTests,
   // Attempt to connect to the server using an valid local IP address
   // but invalid remote address. The specified remote is not routable
   // from the specified local.
-  logger_.add_critera("Unable to establish a control connection to host "
-                      "1.1.1.1 because of the following error:");
+  logger_.add_critera("Could not fetch metadata, error: "
+                      "Control connection pool error: The pool is broken; "
+                      "Last connection failed with: Invalid argument (os error 22)");
```
wprzytula: ditto: portability

wprzytula: BTW, do we target Windows? If so, then we should run unit & integration tests in the CI on Windows, too.

wprzytula (changes requested):

scylla-rust-wrapper/src/cass_error.rs
muzarski: To be quite honest, IDK. I don't even know what exactly "INVALID_STATE" is - in my head, this is when Scylla does something really bad - for example having a bad CQL type in the system table column.

But I have nothing against returning `UNEXPECTED_RESPONSE` here. Should I do it for other variants as well, e.g. when all peers have empty tokens?

scylla-rust-wrapper/src/cass_error.rs
muzarski: Well this will be prone to the bugs even if I add them (non_exhaustive), but I'll add them anyway to aid readability.

tests/src/integration/tests/test_control_connection.cpp
muzarski: I believe we want to support it eventually: https://github.com/scylladb/cpp-rust-driver/issues/130

scylla-rust-wrapper/src/cass_error.rs
muzarski: > Well this will be prone to the bugs even if I add them (non_exhaustive)

Actually, not quite. New error variants would then always return "INVALID_STATE", instead of "LAST_ENTRY".

muzarski:
v1.1: Addressed @wprzytula comments.

scylla-rust-wrapper/src/cass_error.rs
wprzytula: > But I have nothing against returning `UNEXPECTED_RESPONSE` here. Should I do it for other variants as well, e.g. when all peers have empty tokens?

No idea. I'd just like to have some order and reasoning behind our error variants. I leave this decision for you.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -174,6 +170,36 @@ impl ToCassError for NewSessionError {
...
+                }
+
+                // non_exhaustive
+                _ => CassError::CASS_ERROR_LAST_ENTRY,
+            },
+            // Remaining errors indicate a serious server bug - e.g. all peers have empty token lists.
+            MetadataError::Keyspaces(_)
+            | MetadataError::Peers(_)
+            | MetadataError::Tables(_)
+            | MetadataError::Udts(_) => CassError::CASS_ERROR_LIB_INVALID_STATE,
```
wprzytula: Now `InvalidColumnType` is considered an unexpected response, but `Keyspaces` (and similar) error is considered invalid state. I don't see the distinction. I'm lost in how those two error variants differ in meaning.

tests/src/integration/tests/test_control_connection.cpp
```diff
@@ -204,7 +204,9 @@ CASSANDRA_INTEGRATION_TEST_F(ControlConnectionTests, ConnectUsingUnbindableLocal
   CHECK_FAILURE;
 
   // Attempt to connect to the server using an unbindable local IP address
-  logger_.add_critera("Unable to bind local address: address not available");
+  logger_.add_critera("Could not fetch metadata, error: "
+                      "Control connection pool error: The pool is broken; "
+                      "Last connection failed with: Cannot assign requested address");
```
wprzytula, line 209: I'm not convinced other OSes will return the same error string (`Cannot assign requested address`). The portability issue is not solved IMO.

tests/src/integration/tests/test_control_connection.cpp
```diff
@@ -234,8 +236,9 @@ CASSANDRA_INTEGRATION_TEST_F(ControlConnectionTests,
   // Attempt to connect to the server using an valid local IP address
   // but invalid remote address. The specified remote is not routable
   // from the specified local.
-  logger_.add_critera("Unable to establish a control connection to host "
-                      "1.1.1.1 because of the following error:");
+  logger_.add_critera("Could not fetch metadata, error: "
+                      "Control connection pool error: The pool is broken; "
+                      "Last connection failed with: Invalid argument");
```
wprzytula, line 241: ditto: portability

wprzytula (changes requested):

tests/src/integration/tests/test_control_connection.cpp
muzarski: If I remove the suffix, how do you make a distinction between the error message here, and...

tests/src/integration/tests/test_control_connection.cpp
muzarski: ...there.

scylla-rust-wrapper/src/cass_error.rs
muzarski: I'll change them to `UNEXPECTED_RESPONSE`.

tests/src/integration/tests/test_control_connection.cpp
wprzytula: That's indeed the core issue that we are facing... This is what I meant here:

> A possible fix would be not to rely on a particular formulation of the OS-level error. **I'm not sure, though, how to determine that the underlying IoError was truly about being unable to assign the address**.

muzarski: I think that we should worry about it once we decide to run CI on windows. Since these are the tests, I have nothing against adding something like:
```c
#ifdef WIN32
#define CANNOT_ASSIGN_ADDRESS_MSG "whatever"
#else
#define CANNOT_ASSIGN_ADDRESS_MSG "Cannot assign requested address"
#endif
```

wprzytula: OK, fair enough.

wprzytula (approved):

PR #280, opened 2025-04-18
implement `cass_session_get_metrics` and enable some integration tests

### LINKED ISSUE #257: Implement metrics on cpp-rust-driver side (`cass_session_get_metrics`)
We already have everything we need on rust-driver side.

One thing to keep in mind: there are some `CassMetrics` fields that are marked as **deprecated** in `cassandra.h`. Rust-driver does not support them (which is totally fine), so we should 0-initialize those values. We could remove them altogether, but then we wouldn't be able to claim binary compatibility with cpp-driver. We can always remove them later, once users already migrate to cpp-rust-driver (this will be API breaking change).

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/257

This PR implements `cass_session_get_metrics` and enables two integration tests.

##  Metrics tests
To be quite honest, I thought we'll be able to enable more tests, but:
    
1. StatsConnections

Requires `cass_cluster_set_num_threads_io`, `cass_cluster_set_core
_connections_per_host`
and `cass_cluster_set_constant_reconnect`.

2. ErrorsConnectionTimeouts

Requires `cass_cluster_set_core_connections_per_host`.

3. SpeculativeExecutionRequests

Requires `cass_session_get_speculative_execution_metrics`.

4. Requests

This one is interesting. It turns out that cpp-driver stores latency
stats as microseconds, while rust-driver stores them as milliseconds.

Because of that, the mean and median latency is rounded to 0 (at least for my machine).
The test expects them to be greater than 0, which makes sense assuming
the driver collects the stats with microsecond precision.

~I'm not sure how to address this one. Is there any way to force
higher >=1ms latencies in the test?~ Thanks to @Lorak-mmk suggestion, I was able to apply the `HistoryListener` hack in the test - this allows us to enforce `>=1ms` latencies for the requests.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,11 @@ SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :UseKeyspaceCaseSensitiveTests.Integration_Cassandra_ConnectWithKeyspace)
 endif
 
+ifndef SCYLLA_NO_VALGRIND_TEST_FILTER
+SCYLLA_NO_VALGRIND_TEST_FILTER := $(subst ${SPACE},${EMPTY},AsyncTests.Integration_Cassandra_Simple\
+:HeartbeatTests.Integration_Cassandra_HeartbeatFailed)
+endif
+
 ifndef CASSANDRA_TEST_FILTER
 CASSANDRA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :BasicsTests.*\
@@ -70,6 +75,11 @@ CASSANDRA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :*NoCompactEnabledConnection\
 :PreparedMetadataTests.Integration_Cassandra_AlterProperlyUpdatesColumnCount\
 :UseKeyspaceCaseSensitiveTests.Integration_Cassandra_ConnectWithKeyspace)
+endif
+
+ifndef CASSANDRA_NO_VALGRIND_TEST_FILTER
+CASSANDRA_NO_VALGRIND_TEST_FILTER := $(subst ${SPACE},${EMPTY},AsyncTests.Integration_Cassandra_Simple\
+:HeartbeatTests.Integration_Cassandra_HeartbeatFailed)
 endif
 
 ifndef CCM_COMMIT_ID
@@ -226,7 +236,7 @@ endif
 	@echo "Running integration tests on scylla ${SCYLLA_VERSION}"
 	valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${SCYLLA_TEST_FILTER}"
 	@echo "Running timeout sensitive tests on scylla ${SCYLLA_VERSION}"
-	build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="AsyncTests.Integration_Cassandra_Simple"
+	build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${SCYLLA_NO_VALGRIND_TEST_FILTER}"
 
 run-test-integration-cassandra: prepare-integration-test download-ccm-cassandra-image install-java8-if-missing
 ifdef DONT_REBUILD_INTEGRATION_BIN
@@ -237,7 +247,7 @@ endif
 	@echo "Running integration tests on cassandra ${CASSANDRA_VERSION}"
 	valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite build/cassandra-integration-tests --version=${CASSANDRA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${CASSANDRA_TEST_FILTER}"
 	@echo "Running timeout sensitive tests on cassandra ${CASSANDRA_VERSION}"
-	build/cassandra-integration-tests --version=${CASSANDRA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="AsyncTests.Integration_Cassandra_Simple"
+	build/cassandra-integration-tests --version=${CASSANDRA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${CASSANDRA_NO_VALGRIND_TEST_FILTER}"
 
 run-test-unit: install-cargo-if-missing _update-rust-tooling
 	@cd ${CURRENT_DIR}/scylla-rust-wrapper; cargo test
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :LoggingTests.*\
 :PreparedMetadataTests.*\
 :UseKeyspaceCaseSensitiveTests.*\
+:MetricsTests.Integration_Cassandra_ErrorsRequestTimeouts\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
 :HeartbeatTests.Integration_Cassandra_HeartbeatFailed\
 :ControlConnectionTests.Integration_Cassandra_TopologyChange\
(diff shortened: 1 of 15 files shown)


muzarski:
`HeartbeatFailed` test failed. It worked for me locally - I'll investigate.

muzarski:
Ok, it looks like the `HearbeatFailed` test is totally unreliable under `vagrind`. This makes sense, because this test is timeout-sensitive and `valgrind` increases the execution time significantly. This is similar to the test that used `cass_future_wait_timed`.

I've enabled the test under `[C*/SCYLLA]_NO_VALGRIND_TEST_FILTER`.

muzarski:
v1.1: Enabled `Requests` test as well. This test does not work for me locally without valgrind (because sub-millisecond latencies are rounded to 0ms). But it seems to always pass under valgrind (latencies are higher than 1ms). I think it's worth giving it a shot and run it in our CI - the tests are running under valgrind and I expect GH actions runner to be slower than my local machine.

muzarski:
> v1.1: Enabled `Requests` test as well. This test does not work for me locally without valgrind (because sub-millisecond latencies are rounded to 0ms). But it seems to always pass under valgrind (latencies are higher than 1ms). I think it's worth giving it a shot and run it in our CI - the tests are running under valgrind and I expect GH actions runner to be slower than my local machine.

v1.2: Well, I gave it a shot, but it failed for one of the Scylla releases..... So it unfortunately is flaky  (and will be, unless we address microsecond granularity on the rust-driver side). Disabled this test again.

Lorak-mmk (approved):
> I'm not sure how to address this one. Is there any way to force
> higher >=1ms latencies in the test?

What about your idea with setting coalescing to > 1ms?

muzarski:
> What about your idea with setting coalescing to > 1ms?

It is not reliable enough.. Notice that there may be a burst of requests that are handled (and appended to request queue) just before the `coalescing` timer expires. Then such delay has almost no effect on the `mean` request latency value.

Lorak-mmk:
Hmm.... When is the "before" measurement of the latency done? If it is before LBP is consulted you could put sleep(1ms) in custom LBP.

Lorak-mmk:
Another possible place for that would be history listener.

muzarski:
> Hmm.... When is the "before" measurement of the latency done? If it is before LBP is consulted you could put sleep(1ms) in custom LBP.

I see that we measure latency per-attempt. We completely ignore failing attempts when measuring the latency. Is it expected behaviour?

Lorak-mmk:
I have no idea tbh. @piodul ?

This means that both approaches (LBP, history) are not viable, right?

muzarski:
> This means that both approaches (LBP, history) are not viable, right?

LBP will not work for sure. But `history` hack might work actually. I'll try it out.

piodul:
> I have no idea tbh. @piodul ?

Me neither. What do other drivers measure?

muzarski:
v1.3: Applied the `HistoryListener` hack to increase the request latency during the test. The idea behind this is explained in the commit message. Thanks to that, we can enable `Requests` metrics test. All changes are contained in the last commit.

Lorak-mmk (approved):

muzarski:
Rebased on master

muzarski:
Rebased on master (sped up ci)

scylla-rust-wrapper/src/session.rs
```diff
@@ -571,6 +573,91 @@ pub unsafe extern "C" fn cass_session_get_schema_meta(
...
+        Err(MetricsError::Empty) => (0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
+        Err(e) => {
+            tracing::error!("Failed to get metrics snapshot: {}", e);
+            return;
+        }
+    };
+
+    const MILLIS_TO_MICROS: u64 = 1000;
+    // SAFETY: We assume that user provided valid CassMetrics pointer.
+    unsafe {
```
wprzytula, line 634: :question: Don't we prefer to implement some FFI for `CassMetrics` instead?

tests/src/integration/tests/test_heartbeat.cpp
```diff
@@ -83,19 +83,22 @@ CASSANDRA_INTEGRATION_TEST_F(HeartbeatTests, HeartbeatDisabled) {
...
   Cluster cluster =
       default_cluster().with_connection_heartbeat_interval(1).with_connection_idle_timeout(5);
   connect(cluster);
 
   cass_uint64_t initial_connections = session_.metrics().stats.total_connections;
   pause_node(2);
   start_timer();
-  while (session_.metrics().stats.total_connections >= initial_connections &&
+
+  CassMetrics metrics = session_.metrics();
```
wprzytula, line 95: :pick:

wprzytula:
```suggestion
  CassMetrics const metrics = session_.metrics();
```

Makefile
```diff
@@ -25,6 +25,7 @@ SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :LoggingTests.*\
 :PreparedMetadataTests.*\
 :UseKeyspaceCaseSensitiveTests.*\
+:MetricsTests.Integration_Cassandra_ErrorsRequestTimeouts\
```
wprzytula, line 28: > Is there any way to force higher >=1ms latencies in the test?

:thought_balloon: Perhaps we could employ a UDF that spins in a loop long enough?

scylla-rust-wrapper/src/integration_testing.rs
```diff
@@ -54,3 +62,57 @@ pub unsafe extern "C" fn testing_cluster_get_contact_points(
...
+    fn log_attempt_start(
+        &self,
+        _request_id: RequestId,
+        _speculative_id: Option<SpeculativeId>,
+        _node_addr: SocketAddr,
+    ) -> AttemptId {
+        // Sleep to simulate a delay in the request
+        std::thread::sleep(self.0);
+        AttemptId(0)
+    }
```
wprzytula, line 91: :100: Nice hack!

scylla-rust-wrapper/src/integration_testing.rs
```diff
@@ -54,3 +62,57 @@ pub unsafe extern "C" fn testing_cluster_get_contact_points(
...
+    let sleep_time = Duration::from_millis(sleep_time_ms);
+    let history_listener = Arc::new(SleepingHistoryListener(sleep_time));
+
+    match &mut BoxFFI::as_mut_ref(statement_raw).unwrap().statement {
+        BoundStatement::Simple(inner) => inner.query.set_history_listener(history_listener),
+        BoundStatement::Prepared(inner) => Arc::make_mut(&mut inner.statement)
+            .statement
+            .set_history_listener(history_listener),
+    }
+}
```
wprzytula, line 118: :thinking: It's a pity we include this in the library unconditionally.

wprzytula: Can't we use some `#ifdef`?

wprzytula (changes requested):

scylla-rust-wrapper/src/session.rs
muzarski: The pointers that implement `FFI` are the pointers that we (as a driver) create and return to the user. In most cases (if not all), the type is opaque for the user.

Here, this is not the case. It's just a pointer to the mutable `CassMetrics` object user created. Most probably, this is just a pointer to the object on the stack. Notice that `CassMetrics` is explicitly defined in `cassandra.h` (not opaque).

```c++
CassMetrics metrics = {};
cass_session_get_metrics(session_ptr, &metrics);
```

We cannot really make any assumptions about the pointer. It's user's responsibility to ensure that provided pointer points to some valid memory.

scylla-rust-wrapper/src/integration_testing.rs
muzarski: This is true for all methods defined in `integration_testing.rs`. I'll open an issue regarding this.

muzarski: https://github.com/scylladb/cpp-rust-driver/issues/285

tests/src/integration/tests/test_heartbeat.cpp
muzarski: Why? We update the metrics in a loop.

Makefile
muzarski: Aren't UDFs an experimental feature in Scylla? Anyway, we already found a hack that works on client side (history listener).

muzarski:
@wprzytula I addressed all of your commits. There was nothing to change.

tests/src/integration/tests/test_heartbeat.cpp
wprzytula: My bad, nvm.

wprzytula (approved):

PR #281, opened 2025-04-23
Connection pool size configuration

### LINKED ISSUE #240: Implement `cass_cluster_set_core_connections_per_host`
Low hanging fruit - directly corresponds to `SessionConfig::connection_pool_size`.

We should also remove `cass_cluster_set_max_connections_per_host` method from the API - it is already deprecated.

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/240

~Depends on: https://github.com/scylladb/cpp-rust-driver/pull/280 (I need metrics support to implement one integration test).~ <- already on master

Implements:
- `cass_cluster_set_core_connections_per_host`
- `cass_cluster_set_core_connections_per_shard` (as an extension to cpp-driver API)

### Existing ErrorsConnectionTimeouts metrics test
```c++
CASSANDRA_INTEGRATION_TEST_F(MetricsTests, ErrorsConnectionTimeouts) {
  CHECK_FAILURE;

  Session session =
      default_cluster().with_core_connections_per_host(2).with_connect_timeout(1).connect(
          "", false); // Quick connection timeout and no assertion

  CassMetrics metrics = session.metrics();
  EXPECT_GE(2u, metrics.errors.connection_timeouts);
}
```

I don't know who implemented this test, but (IMO) it's just wrong. So my understanding of intentions is:
1. we set 2 connections per host
2. we set low connect timeout
3. we expect at least 2 connection timeouts registered in the metrics

If those were the intentions, the `EXPECT_GE(2u, metrics.errors.connection_timeouts);` is wrong - it should be the other way around. Currently it expects `2 >= connection_timeouts`.

Ok, so let's say we fix it. It still won't work with rust driver. The `Session` object will not even be built - we will fail to open a control connection (because of timeouts) and fail to fetch the metadata. In particular, we won't even be able to call `Session::get_metrics()` - the `cass_session_get_metrics` will return early with some log error message.

Thus, I'm not enabling this test.

Fun fact: in my local setup, this test passes. Even though, the connect timeout is low, it is not triggered. And then, the `metrics.errors.connection_timeouts` is 0 - this means that aforementioned assertion passes as well.

### My new `StatsShardConnections` metrics test
I implemented simple test where we configure a pool size to be 2 connections per shard (`cass_cluster_set_core_connections_per_shard`). It checks whether all connections are registered in the metrics. The expected value is at least `nr_hosts * nr_shards * 2`.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -12,7 +12,7 @@ use openssl::ssl::SslContextBuilder;
 use openssl_sys::SSL_CTX_up_ref;
 use scylla::client::execution_profile::ExecutionProfileBuilder;
 use scylla::client::session_builder::SessionBuilder;
-use scylla::client::{SelfIdentity, WriteCoalescingDelay};
+use scylla::client::{PoolSize, SelfIdentity, WriteCoalescingDelay};
 use scylla::frame::Compression;
 use scylla::policies::load_balancing::{
     DefaultPolicyBuilder, LatencyAwarenessBuilder, LoadBalancingPolicy,
@@ -25,7 +25,7 @@ use std::collections::HashMap;
 use std::convert::TryInto;
 use std::future::Future;
 use std::net::IpAddr;
-use std::num::NonZero;
+use std::num::{NonZero, NonZeroUsize};
 use std::os::raw::{c_char, c_int, c_uint};
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


muzarski:
Rebased on master

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -451,6 +451,32 @@ pub unsafe extern "C" fn cass_cluster_set_connect_timeout(
...
+    match NonZeroUsize::new(num_connections as usize) {
+        Some(non_zero_conns) => {
+            cluster.session_builder.config.connection_pool_size = PoolSize::PerHost(non_zero_conns);
+            CassError::CASS_OK
+        }
+        None => {
+            tracing::error!(
+                "Provided zero connections to cass_cluster_set_core_connections_per_host"
+            );
+            CassError::CASS_ERROR_LIB_BAD_PARAMS
```
wprzytula: :pick: One message has `!` at the end and the other does not.

wprzytula (approved):

wprzytula (approved):

Lorak-mmk (approved):

PR #282, opened 2025-04-24
ci: build in release

muzarski:
In hope to speed up the CI process.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ build-integration-test-bin:
 		echo "Building integration test binary to ${INTEGRATION_TEST_BIN}";\
   		mkdir "${BUILD_DIR}" >/dev/null 2>&1 || true;\
 		cd "${BUILD_DIR}";\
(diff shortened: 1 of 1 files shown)


muzarski:
I'm comparing with: https://github.com/scylladb/cpp-rust-driver/pull/280

For Scylla ITs, we see significant improvement from 14-17 minutes to 9-12 minutes. Unfortunately, the slowest job (cassandra tests) is still slow... We see an improvement (21 minutes to 17 minutes), so we at least gain something. The cassandra tests are bounded by slow `ccm` cluster operations. Take for example: `ConsistencyThreeNodeClusterTests.Integration_Cassandra_TwoNodesDecommissioned (146438 ms)` - over 2 minutes for 1 test case !!!!!!!

I still think it is worth merging. We gain around 4 minutes per CI run.

wprzytula (approved):

PR #287, opened 2025-04-28
Implement `cass_retry_policy_logging_new` and enable some exec profile integration tests

### LINKED ISSUE #260: Implement retry logging policy
In other words, implement `cass_retry_policy_logging_new`.

This is just a simple wrapper over existing retry policy, which logs the underlying retry decision at INFO level.

muzarski:
The logging retry policy is used in tests in order to observe and verify retry decisions.
Therefore, implementing it allows us enable more integration tests.

As the semantics of retry policy differ between C++ driver and Rust driver, the logs are a bit different, so the tests need to be adjusted.
- C++:
```c++
// On request error
    LOG_INFO("Retrying on request error at consistency %s (initial consistency: %s, error: %s, "
               "retries: %d)",

    LOG_INFO("Ignoring request error (initial consistency: %s, error: %s, retries: %d)",
               cass_consistency_string(cl), error->message().to_string().c_str(), num_retries);

// On unavailable
    LOG_INFO("Retrying on unavailable error at consistency %s (initial consistency: %s, required "
               "replica: %d, alive replica: %d, retries: %d)",
               cass_consistency_string(decision.retry_consistency()), cass_consistency_string(cl),
               required, alive, num_retries);

    LOG_INFO("Ignoring unavailable error (initial consistency: %s, required replica: %d, alive "
               "replica: %d, retries: %d)",
               cass_consistency_string(cl), required, alive, num_retries);

// On write timeout
    LOG_INFO("Retrying on write timeout at consistency %s (initial consistency: %s, required "
               "acknowledgments: %d, received acknowledgments: %d, write type: %s, retries: %d)",
               cass_consistency_string(decision.retry_consistency()), cass_consistency_string(cl),
               required, received, cass_write_type_string(write_type), num_retries);

    LOG_INFO("Ignoring write timeout (initial consistency: %s, required acknowledgments: %d, "
               "received acknowledgments: %d, write type: %s, retries: %d)",
               cass_consistency_string(cl), required, received, cass_write_type_string(write_type),
               num_retries);

// On read timeout
     LOG_INFO("Retrying on read timeout at consistency %s (initial consistency: %s, required "
               "responses: %d, received responses: %d, data retrieved: %s, retries: %d)",
               cass_consistency_string(decision.retry_consistency()), cass_consistency_string(cl),
               required, received, data_recevied ? "true" : "false", num_retries);
     
     LOG_INFO("Ignoring read timeout (initial consistency: %s, required responses: %d, received "
               "responses: %d, data retrieved: %s, retries: %d)",
               cass_consistency_string(cl), required, received, data_recevied ? "true" : "false",
               num_retries);
```

- CPP-Rust (based on Rust driver's semantics):
```rust
match &decision {
      RetryDecision::RetrySameTarget(consistency) => tracing::info!(
          "Retrying on the same target; Error: {}; Initial Consistency: {:?}; New Consistency: {:?}",
          error,
          initial_consistency,
          consistency
      ),
      RetryDecision::RetryNextTarget(consistency) => tracing::info!(
          "Retrying on the next target; Error: {}; Initial Consistency: {:?}; New Consistency: {:?}",
          error,
          initial_consistency,
          consistency
      ),
      RetryDecision::IgnoreWriteError => {
          tracing::info!("Ignoring write error; Error: {}", error)
      }
```

Fixes: https://github.com/scylladb/cpp-rust-driver/issues/260
Ref: #132 

## Pre-review checklist

- [x] I have enabled appropriate tests in `Makefile` in `gtest_filter`.

### DIFF
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -109,11 +109,11 @@ pub unsafe extern "C" fn cass_batch_set_retry_policy(
 
     let maybe_arced_retry_policy: Option<Arc<dyn scylla::policies::retry::RetryPolicy>> =
         ArcFFI::as_ref(retry_policy).map(|policy| match policy {
-            CassRetryPolicy::DefaultRetryPolicy(default) => {
+            CassRetryPolicy::Default(default) => {
                 default.clone() as Arc<dyn scylla::policies::retry::RetryPolicy>
             }
-            CassRetryPolicy::FallthroughRetryPolicy(fallthrough) => fallthrough.clone(),
-            CassRetryPolicy::DowngradingConsistencyRetryPolicy(downgrading) => downgrading.clone(),
+            CassRetryPolicy::Fallthrough(fallthrough) => fallthrough.clone(),
+            CassRetryPolicy::DowngradingConsistency(downgrading) => downgrading.clone(),
         });
 
     Arc::make_mut(&mut batch.state)
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -114,6 +114,7 @@ pub unsafe extern "C" fn cass_batch_set_retry_policy(
             }
             CassRetryPolicy::Fallthrough(fallthrough) => fallthrough.clone(),
             CassRetryPolicy::DowngradingConsistency(downgrading) => downgrading.clone(),
+            CassRetryPolicy::Logging(logging) => Arc::clone(logging) as _,
         });
 
     Arc::make_mut(&mut batch.state)
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -115,6 +115,8 @@ pub unsafe extern "C" fn cass_batch_set_retry_policy(
             CassRetryPolicy::Fallthrough(fallthrough) => fallthrough.clone(),
             CassRetryPolicy::DowngradingConsistency(downgrading) => downgrading.clone(),
             CassRetryPolicy::Logging(logging) => Arc::clone(logging) as _,
+            #[cfg(cpp_integration_testing)]
+            CassRetryPolicy::Ignoring(ignoring) => Arc::clone(ignoring) as _,
         });
 
     Arc::make_mut(&mut batch.state)
--- a/tests/src/integration/tests/test_exec_profile.cpp
+++ b/tests/src/integration/tests/test_exec_profile.cpp
@@ -25,8 +25,7 @@ class ExecutionProfileTest : public Integration {
 public:
   ExecutionProfileTest()
       : insert_(NULL)
-      //, child_retry_policy_(IgnoreRetryPolicy::policy()) // Used for counting retry
-      , child_retry_policy_(DefaultRetryPolicy())
+      , child_retry_policy_(IgnoreRetryPolicy()) // Used for counting retry
       , logging_retry_policy_(child_retry_policy_)
       , skip_base_execution_profile_(false) {
     // LWTs do not work with tablets.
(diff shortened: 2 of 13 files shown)


wprzytula:
Rebased on master.

wprzytula (approved):

tests/src/integration/tests/test_exec_profile.cpp
```diff
@@ -645,8 +648,9 @@ CASSANDRA_INTEGRATION_TEST_F(ExecutionProfileTest, RetryPolicy) {
 
   // Execute a simple query with assigned profile
   statement.set_execution_profile("retry_policy");
-  result = session_.execute(statement);
-  ASSERT_EQ(CASS_OK, result.error_code());
+  // Don't expect CASS_OK.
+  result = session_.execute(statement, false);
+  ASSERT_EQ(CASS_ERROR_SERVER_UNAVAILABLE, result.error_code());
   ASSERT_EQ(1u, logger_.count());
 }
```
Lorak-mmk: This seems like a pretty significant difference. In old version the request returned Ok, but now it returns an error. Why?

wprzytula: The answer is in line 28:
```c++
 //, child_retry_policy_(IgnoreRetryPolicy::policy()) // Used for counting retry
 , child_retry_policy_(DefaultRetryPolicy())
```
As you can see, we are using a different policy than the test initially used. That's why the assumptions need to be adjusted, too.

Lorak-mmk: Why do we need to change the policy?

wprzytula: We don't need to change it. It has already been changed.
Well, you should ask yourself why it was needed: 78245baaa3d54a1e6eb5c8af4a255ef44f31c3bf.

Lorak-mmk: I probably removed it to make this project somehow compile at all - which was not an easy task in the beggining.

Do we have `IgnoreRetryPolicy` implemented? If not, can we implement it? I'm asking, because reverting the change seems better to me than modifying the test.

wprzytula: Unfortunately, `IgnoreRetryPolicy` (defined in `ignore_retry_policy.hpp`) resides in `datastax::internal::core` namespace, which is not at all present in our driver. Furthermore, `ignore_retry_policy.hpp` requires `retry_policy.hpp` from the original driver.

I suggest leaving the test as-is, using the `DefaultRetryPolicy`.

Lorak-mmk: I know we can't use the original file, but why not implement the policy ourselves?

wprzytula:
Addressed comments. The test now uses `IgnoreRetryPolicy`.

scylla-rust-wrapper/src/batch.rs
```diff
@@ -109,11 +109,11 @@ pub unsafe extern "C" fn cass_batch_set_retry_policy(
...
         ArcFFI::as_ref(retry_policy).map(|policy| match policy {
-            CassRetryPolicy::DefaultRetryPolicy(default) => {
+            CassRetryPolicy::Default(default) => {
                 default.clone() as Arc<dyn scylla::policies::retry::RetryPolicy>
             }
-            CassRetryPolicy::FallthroughRetryPolicy(fallthrough) => fallthrough.clone(),
-            CassRetryPolicy::DowngradingConsistencyRetryPolicy(downgrading) => downgrading.clone(),
+            CassRetryPolicy::Fallthrough(fallthrough) => fallthrough.clone(),
+            CassRetryPolicy::DowngradingConsistency(downgrading) => downgrading.clone(),
         });
```
Lorak-mmk: Commit: "CassRetryPolicy: cut RetryPolicy- off enum variants "

We have the same problem as in Rust Driver: clippy by default does not trigger some lints for public items. Please apply the same solution as in the Rust Driver (it was some config option).

Lorak-mmk (approved):
Thanks for implementing the ignore policy!

scylla-rust-wrapper/src/batch.rs
wprzytula: I did notice it and I'm going to fix it in a follow-up PR.

PR #291, opened 2025-04-29
execution_profile: whitelist/blacklist filtering

### LINKED ISSUE #243: Implement whitelist/blacklist filtering
Methods:
- `cass_execution_profile_set_whitelist_filtering`
- cass_execution_profile_set_whitelist_filtering_n
- cass_execution_profile_set_blacklist_filtering
- cass_execution_profile_set_blacklist_filtering_n
- cass_execution_profile_set_whitelist_dc_filtering
- cass_execution_profile_set_whitelist_dc_filtering_n
- cass_execution_profile_set_blacklist_dc_filtering
- cass_execution_profile_set_blacklist_dc_filtering_n
- cass_cluster_set_whitelist_filtering
- cass_cluster_set_whitelist_filtering_n
- cass_cluster_set_blacklist_filtering
- cass_cluster_set_blacklist_filtering_n
- cass_cluster_set_whitelist_dc_filtering
- cass_cluster_set_whitelist_dc_filtering_n
- cass_cluster_set_blacklist_dc_filtering
- cass_cluster_set_blacklist_dc_filtering_n

This should probably be implemented using `LoadBalancingPolicy`, `HostFilter` and `ControlConnectionPolicy` (https://github.com/scylladb/scylla-rust-driver/issues/1279) traits. Just `LoadBalancingPolicy` may not be enough, as in cpp-driver these filters apply for choosing the control connection as well.

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/243

This PR introduces whitelist/blacklist host filtering to the driver.

## Filtering rules
There are currently 4 filtering rules (we could introduce 2 more in the future - rack filtering):
- ip whitelist
- ip blacklist
- dc whitelist
- dc blacklist

The logic for checking if a host is accepted is following (in the specified order):
1. if `ip whitelist` is non-empty, and if it does not contain host's ip -> reject
2. if `ip blacklist` is non-empty, and if it contains host's ip -> reject
3. if `dc whitelist` is non-empty, and if it does not contain host's dc -> reject
4. if `dc blacklist` is non-empty, and if it contains host's dc -> reject

## HostFilter
Since all of the execution profiles are defined before session creation (in other words: user cannot define new profiles after session was created and connected), cpp-driver (and so, cpp-rust-driver as well) can see which hosts are disabled by all execution profiles.

If there is a host which is rejected by **all** execution profiles (including the default one), the driver does not open the connection to such host. This is exactly what we do in this PR - after collecting all of the filtering rules of each profile, we compute the filtering rules for our custom `HostFilter` implementor - namely `CassHostFilter`. This is done by computing the unions of whitelists and intersection of blacklists.

## Vec or HashSet
For now, I decided to use `Vec` to represent the set of items in the whitelists/blacklists. I'd say that we do not expect them to contain a lot of items, so I think `Vec` is more efficient than `HashSet` in this specific case. One could argue that this code is not critical (and I agree) - this is a configuration phase. I'm open to suggestions.

### ExecutionProfile suite
I have not enabled any yet - they require `cass_future_coordinator`, which is used to check the ip of host the request was routed to. TBH, I think this is a blocker for merging - I prefer to wait for the `cass_future_coordinator`, and enable the corresponding tests (and maybe implement some additional tests on my own). I still believe that this PR is ready for review, though.

### HostFilter suite
This is the suite I implemented myself - currently there are two tests that reject all nodes using execution profiles - in result, session does not have any connections to route the requests to.

### DisconnectedNullStringApiArgs suite
Enabled remaining 4 test cases from this suite - they check whether driver behaves as expected when null string is provided to filtering config methods.

## Pre-review checklist


scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -13,6 +18,55 @@ pub(crate) struct FilteringConfig {
...
+    });
+
+    if let Some(item_iter) = item_iter {
+        // We should append new entries.
+        filtering_vec.extend(item_iter);
+    } else {
+        // If the string is empty or null, we should clear the list.
+        filtering_vec.clear();
+    }
+
```
Lorak-mmk: Is that really how cpp-driver does it?

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -144,3 +219,60 @@ pub(crate) enum LoadBalancingKind {
...
+        node: NodeRef<'_>,
+        error: &RequestAttemptError,
+    ) {
+        self.child_policy
+            .on_request_failure(request, latency, node, error);
+    }
+
+    fn name(&self) -> String {
+        "FilteringLoadBalancingPolicy".to_string()
+    }
```
Lorak-mmk: We could include a name of child policy here too.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -291,3 +291,110 @@ impl HostFilter for CassHostFilter {
...
+    // Remove the items from the intersection that are not present in the other sets.
+    if let Some(intersection) = intersection.as_mut() {
+        for set in iter {
+            intersection.retain(|item| set.contains(item));
+        }
+    }
+
+    // If the intersection is empty, return None.
+    intersection.filter(|v| !v.is_empty())
+}
```
Lorak-mmk: Commit: lbp: introduce utility methods for set operations 

I think that with host filtering the performance of Vec vs HashSet is not relevant (this is not even close to hot path). What we should look at is readability. Wouldn't methods present in HashSet / BTreeSet help with implementation?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/load_balancing.rs
muzarski: https://github.com/scylladb/cpp-driver/blob/master/src/cluster_config.cpp#L350-L361

(Notice that `SAFE_STRLEN` returns 0 for null pointer)

scylla-rust-wrapper/src/load_balancing.rs
muzarski: On a high level, there is no `HashSet` API that would allow us to implement these as a "one-liners". The logic for intersection would remain the same. However, the logic for union could be simplified - we can convert each set to iterator, then flatten and collect to `HashSet`. Currently, we need a loop, so we don't insert the duplicates.

I have nothing against transitioning to `HashSet` - as you said - this is not close to hot path.

scylla-rust-wrapper/src/load_balancing.rs
muzarski: Since `cass_cluster_set_whitelist_filtering(cluster, nullptr)` clears the list (because of `SAFE_STRLEN(nullptr) == 0`), I decided that `cass_cluster_set_whitelist_filtering_n(cluster, nullptr, <non_zero_length>)` should clear the list as well (as opposed to cpp-driver, which tries to create a `String` from null pointer).

muzarski:
I have already found a bug in my `HostFilter` filtering rules computation. Consider following scenario:
We have two nodes: "ip1" and "ip2". Let's say we have two execution profiles, one with the whitelist being "ip1", the other with empty (disabled) whitelist. In result, the `HostFilter` whitelist will be a union of these -> "ip1". In result, we filter out the "ip2" node - even though it is **accepted** by at least one execution profile (the one with empty whitelist).

I'll fix it - the `nonempty_union` should return `None`, if at least one of the provided sets is empty (i.e. one of the whitelist rules is disabled). This proves that we need the tests before merging this.

Note: the `nonempty_intersection` looks good in such scenario, because if at least one of the blacklists is empty, the resulting `HostFilter` blacklist will be empty (disabled) as well.

muzarski:
Rebased on master.

muzarski:
v2:
- addressed @Lorak-mmk comments (haven't transitioned to `HashSet` yet - still to be discussed)
- fixed the bug I mentioned above
- introduced `HostFilterTest` suite and implemented two test cases where we disable all nodes (driver has no connections to route the queries).

scylla-rust-wrapper/src/load_balancing.rs
muzarski: cc: @wprzytula

muzarski:
v2.1: Previously I implemented new tests, but forgot to enable them..

scylla-rust-wrapper/src/load_balancing.rs
Lorak-mmk: Wouldn't some combination of `fold` (or similar Iterator methods) and `union` / `intersection`  HashSet methods allow to express both of those more cleanly?

muzarski: `union` and `intersection` methods take two `HashSet`s and return an iterator. I don't see a neat way to generalize it and reuse these methods to implement `union`/`intersection` for multiple sets (without needless `collect()`ing in `fold`). I think that the `retain` method is clean and efficient at the same time.

Lorak-mmk (approved):
Looks good to me. I don't really care if we use Vec or HashSet, so I'm approving now. If you decide to change the container, then I'll re-approve.

scylla-rust-wrapper/src/load_balancing.rs
Lorak-mmk: Ok

muzarski:
v2.2:
- Rebased on master
- Implemented `set_[whitelist/blacklist]_filtering` methods using `update_comma_delimited_list` utility method introduced [here](https://github.com/scylladb/cpp-rust-driver/pull/293). I removed the `set_filtering` helper function which was in my first version of this PR - it's the same as `update_comma_delimited_list`.
- Enabled remaining `DisconnectedNullStringApiArgs` tests

Lorak-mmk (approved):

muzarski:
Rebased on master

Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1089,6 +1089,142 @@ pub unsafe extern "C" fn cass_cluster_set_latency_aware_routing_settings(
...
+        let _ = update_comma_delimited_list(
+            &mut cluster.load_balancing_config.filtering.whitelist_hosts,
+            hosts,
+            hosts_size,
+            |s| match IpAddr::from_str(s) {
+                Ok(ip) => Some(ip),
+                Err(err) => {
+                    tracing::warn!("Failed to parse ip address <{}>: {}", s, err);
+                    None
+                }
```
wprzytula: :wrench: I'd make this `tracing::error!`.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1754,6 +1754,132 @@ mod tests {
...
+                        .filtering
+                        .blacklist_hosts
+                        .is_empty()
+                );
+            }
+
+            cass_cluster_free(cluster_raw);
+        }
+    }
+
```
wprzytula: :wrench: Missing analogous unit test for setting filtering on execution profile.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -13,6 +16,73 @@ pub(crate) struct FilteringConfig {
...
+    pub(crate) whitelist_dc: Option<Vec<String>>,
+    pub(crate) blacklist_dc: Option<Vec<String>>,
+}
+
+impl FilteringInfo {
+    /// Checks if the host is valid according to the filtering rules.
+    ///
+    /// If host does not belong to any datacenter, it's datacenter is treated
+    /// as empty string. This way, if for example only `dc1` is whitelisted, the
+    /// node with unknown DC will be rejected.
```
wprzytula: Typo: `it's` -> `its`.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -13,6 +16,73 @@ pub(crate) struct FilteringConfig {
...
+    pub(crate) fn is_host_valid(&self, ip: &IpAddr, dc: &Option<String>) -> bool {
```
wprzytula: :pick: IMO it would be cleaner (being less restrictive) to accept `dc: Option<&str>`.

wprzytula: `Option<&str>` is easily obtained from `&Option<String>` using `.as_deref()`.

wprzytula: :pick: `is_host_allowed` suits better IMO.

scylla-rust-wrapper/src/load_balancing.rs
wprzytula: :question:
1. Can an empty datacenter be specified in the whitelist or in the blacklist?
2. If not, is this semantics really correct? Consider a node that provides no DC info. Then we cannot blacklist it, so the only viable solution to disallow it is to whitelist all other DCs.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -90,3 +165,60 @@ pub(crate) enum LoadBalancingKind {
...
+    ) -> FallbackPlan<'a> {
+        Box::new(
+            self.child_policy
+                .fallback(request, cluster)
+                .filter(|(node, _shard)| {
+                    self.filtering
+                        .is_host_valid(&node.address.ip(), &node.datacenter)
+                }),
+        )
+    }
```
wprzytula: :thought_balloon: Hmm, this way we're back to LBP composition, which we purposefully ditched in the [LBP refactor](https://github.com/scylladb/scylla-rust-driver/pull/612) due to complexity in reasoning about it. Even now I'm unsure if this is correct (but it seems it is - we just filter out disallowed targets).

Efficiency is another concern of mine. `DefaultPolicy` and `ReplicaLocator` were both designed with DC-awareness in mind, but only wrt the single preferred DC. Black- and whitelisting DCs must be done in a layer above, which, among other disadvantages, will likely often make `pick()` return `None` and thus force `fallback()` to execute, requiring allocations also on the "happy" path (i.e., when the first contacted target responds successfully).

While I don't have an immediate solution to this issue, I just wanted to point out that black- and whitelisting nodes likely comes with a performance penalty on the LBP level.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -237,3 +237,120 @@ impl HostFilter for CassHostFilter {
             .is_host_valid(&peer.address.ip(), &peer.datacenter)
     }
 }
+
+/// Returns the union of all non-empty vectors (sets).
+/// If at least one set is empty, it return None.
+fn nonempty_union<'a, T>(iter: impl Iterator<Item = &'a Vec<T>>) -> Option<Vec<T>>
+where
+    T: Clone + Ord + 'a,
+{
```
wprzytula: :question: Why do we require `Ord`? Isn't `Eq` enough?

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -282,6 +282,57 @@ where
     intersection.filter(|v| !v.is_empty())
 }
 
+impl CassHostFilter {
+    /// Computes the filtering rules from the load balancing policies, and
+    /// returns a corresponding `HostFilter`.
+    ///
+    /// In cpp-driver, the following rule is upheld: if a host is rejected
+    /// by **all** policies, the connection to that host is not opened at all.
```
wprzytula, line 290: :wrench: Please provide a link to the relevant piece of code in cpp-driver.

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -282,6 +282,57 @@ where
...
+            configs
+                .clone()
+                .map(|lbp_config| &lbp_config.filtering.whitelist_dc),
+        );
+
+        let blacklist_dc = nonempty_intersection(
+            configs
+                .clone()
+                .map(|lbp_config| &lbp_config.filtering.blacklist_dc),
+        );
```
wprzytula: :pick: Here we don't have to `clone()` - it's the last use of `configs`.

scylla-rust-wrapper/src/session.rs
```diff
@@ -124,6 +128,7 @@ impl CassSessionInner {
         }
 
         let session = session_builder
+            .host_filter(host_filter)
             .build()
```
wprzytula, line 132: :thought_balloon: Maybe, as an optimisation, only provide a `HostFilter` to the `Session` if there is any nonempty whitelist/blacklist?

wprzytula: I suppose minority of users will be using the whitelist/blacklist functionality, and even fewer of them will disallow some hosts in all execution profiles and in the cluster.

wprzytula (changes requested):

scylla-rust-wrapper/src/load_balancing.rs
muzarski: When I first implemented it, I removed the duplicates by sorting and using `dedup`. Good observation - `Ord` is no longer required.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: > 💭 Maybe, as an optimisation, only provide a HostFilter to the Session if there is any nonempty whitelist/blacklist?

Why? HostFilter is only called in places where such micro-optimizations won't matter.

scylla-rust-wrapper/src/load_balancing.rs
muzarski: > ❓
> 
>     1. Can an empty datacenter be specified in the whitelist or in the blacklist?

No - `explode` function in cpp-driver filters out empty strings.

>     2. If not, is this semantics really correct? Consider a node that provides no DC info. Then we cannot blacklist it, so the only viable solution to disallow it is to whitelist all other DCs.

You are correct - we cannot blacklist "no info" dc. But this is the semantics from cpp-driver - I just sticked to it. We could always change it if necessary.

scylla-rust-wrapper/src/cluster.rs
muzarski: True, I was lazy. I decided that testing cluster methods is enough, since both cluster and exec_profile methods are just wrappers over `update_comma_delimited_list`, which contains most of the tested logic.

Should I add these?

scylla-rust-wrapper/src/cluster.rs
muzarski: Why? This does not fail the operation - it just filters out the bad entry in a comma-delimited list.

muzarski: The remaining "good" entries are appended anyway.

scylla-rust-wrapper/src/load_balancing.rs
muzarski: Since we use `&str` in case `None` is provided (thus, we can't use `contains()` Vec method anyway, because it requires `&String`), I agree.

scylla-rust-wrapper/src/session.rs
wprzytula: Because why not? Why even turn on a filter that passes through everything unconditionally?

scylla-rust-wrapper/src/load_balancing.rs
wprzytula: Good catch about `Vec::contains()` - I didn't consider that.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: > Because why not?

Because it requires additional logic and complicates invariants.

scylla-rust-wrapper/src/load_balancing.rs
muzarski: > Black- and whitelisting DCs must be done in a layer above

Correct - unless we introduce the support for it in rust-driver (some configuration on `DefaultPolicy`) - this is not that important IMO.

> will likely often make `pick()` return `None` and thus force `fallback()` to execute

Do you think this will happen often? Why would someone prefer some dc, and then filter out the node from preferred dc.

But maybe you meant the general scenario, where underlying LBP is non-dc-aware. In such case - that's true, `pick` may often return unwanted node.

But I think the general use case for filtering is using a dc-aware LBP, and (on top of it) filter out some nodes from remote dcs which we don't want to fallback to.

wprzytula: > > will likely often make `pick()` return `None` and thus force `fallback()` to execute
> 
> Do you think this will happen often? Why would someone prefer some dc, and then filter out the node from preferred dc.
> 
> But maybe you meant the general scenario, where underlying LBP is non-dc-aware. In such case - that's true, `pick` may often return unwanted node.

Yes, I meant the non-DC-aware underlying LBP.
> 
> But I think the general use case for filtering is using a dc-aware LBP, and (on top of it) filter out some nodes from remote dcs which we don't want to fallback to.

Such use case makes sense. However, another one seems to me just as likely as this one: there are 2 local DCs and 3 remote DCs. We want to filter out the remote DCs. OTOH, we don't want to prefer any local DC over the other local DC. Then we are left with this problem, because `DefaultPolicy` will often `pick()` targets from remote DCs, so `fallback()` will be inevitable.

wprzytula:
:question: You haven't marked:
```
I have enabled appropriate tests in .github/workflows/build.yml in gtest_filter.
I have enabled appropriate tests in .github/workflows/cassandra.yml in gtest_filter.
```
Did you just overlook them?

muzarski:
> ❓ You haven't marked:
> 
> ```
> I have enabled appropriate tests in .github/workflows/build.yml in gtest_filter.
> I have enabled appropriate tests in .github/workflows/cassandra.yml in gtest_filter.
> ```
> 
> Did you just overlook them?

The `ExecutionProfile` tests that are testing filtering methods require support for exposing/enforcing coordinator. I'm waiting for these to be implemented. Then, we can enable the corresponding tests and merge this PR (I don't want to merge it without tests enabled).

scylla-rust-wrapper/src/session.rs
muzarski: I'm resolving this - not planning to add this micro-optimization

muzarski:
v2.2: Addressed @wprzytula comments (some of them still need an answer)

scylla-rust-wrapper/src/cluster.rs
wprzytula: Let's add them. They should mostly involve copying and adjusting existing tests for cluster methods.

scylla-rust-wrapper/src/cluster.rs
wprzytula: `error!` does not necessarily mean a complete failure. It may signify that just part of the operation failed. IMO, passing an unparsable IP address counts as an error.

muzarski: Ok

muzarski:
v2.3: Next iteration of addressing @wprzytula comments. As a bonus (after discussion on Slack), I added casts to `fn(&str) -> T` for closures with empty env.

wprzytula (approved):

muzarski:
Thanks for the approves - I'm not merging yet. I'll wait for `cass_future_coordinator` support to enable the corresponding integration tests.

Lorak-mmk:
> As a bonus (after discussion on Slack), I added casts to fn(&str) -> T for closures with empty env.

Why? Did you check that it really improves something? In other words, are multiple copies of the function generated (in release mode) if we remove those casts, as compared to the current version?
We never did such casts for a gazzilion methods that accept a closure (e.g. in every iterator usage).

muzarski:
> > As a bonus (after discussion on Slack), I added casts to fn(&str) -> T for closures with empty env.
> 
> Why? Did you check that it really improves something? In other words, are multiple copies of the function generated (in release mode) if we remove those casts, as compared to the current version? We never did such casts for a gazzilion methods that accept a closure (e.g. in every iterator usage).

Without casts:
```
nm -an target/release/libscylla_cpp_driver.so | c++filt | grep comma_delimi

0000000000141730 t scylla_cpp_driver::cluster::update_comma_delimited_list::h625b382240391297
0000000000141970 t scylla_cpp_driver::cluster::update_comma_delimited_list::ha5386510b8545111
0000000000141b70 t scylla_cpp_driver::cluster::update_comma_delimited_list::hb5560b92aa22d880
0000000000141d70 t scylla_cpp_driver::cluster::update_comma_delimited_list::hea0076943954fdf8
0000000000466eb8 d scylla_cpp_driver::cluster::update_comma_delimited_list::__CALLSITE::META::hf3d04d6f74274d08
00000000004ce2e8 d scylla_cpp_driver::cluster::update_comma_delimited_list::__CALLSITE::h1a7572fa57806091
```

With casts:
```
nm -an target/release/libscylla_cpp_driver.so | c++filt | grep comma_delimi

00000000001427b0 t scylla_cpp_driver::cluster::update_comma_delimited_list::ha48524fffb4b98c2
00000000001429f0 t scylla_cpp_driver::cluster::update_comma_delimited_list::hf1fb12ec9c9e5165
0000000000465fe8 d scylla_cpp_driver::cluster::update_comma_delimited_list::__CALLSITE::META::hf3d04d6f74274d08
00000000004cd2e8 d scylla_cpp_driver::cluster::update_comma_delimited_list::__CALLSITE::h1a7572fa57806091
```

Lorak-mmk:
So probably it is a tradeoff of inlining vs less copies, right? In other words, with casts we get 2 less copies, but the functions won't be inlined. Do we have reasons to believe this is the right tradeoff to do here?

muzarski:
> Do we have reasons to believe this is the right tradeoff to do here?

I don't have any. Maybe @wprzytula has some.

wprzytula:
> So probably it is a tradeoff of inlining vs less copies, right? In other words, with casts we get 2 less copies, but the functions won't be inlined. Do we have reasons to believe this is the right tradeoff to do here?

By telling the compiler that this is the same function, we leave the decision up to it whether to inline or not.
By not saying it, the compiler may still inline (and then the code ends up equivalent) or not. If not, this is the worst case: there are multiple copies of the same function, each called from another place in code. With our hint, all calls can target the same function.

Tl;dr
@muzarski's experiment proves that the compiler can do better if hinted in a way @muzarski did.

muzarski:
> > So probably it is a tradeoff of inlining vs less copies, right? In other words, with casts we get 2 less copies, but the functions won't be inlined. Do we have reasons to believe this is the right tradeoff to do here?
> 
> By telling the compiler that this is the same function, we leave the decision up to it whether to inline or not. By not saying it, the compiler may still inline (and then the code ends up equivalent) or not. If not, this is the worst case: there are multiple copies of the same function, each called from another place in code. With our hint, all calls can target the same function.

I misunderstood @Lorak-mmk at first as well - he meant inlining of the closure provided to `update_comma_delimited_list`, and not the `update_comma_delimited_list` itself.

muzarski:
v2.4: Removed the casts for closures (after a thorough discussion on Slack)

Lorak-mmk (approved):

wprzytula (approved):

muzarski:
v3: Rebased on master (with getting coordinator support) and enabled 4 integration tests from `[DC]ExecutionProfileTest` suites. (See only last 3 commits)

Lorak-mmk (approved):

wprzytula (approved):

PR #292, opened 2025-05-05
integration: use NTS by default

muzarski:
For the tests with single DC clusters, we would always use `SimpleStrategy`. This PR addresses this, and changes it so we use `NetworkTopologyStrategy` for Scylla clusters. For Cassandra clusters, we will still use `SimpleStrategy` - it's because Cassandra `3.11.19` (the version we currently test against) does not allow `replication_factor` option with NTS.

For `SimpleStrategy` keyspaces, Scylla disables the tablets by default. When we transition to NTS, some tests start to fail - it is expected - some features are currently not working with tablets. In this PR, we disable the tablets for tests which:
1. use counters
2. use LWTs
3. use materialized views
4. try to decomission a node (`ccm` decomission seems to fail for nodes with tablets keyspaces - the exact error is included in corresponding commit's message).

## Pre-review checklist

### DIFF
--- a/tests/src/integration/integration.cpp
+++ b/tests/src/integration/integration.cpp
@@ -268,7 +268,7 @@ std::string Integration::default_replication_strategy() {
     replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': " << number_dc1_nodes_ << ", "
                            << "'dc2': " << number_dc2_nodes_;
   } else {
-    replication_strategy_s << "'SimpleStrategy', 'replication_factor': ";
+    replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': ";
 
     // Ensure the replication factor has not been overridden or already set
     if (replication_factor_ == 0) {
--- a/tests/src/integration/integration.cpp
+++ b/tests/src/integration/integration.cpp
@@ -66,6 +66,7 @@ Integration::Integration()
     , is_keyspace_change_requested_(true)
     , is_test_chaotic_(false)
     , is_beta_protocol_(Options::is_beta_protocol())
+    , disable_tablets_(false)
     , protocol_version_(CASS_PROTOCOL_VERSION_V4)
     , create_keyspace_query_("")
     , start_time_(0ull) {
@@ -349,6 +350,13 @@ void Integration::connect(Cluster cluster) {
              << server_version_.to_string());
   }
 
+  // Check if scylla supports TABLETS feature. If so, and test
+  // does not work with tablets (e.g. it uses LWT), disable the tablets
+  // for test keyspace.
+  if (disable_tablets_ && scylla_supports_feature("TABLETS")) {
+    create_keyspace_query_ = create_keyspace_query_ + " AND TABLETS = { 'enabled': false }";
+  }
+
   // Create the keyspace for the integration test
   session_.execute(create_keyspace_query_);
   CHECK_FAILURE;
@@ -472,6 +480,19 @@ std::string Integration::generate_contact_points(const std::string& ip_prefix,
     contact_points.push_back(contact_point.str());
   }
   return implode(contact_points, ',');
+}
+
+bool Integration::scylla_supports_feature(const std::string& feature) {
+  if (!Options::is_scylla()) {
+    return false;
+  }
+
+  Result res = session_.execute("SELECT supported_features FROM system.local WHERE key='local'");
+  Text supported_features = res.first_row().column_by_name<Text>("supported_features");
+  if (supported_features.is_null()) {
+    return false;
+  }
+  return supported_features.value().find(feature) != std::string::npos;
 }
 
 std::string Integration::format_string(const char* format, ...) const {
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :BatchSingleNodeClusterTests*:BatchCounterSingleNodeClusterTests*:BatchCounterThreeNodeClusterTests*\
 :ErrorTests.*\
 :SslNoClusterTests*:SslNoSslOnClusterTests*\
-:SchemaMetadataTest.*KeyspaceMetadata:SchemaMetadataTest.*MetadataIterator:SchemaMetadataTest.*View*\
+:SchemaMetadataTest.*\
 :TracingTests.*\
 :ByNameTests.*\
 :CompressionTests.*\
@@ -37,6 +37,8 @@ SCYLLA_TEST_FILTER := $(subst ${SPACE},${EMPTY},ClusterTests.*\
 :ExecutionProfileTest.Integration_Cassandra_SerialConsistency\
 :ExecutionProfileTest.Integration_Cassandra_LatencyAwareRouting\
 :-PreparedTests.Integration_Cassandra_PreparedIDUnchangedDuringReprepare\
+:SchemaMetadataTest.Integration_Cassandra_RegularMetadataNotMarkedVirtual\
+:SchemaMetadataTest.Integration_Cassandra_VirtualMetadata\
 :HeartbeatTests.Integration_Cassandra_HeartbeatFailed\
 :ControlConnectionTests.Integration_Cassandra_TopologyChange\
 :ControlConnectionTests.Integration_Cassandra_FullOutage\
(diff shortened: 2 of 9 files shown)


Lorak-mmk:
Regarding "it: disable tablets for ConsistencyThreeNodeClusterTests ":
What is the replication factor in tests, and how many nodes are there?
Imo the decomission error message suggests that there would be too few nodes left to satisfy replication factor - in which case we need to change the test, or the cluster used in tests.

tests/src/integration/integration.cpp
```diff
@@ -268,7 +268,9 @@ std::string Integration::default_replication_strategy() {
     replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': " << number_dc1_nodes_ << ", "
                            << "'dc2': " << number_dc2_nodes_;
   } else {
-    replication_strategy_s << "'SimpleStrategy', 'replication_factor': ";
+    // Currently we use Cassandra 3.11.19, which disallows the use of `replication_factor` with NTS.
+    std::string strategy = Options::is_scylla() ? "'NetworkTopologyStrategy'" : "'SimpleStrategy'";
+    replication_strategy_s << strategy << ", 'replication_factor': ";
```
Lorak-mmk: What is the error message when trying to use `replication_factor` with Cassandra?

muzarski: "Configuration error: Database returned an error: The query is invalid because of some configuration issue, Error message: replication_factor is an option for SimpleStrategy, not NetworkTopologyStrategy"

muzarski:
> Imo the decomission error message suggests that there would be too few nodes left to satisfy replication factor - in which case we need to change the test, or the cluster used in tests.

Assuming this is true (I haven't checked the values you asked for yet), why we can't satisfy repfactor for tablets, but we could do that for vnodes?

tests/src/integration/integration.cpp
Lorak-mmk: Afaik using `'replication_factor': n` is the same as specifying this same RF for each DC.
I think in this branch there is only one DC, right? In which case `'dc1': n` should work.

wprzytula (approved):

tests/src/integration/integration.cpp
muzarski: Correct, but I'm not sure how `ccm` works in case of creating a cluster with single dc - i.e. I don't know what's the default name for this single dc (if any). I'll need to check.

What I know for sure is that if there are multiple dcs, ccm calls them `dc1`, `dc2`, ..., `dc_n`.

Lorak-mmk:
> Assuming this is true (I haven't checked the values you asked for yet), why we can't satisfy repfactor for tablets, but we could do that for vnodes?

You couldn't, but the DB did not verify that. For example, it was possible to create a keyspace with RF=4 on a 3-node cluster (but there would not be 4 copies of data).
With tablets such verifications are now in place, so the previously wrong behaviors are now rejected explicitly.

muzarski:
> You couldn't, but the DB did not verify that. For example, it was possible to create a keyspace with RF=4 on a 3-node cluster (but there would not be 4 copies of data). With tablets such verifications are now in place, so the previously wrong behaviors are now rejected explicitly.

Makes sense, thanks. I'll investigate the `ConsistencyThreeNodeCluster` test suite further then.

muzarski:
Regarding `ConsistencyThreeNodeTests`: As the name suggests, it starts with 3 nodes cluster.

It looks like this suite explicitly sets the RF to 3. We want to keep RF=3 on the keyspace, decommission some node(s), and expect the errors on higher consistency levels (such as `ALL` or `THREE`).

Since Scylla disallows us to decommission the node (which makes sense), we can simply rewrite the test to use `ccm stop` instead of `ccm decommission` - this would simulate a scenario when node crashes (and it does not cleanly migrate its data to other nodes, as opposed to `ccm decommission`).

Example test case (`OneNodeDecommissioned`):
```c
  // Perform a sanity check against a full healthy cluster (N=3, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_ALL);
  select_.set_consistency(CASS_CONSISTENCY_ALL);
  session_.execute(insert_);
  session_.execute(select_);

  // Decommission node two
  decommission_node(2);

  // Perform a check using consistency `QUORUM` (N=2, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_QUORUM);
  select_.set_consistency(CASS_CONSISTENCY_QUORUM);
  session_.execute(insert_);
  session_.execute(select_);

  // Perform a check using consistency `ONE` (N=2, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_ONE);
  select_.set_consistency(CASS_CONSISTENCY_ONE);
  session_.execute(insert_);
  session_.execute(select_);

  // Perform a check using consistency `TWO` (N=2, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_TWO);
  select_.set_consistency(CASS_CONSISTENCY_TWO);
  session_.execute(insert_);
  session_.execute(select_);

  // Perform a check using consistency `ALL` (should fail N=2, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_ALL);
  select_.set_consistency(CASS_CONSISTENCY_ALL);
  ASSERT_NE(CASS_OK, session_.execute(insert_, false).error_code());
  ASSERT_NE(CASS_OK, session_.execute(select_, false).error_code());

  // Perform a check using consistency `THREE` (should fail N=2, RF=3)
  insert_.set_consistency(CASS_CONSISTENCY_THREE);
  select_.set_consistency(CASS_CONSISTENCY_THREE);
  ASSERT_NE(CASS_OK, session_.execute(insert_, false).error_code());
  ASSERT_NE(CASS_OK, session_.execute(select_, false).error_code());
```

We can simply replace `decommission_node()` call with `stop_node` - the test then passes for me locally. WDYT? @Lorak-mmk @wprzytula

Lorak-mmk:
Makes sense

muzarski:
Rebased on master

muzarski:
v1.1:
- using `ccm stop` instead of `ccm decommission` in `ConsistencyThreeNodeClusterTests`
- using NTS for cassandra as well. The default dc name for single-dc cluster is `dc1`, so we can just do `'class': 'NetworkTopologyStrategy', 'dc1': <rf>`

wprzytula (approved):

tests/src/integration/integration.cpp
```diff
@@ -268,7 +268,9 @@ std::string Integration::default_replication_strategy() {
     replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': " << number_dc1_nodes_ << ", "
                            << "'dc2': " << number_dc2_nodes_;
   } else {
-    replication_strategy_s << "'SimpleStrategy', 'replication_factor': ";
+    // Cassandra 3.11.19 (which we currently use) disallows the usage of `replication_factor` with NTS.
+    std::string rf_target = Options::is_scylla() ? "'replication_factor'" : "'dc1'";
+    replication_strategy_s << "'NetworkTopologyStrategy', " << rf_target << ": ";
```
Lorak-mmk: Is there some downside to always using "'dc1'"?

Lorak-mmk (approved):

tests/src/integration/integration.cpp
muzarski: I don't think so. I can change it to being always `'dc1'`.

Lorak-mmk: Let's do that

wprzytula:
Hmm...
> [  FAILED  ] ExecutionProfileTest.Integration_Cassandra_RequestTimeout

muzarski:
> Hmm...
> 
> > [  FAILED  ] ExecutionProfileTest.Integration_Cassandra_RequestTimeout

```
      Expected: CASS_ERROR_LIB_REQUEST_TIMED_OUT
      Which is: 16777230
To be equal to: result.error_code()
      Which is: 33563136
```

Value `335...` corresponds to `CASS_ERROR_SERVER_INVALID_QUERY`. I wonder what exact error server returned..

muzarski:
> Hmm...
> 
> > [  FAILED  ] ExecutionProfileTest.Integration_Cassandra_RequestTimeout

I've run this single test case 100 times in a loop locally against Scylla 5.4.8 - the error did not reproduce.

Lorak-mmk (approved):

wprzytula (approved):

PR #294, opened 2025-05-05
Enable some of the `ServerSideFailure` tests

muzarski:
Ref: #132 

cpp-driver does not expose API to inspect the warnings emitted by Scylla. Instead, it logs the warnings at `WARN` level.

In this PR, we also enable non-UDF `ServerSideFailure` tests:
- `ErrorTableAlreadyExists` -> this works without any adjustments
- `Warning` -> this works only for Cassandra. It's because Scylla does not emit similar warning. I decided to enable the test for Scylla as well, but skip it during runtime (with proper message, informing why the test is skipped).

Remaining two tests from this suite are using UDFs. For now, I'm leaving them disabled for both Scylla and Cassandra. In the future, we might want to enable UDF Scylla feature and run those tests.

## Pre-review checklist


scylla-rust-wrapper/src/session.rs
```diff
@@ -220,11 +220,17 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
+                for warning in result.warnings() {
+                    tracing::warn!("Server-side warning: {}", warning);
+                }
+
+                Ok(CassResultValue::QueryResult(Arc::new(CassResult {
+                    tracing_id: None,
+                    paging_state_response: PagingStateResponse::NoMorePages,
+                    kind: CassResultKind::NonRows,
+                })))
+            }
```
Lorak-mmk: I'm confused. This issue: https://github.com/scylladb/scylla-rust-driver/issues/265 says that Rust Driver already prints warnings. Why do you need to print them again?

muzarski: Maybe I did not notice that because I was running the test locally only against Scylla (which does not emit this specific warning). I'll check once again against Cassandra.

muzarski: Yeah..
```
WARN: message: Response from the database contains a warning, warning: "Aggregation query used without partition key"
```

I'll remove the first commit and adjust the test case.

Lorak-mmk (approved):

PR #252, opened 2025-04-15
Configure coalescing delay

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/238

This PR implements  `cass_cluster_set_coalesce_delay`.

## Pre-review checklist

- [x] I have split my patch into logically separate commits.
- [x] All commit messages clearly explain what they change and why.
- [x] PR description sums up the changes and reasons why they should be introduced.
- [x] I have implemented Rust unit tests for the features/changes introduced.
- ~[ ] I have enabled appropriate tests in `.github/workflows/build.yml` in `gtest_filter`.~
- ~[ ] I have enabled appropriate tests in `.github/workflows/cassandra.yml` in `gtest_filter`.~

PR #295, opened 2025-05-06
Implement timestamp generators and enable 4 out of 5 TimestampTests

### LINKED ISSUE #46: Implement timestamp generator configuration
Function list:

- [ ] `cass_timestamp_gen_server_side_new`
- [ ] `cass_timestamp_gen_monotonic_new`
- [ ] `cass_timestamp_gen_monotonic_new_with_settings`
- [ ] `cass_timestamp_gen_free`
- [ ] `cass_cluster_set_timestamp_gen`

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/46
Ref: #132 

This PR implements timestamp methods.

## Enabled integration `TimestampTests`
-  `Statement`
- `BatchStatement`
- `ServerSideTimestampGeneratorStatement`
- `ServerSideTimestampGeneratorBatchStatement`

## Disabled integration test
There is one test that we need to keep disabled (for now). I want to create a separate issue for this. issue: https://github.com/scylladb/cpp-rust-driver/issues/296.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -6,6 +6,7 @@ use crate::future::CassFuture;
 use crate::retry_policy::CassRetryPolicy;
 use crate::retry_policy::RetryPolicy::*;
 use crate::ssl::CassSsl;
+use crate::timestamp_generator::CassTimestampGen;
 use crate::types::*;
 use crate::uuid::CassUuid;
 use openssl::ssl::SslContextBuilder;
@@ -19,6 +20,7 @@ use scylla::policies::load_balancing::{
 };
 use scylla::policies::retry::RetryPolicy;
 use scylla::policies::speculative_execution::SimpleSpeculativeExecutionPolicy;
+use scylla::policies::timestamp_generator::TimestampGenerator;
 use scylla::routing::ShardAwarePortRange;
 use scylla::statement::{Consistency, SerialConsistency};
 use std::collections::HashMap;
@@ -464,6 +466,33 @@ pub unsafe extern "C" fn cass_cluster_set_tcp_keepalive(
     let tcp_keepalive_interval = enabled.then(|| Duration::from_secs(delay_secs as u64));
 
     cluster.session_builder.config.tcp_keepalive_interval = tcp_keepalive_interval;
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_cluster_set_timestamp_gen(
+    cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
+    timestamp_gen_raw: CassBorrowedSharedPtr<CassTimestampGen, CMut>,
+) {
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
+        tracing::error!("Provided null cluster pointer to cass_cluster_set_timestamp_gen!");
+        return;
+    };
+    let Some(timestamp_gen) = BoxFFI::as_ref(timestamp_gen_raw) else {
+        tracing::error!(
+            "Provided null timestamp generator pointer to cass_cluster_set_timestamp_gen!"
+        );
+        return;
+    };
+
+    let rust_timestamp_gen: Option<Arc<dyn TimestampGenerator>> = match timestamp_gen {
+        // In rust-driver, `None` is equivalent to using server-side timestamp generator.
+        CassTimestampGen::ServerSide => None,
+        CassTimestampGen::Monotonic(monotonic_timestamp_generator) => {
+            Some(Arc::clone(monotonic_timestamp_generator) as _)
+        }
+    };
+
+    cluster.session_builder.config.timestamp_generator = rust_timestamp_gen;
 }
 
 #[unsafe(no_mangle)]
--- a/scylla-rust-wrapper/src/timestamp_generator.rs
+++ b/scylla-rust-wrapper/src/timestamp_generator.rs
@@ -1,0 +1,60 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use scylla::policies::timestamp_generator::MonotonicTimestampGenerator;
+
+use crate::argconv::{BoxFFI, CMut, CassOwnedExclusivePtr, FFI, FromBox};
+
+pub enum CassTimestampGen {
+    ServerSide,
+    Monotonic(Arc<MonotonicTimestampGenerator>),
+}
+
+impl FFI for CassTimestampGen {
+    type Origin = FromBox;
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_timestamp_gen_server_side_new()
+-> CassOwnedExclusivePtr<CassTimestampGen, CMut> {
+    BoxFFI::into_ptr(Box::new(CassTimestampGen::ServerSide))
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_timestamp_gen_monotonic_new()
+-> CassOwnedExclusivePtr<CassTimestampGen, CMut> {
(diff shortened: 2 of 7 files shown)


scylla-rust-wrapper/src/timestamp_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+            let warning_interval =
+                Duration::from_millis(warning_interval_ms.try_into().unwrap_or(1_u64));
+
+            MonotonicTimestampGenerator::new()
+                .with_warning_times(warning_threshold, warning_interval)
+        }
+    };
+
+    BoxFFI::into_ptr(Box::new(CassTimestampGen::Monotonic(Arc::new(generator))))
+}
```
Lorak-mmk: I'm curious, why did you use `.cmp` instead of operator like `<=`?
I'm not used to seeing or writing this method, so it was not obvious to me what is its ordering (does `Less` mean that 0 is less or warning_interval_ms is less?). What is the benefit of using it?

scylla-rust-wrapper/src/timestamp_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+    warning_interval_ms: i64,
+) -> CassOwnedExclusivePtr<CassTimestampGen, CMut> {
+    let generator = match warning_interval_ms.cmp(&0) {
+        // If threshold is <= 0, we disable the warnings.
+        Ordering::Less | Ordering::Equal => MonotonicTimestampGenerator::new().without_warnings(),
+        Ordering::Greater => {
+            let warning_threshold = Duration::from_micros(warning_threshold_us as u64);
+            // Inverval <= 0 fallbacks to 1ms.
+            let warning_interval =
+                Duration::from_millis(warning_interval_ms.try_into().unwrap_or(1_u64));
```
Lorak-mmk: Also the comments look mixed up. When comparing interval, the comment talks about threshold. When operating on threshold it mentions interval.

src/timestamp_generator.cpp
```diff
@@ -22,32 +22,6 @@
...
-
-CassTimestampGen* cass_timestamp_gen_monotonic_new_with_settings(int64_t warning_threshold_us,
-                                                                 int64_t warning_interval_ms) {
-  TimestampGenerator* timestamp_gen =
-      new MonotonicTimestampGenerator(warning_threshold_us, warning_interval_ms);
-  timestamp_gen->inc_ref();
-  return CassTimestampGen::to(timestamp_gen);
-}
-
-void cass_timestamp_gen_free(CassTimestampGen* timestamp_gen) { timestamp_gen->dec_ref(); }
```
Lorak-mmk, line 47: Commit: cpp: remove timestamp methods from src/ 

Did you check if its possible to remove more code? For example this whole file?
IIRC, the src folder was a necessary hack for testing. If we can get rid of some (all?) of it, we definitely should.

Lorak-mmk (approved):

Lorak-mmk (changes requested):
Ooops, that was not supposed to be an approve

src/timestamp_generator.cpp
muzarski: I tried - and we cannot. The remaining code is used by one test case (see the issue #296 )

scylla-rust-wrapper/src/timestamp_generator.rs
muzarski: Oops. This is a bug - the comments are correct - if threshold is <= 0, we should disable the warnings. I wanted to write unit tests to spot stupid bugs like this one, but we don't have any public getters on `MonotonicTimestampGenerator` in rust-driver.

scylla-rust-wrapper/src/timestamp_generator.rs
muzarski: I started using this after clippy once shouted at me for using `if x > 0 {} else if x == 0 {} else {}` instead of `cmp()` method. See the reference: https://github.com/scylladb/cpp-rust-driver/pull/252#issuecomment-2808840986. I'm not sure if clippy would complain in this case as well (since `< 0` and `== 0` branches are the same). I could try to rewrite this if you suggest doing so.

`Less` in this case means that `warning_interval_ms  < 0` - see https://doc.rust-lang.org/std/cmp/trait.Ord.html#tymethod.cmp

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -466,6 +468,33 @@ pub unsafe extern "C" fn cass_cluster_set_tcp_keepalive(
...
+    let rust_timestamp_gen: Option<Arc<dyn TimestampGenerator>> = match timestamp_gen {
+        // In rust-driver, `None` is equivalent to using server-side timestamp generator.
+        CassTimestampGen::ServerSide => None,
+        CassTimestampGen::Monotonic(monotonic_timestamp_generator) => {
+            Some(Arc::clone(monotonic_timestamp_generator) as _)
+        }
+    };
+
+    cluster.session_builder.config.timestamp_generator = rust_timestamp_gen;
+}
```
wprzytula, line 496: :warning: Please confirm that such sharing of `MonotonicTimestampGenerator` possibly between different  sessions / clusters behaves correctly.

wprzytula: I'd prefer to have the `MonotonicTimestampGenerator` cloned into `CassCluster` in this method. The problem is, `MonotonicTimestampGenerator` does not implement `Clone` (nor even `Debug`...). I believe this could/should be added in the Rust driver.

wprzytula (changes requested):

scylla-rust-wrapper/src/timestamp_generator.rs
Lorak-mmk: I don't think clippy will complain about `<=` here, so can we use it?

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: > Please confirm that such sharing of MonotonicTimestampGenerator possibly between different sessions / clusters behaves correctly.

Why would it not? The interface of `TimestampGenerator` is very simple and not session-specific. Unless we have any reasons to believe sharing will cause problems, why spend time trying to un-share the generator?

wprzytula: `MonotonicTimestampGenerator` contains state. If something contains state, it has potential to get corrupted when that state is shared between `Session`s.

Lorak-mmk: >  If something contains state, it has potential to get corrupted when that state is shared between Sessions.

I still don't see why that would be the case. There is nothing session-specific in `TimestampGenerator` trait. It is just an interface for something that generates numbers. It does not care who calls its methods.

wprzytula: OK.

wprzytula: My point about implementing `Debug` for `MonotonicTimestampGenerator` still stands.

Lorak-mmk: Yes, Debug would probably be useful.

muzarski: > My point about implementing `Debug` for `MonotonicTimestampGenerator` still stands.

https://github.com/scylladb/scylla-rust-driver/pull/1341

muzarski:
Rebased on `master` (did not address comments yet)

muzarski:
v1.1: Addressed comments - most importantly fixed a bug where we would compare `interval <= 0` instead of `threshold <= 0`.

wprzytula (approved):

Lorak-mmk (approved):

muzarski:
v1.3: Forgot to cover `warning_interval_ms == 0` case correctly...

Lorak-mmk (approved):

wprzytula (approved):

PR #299, opened 2025-05-12
Implement exposing/enforcing coordinator for request

### LINKED ISSUE #242: Implement enforcing coordinator for the request execution.
Methods:
- `cass_statement_set_host`
- `cass_statement_set_host_n`
- `cass_statement_set_host_inet`
- `cass_statement_set_node`

Waits for https://github.com/scylladb/scylla-rust-driver/issues/1031.

### LINKED ISSUE #249: Implement `cass_future_coordinator`
It waits for https://github.com/scylladb/scylla-rust-driver/issues/1030 on rust-driver side.

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/249, Fixes: https://github.com/scylladb/cpp-rust-driver/issues/242
Ref: #132 

## Implemented methods:
- `cass_statement_set_host`
- `cass_statement_set_host_n`
- `cass_statement_set_host_inet`
- `cass_statement_set_node`
- `cass_future_coordinator`

## Problems with implementing `cass_future_coordinator`
`cass_future_coordinator` returns `CassNode*` which should be borrowed from `CassFuture` and live as long as underlying `CassResult` lives.

The problem is that `CassResult` is stored under a `Mutex`. This is why the obtained reference of `Coordinator` has very short lifetime - it's the lifetime of acquired `MutexGuard`. But we need to return something with longer lifetime - our safe FFI API (with borrow-checker) starts to complain.

In the commit where I implement `cass_future_coordinator` I `unsafe`ly extend the lifetime of the coordinator. We can do that, because we are guaranteed that returned pointer will be valid as long as `CassFutureResult` is valid (see SAFETY comment in the code).

However, there is a way to omit this unsafe code and represent the `CassFutureResult` immutability guarantees (after the future is resolved) on a type level. Last commit changes `CassFuture` a bit, so the `CassFutureResult` is stored inside `OnceLock`, and not inside `Mutex`. Only then, can we remove the unsafe code responsible for extending the lifetime in `cass_future_coordinator`. This commit serves as a suggestion of one of (probably many) solutions to this problem. If we decide to do that, I'll probably split this commit into multiple commits before merging.

## Integration tests
Enabled 3 test suites: `StatementTests`, `StatementNoClusterTests` and `ServerSideFailureThreeNodeClusterTests` - 11 tests in total! The logic of each test is explained in commit messages.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -10,7 +10,7 @@ categories = ["database"]
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.1.0", features = [
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.2.0", features = [
     "openssl-010",
     "metrics",
 ] }
@@ -34,7 +34,7 @@ bindgen = "0.65"
 chrono = "0.4.20"
 
 [dev-dependencies]
-scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.1.0" }
+scylla-proxy = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.2.0" }
 bytes = "1.10.0"
 
 assert_matches = "1.5.0"
--- a/scylla-rust-wrapper/src/integration_testing.rs
+++ b/scylla-rust-wrapper/src/integration_testing.rs
@@ -63,11 +63,6 @@ pub unsafe extern "C" fn testing_cluster_get_contact_points(
 }
 
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn testing_free_contact_points(contact_points: *mut c_char) {
-    let _ = unsafe { CString::from_raw(contact_points) };
-}
-
-#[unsafe(no_mangle)]
 pub unsafe extern "C" fn testing_future_get_host(
     future_raw: CassBorrowedSharedPtr<CassFuture, CConst>,
     host: *mut *mut c_char,
@@ -107,8 +102,8 @@ pub unsafe extern "C" fn testing_future_get_host(
 }
 
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn testing_free_host(host: *mut c_char) {
-    let _ = unsafe { CString::from_raw(host) };
+pub unsafe extern "C" fn testing_free_cstring(s: *mut c_char) {
+    let _ = unsafe { CString::from_raw(s) };
 }
 
 #[derive(Debug)]
--- a/scylla-rust-wrapper/src/integration_testing.rs
+++ b/scylla-rust-wrapper/src/integration_testing.rs
@@ -7,9 +7,12 @@ use scylla::errors::{RequestAttemptError, RequestError};
 use scylla::observability::history::{AttemptId, HistoryListener, RequestId, SpeculativeId};
 use scylla::policies::retry::RetryDecision;
 
-use crate::argconv::{BoxFFI, CMut, CassBorrowedExclusivePtr};
+use crate::argconv::{
+    ArcFFI, BoxFFI, CConst, CMut, CassBorrowedExclusivePtr, CassBorrowedSharedPtr,
+};
 use crate::batch::CassBatch;
 use crate::cluster::CassCluster;
+use crate::future::{CassFuture, CassResultValue};
 use crate::statement::{BoundStatement, CassStatement};
 use crate::types::{cass_int32_t, cass_uint16_t, cass_uint64_t, size_t};
 
@@ -62,6 +65,50 @@ pub unsafe extern "C" fn testing_cluster_get_contact_points(
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn testing_free_contact_points(contact_points: *mut c_char) {
     let _ = unsafe { CString::from_raw(contact_points) };
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn testing_future_get_host(
+    future_raw: CassBorrowedSharedPtr<CassFuture, CConst>,
+    host: *mut *mut c_char,
+    host_length: *mut size_t,
+) {
+    let Some(future) = ArcFFI::as_ref(future_raw) else {
+        tracing::error!("Provided null future pointer to testing_future_get_host!");
+        unsafe {
+            *host = std::ptr::null_mut();
+            *host_length = 0;
+        };
+        return;
+    };
+
+    future.with_waited_result(|r| match r {
+        Ok(CassResultValue::QueryResult(result)) => {
+            // unwrap: Coordinator is none only for unit tests.
+            let coordinator = result.coordinator.as_ref().unwrap();
+
+            let ip_addr_str = coordinator.node().address.ip().to_string();
+            let length = ip_addr_str.len();
+
+            let ip_addr_cstr = CString::new(ip_addr_str).expect(
+                "String obtained from IpAddr::to_string() should not contain any nul bytes!",
+            );
+
+            unsafe {
+                *host = ip_addr_cstr.into_raw();
+                *host_length = length as size_t
+            };
+        }
+        _ => unsafe {
+            *host = std::ptr::null_mut();
+            *host_length = 0;
+        },
+    })
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn testing_free_host(host: *mut c_char) {
+    let _ = unsafe { CString::from_raw(host) };
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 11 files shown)


wprzytula:
@muzarski Please fill the pre-review checklist.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -639,3 +740,139 @@ make_binders!(
...
+    use crate::argconv::BoxFFI;
+    use crate::cass_error::CassError;
+    use crate::inet::CassInet;
+    use crate::statement::{cass_statement_set_host, cass_statement_set_host_inet};
+    use crate::testing::assert_cass_error_eq;
+
+    use super::cass_statement_new;
+
+    #[test]
+    fn test_statement_set_host() {
```
muzarski, line 845: Reminder for myself: `cass_statement_free` at the end of test (currently missing)

scylla-rust-wrapper/src/statement.rs
```diff
@@ -410,6 +414,103 @@ pub unsafe extern "C" fn cass_statement_set_tracing(
...
+    unsafe { cass_statement_set_host_n(statement_raw, host, strlen(host), port) }
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_statement_set_host_n(
+    statement_raw: CassBorrowedExclusivePtr<CassStatement, CMut>,
+    host: *const c_char,
+    host_length: size_t,
+    port: c_int,
+) -> CassError {
```
Lorak-mmk, line 462: One question regarding this just to make sure: does cpp-driver also only allow ip addresses here?
I can imagine it doind something like remembering hostnames for nodes, and accepting a hostname here.

scylla-rust-wrapper/src/integration_testing.rs
```diff
@@ -64,6 +67,50 @@ pub unsafe extern "C" fn testing_free_contact_points(contact_points: *mut c_char
     let _ = unsafe { CString::from_raw(contact_points) };
 }
 
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn testing_future_get_host(
+    future_raw: CassBorrowedSharedPtr<CassFuture, CConst>,
+    host: *mut *mut c_char,
+    host_length: *mut size_t,
+) {
+    let Some(future) = ArcFFI::as_ref(future_raw) else {
```
Lorak-mmk: Side note: In this function we allocate a string, leak it (using `into_raw`) and assign it to user-provided pointer.
The user is now responsible for freeing this string, but since this is a normal string there is no cpp-driver API for this (think `cass_something_free`), so C-provided `free` should be used.

This is not the first appearance of this pattern.

I have two concerns:
- Is it even legal to free Rust-allocated stuff using `free` from C? It very well might be, but I'm, not sure.
- I think there won't be any issues caused by that pattern when developing custom allocator API from cpp-driver, but again I'm not sure. We'll need to be careful.

Lorak-mmk: Oh wait, this is just testing API, I didn't notice. I'm sure this pattern appears in some public APIs, so the comment should still be valid.

Lorak-mmk (approved):
I like the change in the last commit.

scylla-rust-wrapper/src/integration_testing.rs
muzarski: > Side note: In this function we allocate a string, leak it (using `into_raw`) and assign it to user-provided pointer. The user is now responsible for freeing this string, but since this is a normal string there is no cpp-driver API for this (think `cass_something_free`), so C-provided `free` should be used.

I introduced `testing_free_host` which should be called on a pointer obtained from `testing_future_get_host`. User does not need to call C `free`. Or am I understanding your concerns wrong?

Note: In the next commit `testing_free_host` and `testing_free_contact_points` are replaced with `testing_free_cstring` which can be used instead of them. These two functions did the same, so I introduced a common function (with a more suitable name).

scylla-rust-wrapper/src/statement.rs
muzarski: Does this test case answer your question: https://github.com/scylladb/cpp-rust-driver/blob/master/tests/src/integration/tests/test_statement.cpp#L162?

From what I understood, "notvalid" could technically be a hostname which driver remembers. BTW, notice that this test works without any cluster (technically a unit test). OTOH, this API is called on `CassStatement`, which has no access to the driver's metadata such as reachable hosts. The API is supposed to work "offline" (as in the test).

For reference (cpp-driver):
- implementation of `cass_statement_set_host_n`: https://github.com/scylladb/cpp-driver/blob/master/src/statement.cpp#L155-L163 - don't let "is_valid_and_**resolved**" mislead you - no DNS hostname resolution takes place here. It's just translation of numeric address to bytes.
- corresponding `Address` constructor: https://github.com/scylladb/cpp-driver/blob/master/src/address.cpp#L50-L64 (`server_name` argument default-initialized as empty string)

scylla-rust-wrapper/src/integration_testing.rs
Lorak-mmk: > Or am I understanding your concerns wrong?

In testing your `testing_free_host`, but we have do places in the public API where the string is filled by us, and freed by the user with `free`, right?

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: cpp driver source you linked does answer my question, thanks.

scylla-rust-wrapper/src/integration_testing.rs
muzarski: Good question. I'll need to check that, because in the examples of such pattern I have on top of my head, the string is already allocated and owned by some other struct (which frees the string when dropped/free'd). I'll do it tomorrow.

muzarski:
Rebased on master and added `cass_statement_free` to unit test.

scylla-rust-wrapper/Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = ["database"]
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.1.0", features = [
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "32d179cb2", features = [
     "openssl-010",
     "metrics",
 ] }
```
wprzytula: Tip: If we release 1.2 soon (tomorrow?), you will be able to use a tag.

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -50,6 +50,9 @@ pub struct CassResult {
     pub tracing_id: Option<Uuid>,
     pub paging_state_response: PagingStateResponse,
     pub kind: CassResultKind,
+    // None only for tests - currently no way to mock coordinator in rust-driver.
+    // Should be able to do so under "cpp_rust_unstable".
+    pub(crate) coordinator: Option<Coordinator>,
```
wprzytula, line 66: :thought_balloon: If you believe it should be added to Rust driver, go ahead.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -541,6 +541,35 @@ pub unsafe extern "C" fn cass_statement_set_host_inet(
     CassError::CASS_OK
 }
 
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_statement_set_node(
```
wprzytula, line 545: commit `statement: implement cass_future_set_node`: `future` -> `statement`.

scylla-rust-wrapper/src/integration_testing.rs
wprzytula: @muzarski Any findings since?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -29,17 +29,20 @@ use std::sync::Arc;
 use thiserror::Error;
 use uuid::Uuid;
 
+#[derive(Debug)]
 pub enum CassResultKind {
```
wprzytula, line 33: :recycle: Please extract deriving of these `Debug`s to another commit, in order to declutter this commit.

scylla-rust-wrapper/src/integration_testing.rs
muzarski: Oh, I forgot about it. I've inspected all `write_str_to_c` calls and did not find anything suspicious. We always provide a pointer to data that is owned by some structure. We never do something like:
1. allocate the string
2. pass the ownership to the user
3. user does not have a method to free the string (leak)

scylla-rust-wrapper/src/query_result.rs
muzarski: I'm not that convinced it should be added. WDYT about this Option?

muzarski:
v1.1: Addressed @wprzytula comments

scylla-rust-wrapper/src/query_result.rs
wprzytula: I dislike that we allow for a special case that is possible only in tests.

wprzytula (approved):

scylla-rust-wrapper/src/query_result.rs
muzarski: OK, I'll try to do it. I don't want it to block this PR though.

PR #300, opened 2025-05-14
Don't panic when null pointer is provided - log error and return instead

### LINKED ISSUE #253: Don't panic when user provides null pointers to methods
Currently, we assume that user uses the API correctly, and we panic if the provided pointer is null.

Example:
```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn cass_cluster_set_connect_timeout(
    cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
    timeout_ms: c_uint,
) {
    let cluster = BoxFFI::as_mut_ref(cluster_raw).unwrap();
    cluster.session_builder.config.connect_timeout = Duration::from_millis(timeout_ms.into());
}
```

This `unwrap()` panics if `cluster_raw` is null.

This should be converted to something like:
```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn cass_cluster_set_connect_timeout(
    cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
    timeout_ms: c_uint,
) {
    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
        tracing::error!("Provided null pointer to cass_cluster_set_connect_timeout!");
    };
    cluster.session_builder.config.connect_timeout = Duration::from_millis(timeout_ms.into());
}
```

muzarski:
Fixes: https://github.com/scylladb/cpp-rust-driver/issues/253

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -381,7 +381,10 @@ pub unsafe extern "C" fn cass_cluster_set_application_name_n(
     app_name: *const c_char,
     app_name_len: size_t,
 ) {
-    let cluster = BoxFFI::as_mut_ref(cluster_raw).unwrap();
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
+        tracing::error!("Provided null cluster pointer to cass_cluster_set_application_name_n!");
+        return;
+    };
     let app_name = unsafe { ptr_to_cstr_n(app_name, app_name_len) }
         .unwrap()
         .to_string();
@@ -407,7 +410,10 @@ pub unsafe extern "C" fn cass_cluster_set_application_version_n(
     app_version: *const c_char,
     app_version_len: size_t,
 ) {
-    let cluster = BoxFFI::as_mut_ref(cluster_raw).unwrap();
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
+        tracing::error!("Provided null cluster pointer to cass_cluster_set_application_version_n!");
+        return;
+    };
     let app_version = unsafe { ptr_to_cstr_n(app_version, app_version_len) }
         .unwrap()
         .to_string();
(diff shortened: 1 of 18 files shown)


Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -996,7 +1067,13 @@ pub unsafe extern "C" fn cass_cluster_set_use_beta_protocol_version(
...
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
+        tracing::error!(
+            "Provided null cluster pointer to cass_cluster_set_use_beta_protocol_version!"
+        );
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
     cluster.use_beta_protocol_version = enable == cass_true;
 
     CassError::CASS_OK
```
wprzytula, line 1079: :question: Does `cass_cluster_set_use_beta_protocol_version()` do anything? AFAIK Rust driver supports just one protocol version, so this is a no-op. Shouldn't we rather mark this as deprecated and/or remove its implementation?

wprzytula (approved):
Phew! This was one of the most boring PRs to review ;P

scylla-rust-wrapper/src/cluster.rs
muzarski: It is read here:
```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn cass_cluster_set_protocol_version(
    cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
    protocol_version: c_int,
) -> CassError {
    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
        tracing::error!("Provided null cluster pointer to cass_cluster_set_protocol_version!");
    };

    if protocol_version == 4 && !cluster.use_beta_protocol_version {
        // Rust Driver supports only protocol version 4
        CassError::CASS_OK
    } else {
        CassError::CASS_ERROR_LIB_BAD_PARAMS
    }
}
```

PR #306, opened 2025-05-28
root: add MAINTENANCE.md

### LINKED ISSUE #284: Add `MAINTENANCE.md`
This would serve as a helpful resource for developers/maintainers. A great reference to create such file would be rust-driver: https://github.com/scylladb/scylla-rust-driver/blob/main/MAINTENANCE.md.

We should definitely include information about the release process. Most importantly, during each release we should update the corresponding entries in `cassandra.h`:
```c
#define CASS_VERSION_MAJOR 2
#define CASS_VERSION_MINOR 16
#define CASS_VERSION_PATCH 1
#define CASS_VERSION_SUFFIX ""
```

muzarski:
Fixes: #284 

It's heavily inspired by rust-driver maintenance.md (which is awesome btw!).

Currently it contains information about release process.

## Pre-review checklist

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -1,0 +1,146 @@
+# Maintenance
+
+The following document includes information how to release cpp-rust-driver and
+other information / procedures useful for maintainers. It is heavily inspired by https://github.com/scylladb/scylla-rust-driver/blob/v1.2.0/MAINTENANCE.md.
+
+## Documentation
+
+TODO (no docs written yet)
+Note: Due to API compatibility, cpp-driver docs apply to cpp-rust-driver in general. Though, some discrepancies should be well-documented in cpp-rust-driver (some of them are already mentioned in `cassandra.h`).
+
+## Version bump commit
+
+There are two places to be addressed when bumping the version of the driver.
+1. `scylla-rust-wrapper/Cargo.toml` -> `version` field should be adjusted
+2. `include/cassandra.h` -> there are `CASS_VERSION_[MAJOR/MINOR/PATCH/SUFFIX]` definitions which should be adjusted accordingly
+
+## Releasing
+
+IMPORTANT: Read this whole document before attempting to do a release.
+
+Prerequisites:
+- Have write access to GitHub repo (so that you are able to push to master and create a new branch).
+- Decide what the new version should be.
+
+### Releasing a new version
+
+1. Checkout `master`.
+2. Create some new branch (which you'll later push to your fork).
+3. (Optional, but recommended) Create a commit with changes applied by running `cargo update`. This will update the dependencies in `Cargo.lock` file, which we keep track of in the repository.
+4. Create commit that bumps version numbers of the driver (see `Version bump commit` section in this document).
+5. Prepare release notes (see the template at the bottom of this document).
+6. (Optional, but recommended) Push the branch to your fork and create the PR to `master` so that another maintainer can review your work.
+Description of this PR should consist of just release notes. **NOTE: Preferably merge this PR with CLI as described in next step. If you really want to use GitHub UI only use REBASE merge.** This is to make all commits appear on the version branch, without any merge commits / squash commits.
+7. Checkout `master` and fast-forward merge your changes (`git merge --ff-only your_branch`).
+8. Create a new tag (e.g. `git tag -a v1.2.0 -m "Release 1.2.0"`).
+9. Push `master` and the new tag, preferably using atomic push (e.g. `git push --atomic origin master v1.2.0`).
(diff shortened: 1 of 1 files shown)


MAINTENANCE.md
```diff
@@ -0,0 +1,145 @@
+# Maintenance
+
+The following document includes information how to release cpp-rust-driver and
+other information / procedures useful for maintainers. It is heavily inspired by https://github.com/scylladb/scylla-rust-driver/blob/v1.2.0/MAINTENANCE.md.
+
+## Documentation
+
+TODO (no docs written yet)
```
wprzytula, line 8: :question: Can you mention that due to the API compatibility with the original CPP driver, its docs apply in general?

wprzytula (approved):

wprzytula:
I'd also like to have a comprehensive documentation of:
- how to build and run examples,
- how to build and run tests,
- how to build static vs dynamic lib.

It's possible that these are partly described in other readmes, but please make sure that exhaustively enough.

MAINTENANCE.md
```diff
@@ -0,0 +1,146 @@
...
+Note: Due to API compatibility, cpp-driver docs in apply to cpp-rust-driver in general. Though, some discrepancies should be well-documented in cpp-rust-driver (some of them are already mentioned in `cassandra.h`).
```
wprzytula:
```suggestion
Note: Due to API compatibility, cpp-driver docs apply to cpp-rust-driver in general. Though, some discrepancies should be well-documented in cpp-rust-driver (some of them are already mentioned in `cassandra.h`).
```

wprzytula (changes requested):

muzarski:
> I'd also like to have a comprehensive documentation of:
> 
>     * how to build and run examples,

Yesterday we checked that it's not possible to build all examples. There is no neat way to filter out some examples during building. Someone needs to modify the cmake files so examples are linked with `testing_unimplemented.cpp`.

In general, the examples are built by providing `-DCASS_BUILD_EXAMPLES=ON` to cmake command. 

>     * how to build and run tests,

It is documented in README.md

>     * how to build static vs dynamic lib.

WDYM? Just `cmake .. && make` (called in `build` directory` produces:
- libscylla-cpp-driver_static.a
- libscylla-cpp-driver.so

I think you meant, how to build test/examples while linking with static libraries (by default, we shared libraries are used). To answer your question, the easiest way is to show you the cmake options defined at the top of `CMakeLists.txt` file:
```cmake
option(CASS_BUILD_EXAMPLES "Build examples" OFF)
option(CASS_BUILD_INTEGRATION_TESTS "Build integration tests" OFF)
option(CASS_BUILD_SHARED "Build shared library" ON)
option(CASS_BUILD_STATIC "Build static library" ON)
option(CASS_BUILD_TESTS "Build tests" OFF)
option(CASS_BUILD_UNIT_TESTS "Build unit tests" OFF)
option(CASS_DEBUG_CUSTOM_ALLOC "Debug custom allocator" OFF)
option(CASS_INSTALL_HEADER "Install header file" ON)
option(CASS_INSTALL_HEADER_IN_SUBDIR "Install header file under 'include/cassandra'" OFF)
option(CASS_INSTALL_PKG_CONFIG "Install pkg-config file(s)" ON)
option(CASS_MULTICORE_COMPILATION "Enable multicore compilation" ON)
option(CASS_USE_BOOST_ATOMIC "Use Boost atomics library" OFF)
option(CASS_USE_KERBEROS "Use Kerberos" OFF)
option(CASS_USE_LIBSSH2 "Use libssh2 for integration tests" OFF)
option(CASS_USE_OPENSSL "Use OpenSSL" ON)
option(CASS_USE_STATIC_LIBS "Link static libraries when building executables" OFF)
option(CASS_USE_STD_ATOMIC "Use std::atomic library" ON)
option(CASS_USE_ZLIB "Use zlib" OFF)
option(CASS_USE_TIMERFD "Use timerfd (Linux only)" ON)
option(CASS_USE_LIBUV "Use libuv" OFF)
```

As you can see, `CASS_BUILD_SHARED` and `CASS_BUILD_STATIC` are `ON` by default (thus, we get `.so` and `.a` libraries).
To use static libs for executables (tests/examples), you just need to specify `-DCASS_USE_STATIC_LIBS=ON`.

> It's possible that these are partly described in other readmes, but please make sure that exhaustively enough.

This is true. I'll take a look anyway and provide some missing information if needed. I'll do it in separate PR, though.

muzarski:
@wprzytula I am merging - if you have more requests regarding documents, see https://github.com/scylladb/cpp-rust-driver/pull/307

PR #317, opened 2025-06-06
Allow clippy public API related complaints

wprzytula:
This is fully analogous to https://github.com/scylladb/scylla-rust-driver/pull/1359/commits/15448c4fe47a31507abe4113d4ebb2b43d57fa3f. Pasting the commit message:

> Turns out clippy has a terrible and shocking default: it won't make
> suggestions that require introducing breaking changes.
> This means that e.g. if we introduce a type which has a naming that
> clippy would normally complain about, it won't if the type is pub!

1. Clippy is made check public API, too.
2. The affected names are fixed.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -16,7 +16,7 @@ pub(crate) use crate::cass_data_types::CassValueType;
 
 #[derive(Clone, Debug)]
 #[cfg_attr(test, derive(PartialEq, Eq))]
-pub struct UDTDataType {
+pub struct UdtDataType {
     // Vec to preserve the order of types
     pub field_types: Vec<(String, Arc<CassDataType>)>,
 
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/src/cass_types.rs
Lorak-mmk, line 1: Commit:  make UDT types' names Rust case conformant 

I don't agree with removing `DataType` suffix from `UDTDataType` - I think changing to `UdtDataType` is enough.
Why?
- The prefix is not really redundant. `DataType` is a concept from cpp-driver, so a separate thing from `defined type` (which `DT` in `UDT` expands to).
- `Udt` may suggest it holds the contents of UDT, while it really only represents the type.
- It introduces inconsistency with other variants of `CassDataTypeInner`.

Lorak-mmk (changes requested):

Lorak-mmk (approved):

PR #318, opened 2025-06-06
Cratewide API privatisation

wprzytula:
# Motivation

Lots of items in the driver are exposed as `pub` for no good reason. This involves:
- `pub` structs and enums,
- `pub` traits,
- `pub` fields,
- `pub` functions and methods.

This has a number of drawbacks:
1. _[Lack of clarity]_ It's unclear what items are part of the true public API. It's, fortunately, a less of an issue in the specific case of CPP-Rust Driver, because the whole API is defined in `cassandra.h`.
2. _[Clutter]_ Items that are unused are not detected by static checks. This is because `pub` items are considered used by the compiler.
3. _[Small performance penalty]_ `pub` functions must be exposed as proper (noninlined) functions, and they must obey calling conventions. `pub` types and their `pub` fields must not be individually optimised, so that other crates can understand how to use them.

So privatisation is profitable!

### Privatisation

I perused the whole crate and replaced `pub` with `pub(crate)` (or even `pub(self)`) where possible.

### Found & removed a bug in materialized views API
    
Two functions in `metadata.rs` were missing the `#[no_mangle]` attribute, which is necessary for FFI compatibility:
- `cass_materialized_view_meta_partition_key_count`;
- `cass_materialized_view_meta_clustering_key`.

So I added the missing attribute. I have an idea how to prevent similar bugs - will open an issue.

### Cleanup after privatisation

A number of items were found to be unused. I either removed them or silenced the warnings.

## Pre-review checklist


wprzytula:
CI failed because of clippy. Once UDT is no longer part of public API, clippy starts complaining about its name violating the conventions.
As a result, we should merge #317 first.

Lorak-mmk (commented):
Why did you decide to not move the commit changing the privacy to the end? This way no commit would emit warnings.

Lorak-mmk (approved):
Nice!

PR #321, opened 2025-06-12
DC-aware LBP - better support the deprecated parameters

### LINKED ISSUE #315: LBP: support DCAwarePolicy's `used_hosts_per_remote_dc` parameter
# CPP Driver semantics

Both `cass_{cluster,execution_profile}_set_load_balancing_dc_aware()` accept the **deprecated** `used_hosts_per_remote_dc` parameter.
The CPP Driver, although Datastax [deprecated the parameter in 2019](https://datastax-oss.atlassian.net/jira/software/c/projects/CPP/issues/CPP-705), has retained support for the parameter. We must decide if we want to support it or not. Or, perhaps we want to support it partially (with caveats).

## Big Picture semantics

The `used_hosts_per_remote_dc` controls the number of hosts _per remote DC_ (i.e., for every remote DC the number is the same, and all remote DC have their own limit) that may be used be the driver.

It influences two key aspects of the driver's behaviour:
1. Which (how many) remote nodes the driver opens connections to.
2. Which (how many) remote nodes the driver puts in the query plans.

Both limits work similarly in the following way:
- Driver keeps a set of nodes per every DC.
- Driver maintains the state (UP/DOWN) of each node, based exclusively on CQL events.
- When Driver needs to do something with the known nodes (open connections or send a request), first it filters them by retaining nodes from the local DC and only first (in an arbitrary driver's order) `used_hosts_per_remote_dc` from every remote DC; other remote nodes are ignored.
- After the initial metadata fetch, connection pools are created only for the non-ignored known nodes.
  - **NOTE:** How I understand the code, connection pool can later be created only for newly added nodes. This means that if all the nodes from a remote DC that we have opened connections to get DOWN, then **the driver will not open connections to another node from that DC**!
- Upon a request issued, the LBP includes only non-ignored nodes in the query plan.

## In-depth semantics

1. Used in `DCAwarePolicy::DCAwareQueryPlan::compute_next()` to limit the maximum index of host in the host vec (essentially wrapping modulo the bound). This is only about the query plan, not about opened connections.
2. Used in `DCAwarePolicy::distance()` to bound the number of hosts per remote (`used_hosts_per_remote_dc` first remote hosts get REMOTE distance, rest gets IGNORE).

```c++
CassHostDistance DCAwarePolicy::distance(const Host::Ptr& host) const {
  if (local_dc_.empty() || host->dc() == local_dc_) {
    return CASS_HOST_DISTANCE_LOCAL;
  }

  const CopyOnWriteHostVec& hosts = per_remote_dc_live_hosts_.get_hosts(host->dc());
  size_t num_hosts = std::min(hosts->size(), used_hosts_per_remote_dc_);
  for (size_t i = 0; i < num_hosts; ++i) {
    if ((*hosts)[i]->address() == host->address()) {
      return CASS_HOST_DISTANCE_REMOTE;
    }
  }

  return CASS_HOST_DISTANCE_IGNORE;
}
```

  - As can be seen, first `used_hosts_per_remote_dc_` hosts (in the arbitrary map order) in each remote DC are considered REMOTE, the rest are IGNOREd. **This does not take into account whether connections exist or not**; it is purely about the node considered UP or DOWN based on cluster topology events.

  - This is used, among others, to determine whether to open a connection to given node.
    - Upon Session initialization, in `SessionBase::on_initialize()`, `cluster_->available_hosts()` is called. This returns only hosts that are not ignored.
      - `Cluster::is_host_ignored()` scans all LBPs and returns true if **ALL** of them ignore the host.
    - `RequestProcessor::internal_host_add()` also creates a connection pool for a new host only if the host is not ignored.

# Discussion

## Option 1: Dated deprecation -> simply ignore?
The discussion should start with mention that the parameter was deprecated in 2019. This means that there is a probably simplest possible solution: to ignore the parameter, emitting a warning when it's non-zero. OTOH, the problem is that our current behaviour (the only configurable in the driver) is more like `+inf` (all connections to remote nodes opened) passed as a parameter than like `0` (no connections to remote nodes opened).

## Option 2: Let's support it! (At least partly...)

1. **Query plan - doable**. LBP has access to nodes' status (based on opened connections, not CQL events, as a more robust indicator). So it could easily retain only a specified number of nodes in the query plan. This, however, seems to be the less important part of this semantics, with (not) opening connections being the more important (as we want to avoid overhead of excess, anyway unused, TCP/CQL connections).

2. **Connection pool - not doable with current Rust Driver**.  - We can only support the particular case of `used_hosts_per_remote_dc=0` by employing `DcHostFilter`.
  - For `used_hosts_per_remote_dc>0`, we first have to establish the semantics.
    - As noted above, CPP Driver misbehaves when the only nodes from a given remote DC that have connections opened to get DOWN. Should we fix this (IMO a bug) in our CPP-Rust implementation?
    - For nodes in remote DCs, should we open connections to all shards or just one shard?
    - Does the present API (an `unsigned` parameter) mean that there always must be some bound given? In such situation, if a user wants effectively no bound, they must pass a large number (say, max possible `unsigned`). This is ugly.
    - Why would anyone want to pass anything different than `{0, 1, +inf}`? What's even the usecase of this?

### A solution to consider: support just two cases:
  -  `=0` -> employ `DcHostFilter`, disallow DC failover in the LBP, and we're done. Our semantics here are in line with the CPP Driver.
  - `!=0` -> Treat this as `+inf` case - don't limit connections to remote nodes. Emit a warning that the semantics are different than in CPP Driver, + attach a deprecation notice.

wprzytula:
# Background
The parameters: `used_hosts_per_remote_dc`,  `allow_remote_dcs_for_local_cl` of the DC-aware LBP have been long deprecated. However, they've been still supported in the cpp-driver, so we should probably support it. At least, we should accept the same set of values as the cpp-driver, which has not been true: values other than `0` would trigger `CASS_ERROR_LIB_BAD_PARAMS`, and `0` would result in a behaviour different than cpp-driver's: it would be ignored and not restrict remote DC hosts at all.

# What's done

## More informative warnings/errors

The messages logged when an invalid / not supported / partially supported value is given as a parameter are made more informative.

## Made `used_hosts_per_remote_dc` semantics closer to cpp-driver's

Due to Rust driver API's, it'd be very hard to support it correctly in the case of `>0`, because of reasons explained in     https://github.com/scylladb/cpp-rust-driver/issues/315. Therefore, the partial support approach is taken:
- case `=0` -> employ `CassHostFilter`, disable DC failover and we're done. Our semantics here are in line with the CPP Driver.
- case `>0` -> Treat this as `+inf` case - don't limit connections nor requests to remote nodes. Emit a warning that the semantics are different than in CPP Driver.

### Allowed `used_hosts_per_remote_dc > 0` and described its limitations

Any nonzero value is treated as `+inf`, which means that the datacenter failover is permitted without restrictions. All remote nodes can have opened connections to and be targeted with requests.

### Fixed `used_hosts_per_remote_dc = 0` semantics and made it fully consistent with cpp-driver's

1. Datacenter failover is disabled, meaning that remote nodes do not appear in query plans.
2. If a DC is remote to **all** LBPs **and** those LBPs disallow DC failover, then the driver will not connect to any hosts in that DC. This addresses the need to minimise overhead and costs of needless connecting to remote DCs.

## Implemented full support of `allow_remote_dcs_for_local_cl`

This was as simple as using another filtering logic in the existing `FilteringLoadBalancingPolicy`. 

# Testing

## Unit tests

Unit tests were introduced for internal abstractions used to configure the `CassHostFilter` to filter out remote nodes, if those are disallowed by all LBPs. Existing unit tests were adjusted to expect `used_hosts_per_remote_dc > 0` as supported.
A unit test was written for the backbone of `allow_remote_dcs_for_local_cl`, `DcLocalConsistencyAllowance`.

## Integration tests

1. Enabled `DCExecutionProfileTest_DCAware` test.
2. Enabled `DcAwarePolicyTest_UsedHostsRemoteDc` test.
3. Written a missing test for the `allow_remote_dcs_for_local_cl` option: `DcAwarePolicyTest_AllowRemoteDcsForLocalCl`.

# Unresolved questions

Behaviour of both our previous and current implementation of functions taking `used_hosts_per_remote_dc` parameter differ from one described in `cassandra.h`. Should we adjust the documentation there to be in line with the implementation?

Fixes: #315

## Pre-review checklist

~~- [ ] I have enabled appropriate tests in `Makefile` in `{SCYLLA,CASSANDRA}_(NO_VALGRIND_)TEST_FILTER`.~~

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -838,8 +838,19 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
-    if used_hosts_per_remote_dc != 0 || allow_remote_dcs_for_local_cl != 0 {
-        // TODO: Add warning that the parameters are deprecated and not supported in the driver.
+    if used_hosts_per_remote_dc != 0 {
+        tracing::error!(
+            "cass_*_set_load_balance_dc_aware(_n): `used_hosts_per_remote_dc` parameter is no longer \
+            supported in the driver. Set it to 0 to avoid this error."
+        );
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    }
+
+    if allow_remote_dcs_for_local_cl != 0 {
+        tracing::error!(
+            "cass_*_set_load_balance_dc_aware(_n): `allow_remote_dcs_for_local_cl` parameter is no longer \
+            supported in the driver. Set it to 0 to avoid this error."
+        );
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -827,6 +827,9 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
     allow_remote_dcs_for_local_cl: cass_bool_t,
 ) -> CassError {
     if local_dc_raw.is_null() || local_dc_length == 0 {
+        tracing::error!(
+            "Provided null or empty local DC name to cass_*_set_load_balance_dc_aware(_n)!"
+        );
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -854,17 +854,12 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
         false
     };
 
-    if allow_remote_dcs_for_local_cl != 0 {
-        tracing::error!(
-            "cass_*_set_load_balance_dc_aware(_n): `allow_remote_dcs_for_local_cl` parameter is no longer \
-            supported in the driver. Set it to 0 to avoid this error."
-        );
-        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
-    }
+    let allow_remote_dcs_for_local_cl = allow_remote_dcs_for_local_cl != 0;
 
     load_balancing_config.load_balancing_kind = Some(LoadBalancingKind::DcAware {
         local_dc: local_dc.to_owned(),
         permit_dc_failover,
+        allow_remote_dcs_for_local_cl,
     });
     load_balancing_config.filtering.dc_restriction = if permit_dc_failover {
         DcRestriction::None
@@ -1811,9 +1806,11 @@ mod tests {
                         Some(LoadBalancingKind::DcAware {
                             local_dc,
                             permit_dc_failover,
+                            allow_remote_dcs_for_local_cl,
                         }) => {
                             assert_eq!(local_dc, "eu");
                             assert!(!permit_dc_failover);
+                            assert!(!allow_remote_dcs_for_local_cl);
                         }
                         _ => panic!("Expected preferred dc"),
                     }
@@ -1849,8 +1846,8 @@ mod tests {
                         cass_cluster_set_load_balance_dc_aware(
                             cluster_raw.borrow_mut(),
                             c"eu".as_ptr(),
-                            42, // allow DC failover
-                            0
+                            42,        // allow DC failover
+                            cass_true  // allow remote DCs for local CL
                         ),
                         CassError::CASS_OK
                     );
@@ -1862,26 +1859,17 @@ mod tests {
                         Some(LoadBalancingKind::DcAware {
                             local_dc,
                             permit_dc_failover,
+                            allow_remote_dcs_for_local_cl,
                         }) => {
                             assert_eq!(local_dc, "eu");
                             assert!(permit_dc_failover);
+                            assert!(allow_remote_dcs_for_local_cl);
                         }
                         _ => panic!("Expected preferred dc"),
                     }
                 }
                 /* Test invalid configurations */
                 {
-                    // Nonzero (deprecated and unsupported) parameter
-                    assert_cass_error_eq!(
-                        cass_cluster_set_load_balance_dc_aware(
-                            cluster_raw.borrow_mut(),
-                            c"eu".as_ptr(),
-                            0,
-                            1
-                        ),
-                        CassError::CASS_ERROR_LIB_BAD_PARAMS
-                    );
-
                     // null pointers
                     assert_cass_error_eq!(
                         cass_cluster_set_load_balance_dc_aware(
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -826,10 +826,15 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
     used_hosts_per_remote_dc: c_uint,
     allow_remote_dcs_for_local_cl: cass_bool_t,
 ) -> CassError {
-    if local_dc_raw.is_null() || local_dc_length == 0 {
+    let Some(local_dc) = (unsafe { ptr_to_cstr_n(local_dc_raw, local_dc_length) }) else {
         tracing::error!(
-            "Provided null or empty local DC name to cass_*_set_load_balance_dc_aware(_n)!"
+            "Provided null or non-UTF-8 local DC name to cass_*_set_load_balance_dc_aware(_n)!"
         );
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
+    if local_dc_length == 0 {
+        tracing::error!("Provided empty local DC name to cass_*_set_load_balance_dc_aware(_n)!");
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
@@ -838,11 +843,9 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     }
 
-    let local_dc = unsafe { ptr_to_cstr_n(local_dc_raw, local_dc_length) }
-        .unwrap()
-        .to_string();
-
-    load_balancing_config.load_balancing_kind = Some(LoadBalancingKind::DcAware { local_dc });
+    load_balancing_config.load_balancing_kind = Some(LoadBalancingKind::DcAware {
+        local_dc: local_dc.to_owned(),
+    });
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -3,7 +3,9 @@ use crate::cass_error::CassError;
 use crate::cass_types::CassConsistency;
 use crate::exec_profile::{CassExecProfile, ExecProfileName, exec_profile_builder_modify};
 use crate::future::CassFuture;
-use crate::load_balancing::{CassHostFilter, LoadBalancingConfig, LoadBalancingKind};
+use crate::load_balancing::{
+    CassHostFilter, DcRestriction, LoadBalancingConfig, LoadBalancingKind,
+};
 use crate::retry_policy::CassRetryPolicy;
 use crate::ssl::CassSsl;
 use crate::timestamp_generator::CassTimestampGen;
@@ -864,6 +866,11 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
         local_dc: local_dc.to_owned(),
         permit_dc_failover,
     });
+    load_balancing_config.filtering.dc_restriction = if permit_dc_failover {
+        DcRestriction::None
+    } else {
+        DcRestriction::Local(local_dc.to_owned())
+    };
 
     CassError::CASS_OK
 }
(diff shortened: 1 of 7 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -826,23 +826,21 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
...
     allow_remote_dcs_for_local_cl: cass_bool_t,
 ) -> CassError {
-    if local_dc_raw.is_null() || local_dc_length == 0 {
+    let Some(local_dc) = (unsafe { ptr_to_cstr_n(local_dc_raw, local_dc_length) }) else {
         tracing::error!(
             "Provided null or empty local DC name to cass_*_set_load_balance_dc_aware(_n)!"
         );
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
-    }
+    };
```
Lorak-mmk: Commit: "set_lb_dc_aware: unwrap() -> pattern matching "

Previously we entered this also when string was non-null, but empty. Now we don't. It looks like a bug.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -833,8 +833,19 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
...
+            supported in the driver. Set it to 0 to avoid this error."
+        );
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    }
+
+    if allow_remote_dcs_for_local_cl != 0 {
+        tracing::error!(
+            "cass_*_set_load_balance_dc_aware(_n): `allow_remote_dcs_for_local_cl` parameter is no longer \
+            supported in the driver. Set it to 0 to avoid this error."
+        );
```
Lorak-mmk: Wait, so do I understand correctly that cpp-driver deprecated the possibility of DC-failover? If so, it does not look like something we should follow, it is imo important functionality.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -833,13 +833,19 @@ pub(crate) unsafe fn set_load_balance_dc_aware_n(
...
-            "cass_*_set_load_balance_dc_aware(_n): `used_hosts_per_remote_dc` parameter is no longer \
-            supported in the driver. Set it to 0 to avoid this error."
+    let permit_dc_failover = if used_hosts_per_remote_dc > 0 {
+        // TODO: update cassandra.h documentation to reflect this behaviour.
+        tracing::warn!(
+            "cass_*_set_load_balance_dc_aware(_n): `used_hosts_per_remote_dc` parameter is only partially \
+            supported in the driver: `0` is supported correctly, and any value `>0` has the semantics of \"+inf\", \
+            which means no limit on the number of hosts per remote DC. This is different from the original cpp-driver! \
+            To clarify, you can understand this parameter as \"permit_dc_failover\", with `0` being `false` and `>0` \
+            being `true`."
```
Lorak-mmk, line 850: I understand it is difficult to support connection-opening part of this parameter. We could expose cpp_rust_unstable method to allow that.

What is the problem with exposing LBP part of that? You can wrap LBP, and make an iterator that counts returned nodes per dc, and filters out nodes of DCs that reached the limit.

Lorak-mmk: We could also handle the failover-for-local-cl parameter in wrapper.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk, line 1: Commit:  lbp: used_hosts_per_remote_dc=0 prevents opening connections 

typo ff a DC -> iff a DC

scylla-rust-wrapper/src/load_balancing.rs
```diff
@@ -49,6 +82,16 @@ impl FilteringInfo {
...
+        if let AllowedDcs::Whitelist(ref allowed_dcs) = self.allowed_dcs {
+            // If the host's DC is not in the whitelist of DCs, reject it.
+            if !allowed_dcs
+                .iter()
+                .any(|allowed_dc| allowed_dc.as_str() == dc)
+            {
+                return false;
+            }
+        }
+
```
Lorak-mmk, line 95: The default container for `contains` operation is Set. Why not use it? This applies to other things in FilteringInfo too.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cluster.rs
wprzytula: No; cpp-driver deprecated the possibility of **disallowing** DC-failover.

You're most likely confused because of the stated requirement to pass 0 (which would normally signify **no** failover). I agree this is confusing. This PR brings back reasonable semantics for `used_hosts_per_remote_dc`. If you believe it should be done for `allow_remote_dcs_for_local_cl`, I can do this, too.

Actually, we could support `allow_remote_dcs_for_local_cl`, by just using a wrapping LBP, as you noted in another comment. I like the idea.

scylla-rust-wrapper/src/cluster.rs
wprzytula: > I understand it is difficult to support connection-opening part of this parameter. We could expose cpp_rust_unstable method to allow that.

It's not the matter of the Rust driver's API; it's how its connection pooling works inside.
It's even more about the very cpp-driver's mechanism being flawed. See my findings in #315:

> After the initial metadata fetch, connection pools are created only for the non-ignored known nodes.
>  - NOTE: How I understand the code, connection pool can later be created only for newly added nodes. This means that if all the nodes from a remote DC that we have opened connections to get DOWN, then the driver will not open connections to another node from that DC!

With such flawed semantics AND a large and nontrivial effort that would be needed to do this in the Rust Driver's inner parts, I'm strongly for **not** supporting `<1, +inf)` interval in the same way as cpp-driver.


> 
> What is the problem with exposing LBP part of that? You can wrap LBP, and make an iterator that counts returned nodes per dc, and filters out nodes of DCs that reached the limit.

Yes, we could do that. But **what would be the benefit of this?** I see none. The important part is not to open unnecessary connections. With connections already opened, I don't see any benefit from omitting _a strict subset of_ remote targets from the query plan.

If you see any benefit, please share it. But not seeing it, I'm not going to implement this.

scylla-rust-wrapper/src/cluster.rs
wprzytula: Good catch!

scylla-rust-wrapper/src/load_balancing.rs
wprzytula: The expected number of elements in these collections is low. This is why Mikołaj and I chose to use `Vec`, as it's simpler and likely has lower overhead than Set.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: > With such flawed semantics

Oh I'm not suggesting to follow such weird behaviors. The important part is limiting the connections to N per remote DC - if we have exact semantics more that are more reasonable than cpp-driver, even better.

Ofc there is no need to do this in this PR.

wprzytula:
The `Scylla ITs (5.4.8)` check is failing. The logs are the following:

```
 ccm> ccm node1 stop --not-gently
ccm> ccm node2 stop --not-gently
ccm> ccm node1 start --wait-other-notice --wait-for-binary-proto
unknown file: Failure
C++ exception with description "Traceback (most recent call last):
  File "/home/runner/.local/bin/ccm", line 74, in <module>
    cmd.run()
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/cmds/node_cmds.py", line 196, in run
    self.node.start(not self.options.no_join_ring,
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/scylla_node.py", line 762, in start
    scylla_process = self._start_scylla(args=args, marks=marks, update_pid=update_pid,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/scylla_node.py", line 413, in _start_scylla
    node.watch_log_for_alive(self, from_mark=mark, timeout=t)
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/node.py", line 535, in watch_log_for_alive
    self.watch_log_for(tofind, from_mark=from_mark, timeout=timeout, filename=filename)
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/node.py", line 499, in watch_log_for
    raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Missing: " + str(
ccmlib.node.TimeoutError: 17 Jun 2025 13:10:28 [node3] Missing: ['127.0.0.1.* now UP']:
INFO  2025-06-17 13:08:04,593 [shard 0:main] raft_.....
See system.log for remainder
" thrown in TearDown().
[  FAILED  ] DcAwarePolicyTest.Integration_Cassandra_UsedHostsRemoteDc (165376 ms)
```

It looks like ccm encounters an error trying to start a node that was killed. Interestingly, the problem does not appear with the newer Scyllas. @Lorak-mmk any ideas?

My observations: the test that fails is one that I've newly added. 

EDIT: we decided that we can remove that old Scylla version from the CI.

Lorak-mmk (approved):

PR #322, opened 2025-06-17
IT: Support `set_record_contacted_hosts()` and `get_attempted_hosts_from_future()`

wprzytula:
This PR re-implements `set_record_contacted_hosts()` and associated `get_attempted_hosts_from_future()` integration test functions in the Rust wrapper.

For now, no test using it is enabled. This is because none of them has its other requirements met. In #321, I'll be able to enable DCAware test suite. 

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,4 +292,4 @@ endif
 	build/cassandra-integration-tests --version=${CASSANDRA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${CASSANDRA_NO_VALGRIND_TEST_FILTER}"
 
 run-test-unit: install-cargo-if-missing _update-rust-tooling
-	@cd ${CURRENT_DIR}/scylla-rust-wrapper; cargo test
+	@cd ${CURRENT_DIR}/scylla-rust-wrapper; RUSTFLAGS="--cfg cpp_rust_unstable --cfg cpp_integration_testing" cargo test
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -81,7 +81,7 @@ impl CassSessionInner {
         let exec_profile_map = cluster.execution_profile_map().clone();
         let host_filter = cluster.build_host_filter();
 
-        CassFuture::make_raw(Self::connect_fut(
+        let fut = Self::connect_fut(
             session_opt,
             session_builder,
             exec_profile_map,
@@ -91,7 +91,13 @@ impl CassSessionInner {
                 // If user did not set a client id, generate a random uuid v4.
                 .unwrap_or_else(uuid::Uuid::new_v4),
             keyspace,
-        ))
+        );
+
+        CassFuture::make_raw(
+            fut,
+            #[cfg(cpp_integration_testing)]
+            None,
+        )
     }
 
     async fn connect_fut(
@@ -255,10 +261,16 @@ pub unsafe extern "C" fn cass_session_execute_batch(
     };
 
     match request_timeout_ms {
-        Some(timeout_ms) => {
-            CassFuture::make_raw(async move { request_with_timeout(timeout_ms, future).await })
-        }
-        None => CassFuture::make_raw(future),
+        Some(timeout_ms) => CassFuture::make_raw(
+            async move { request_with_timeout(timeout_ms, future).await },
+            #[cfg(cpp_integration_testing)]
+            None,
+        ),
+        None => CassFuture::make_raw(
+            future,
+            #[cfg(cpp_integration_testing)]
+            None,
+        ),
     }
 }
 
@@ -297,19 +309,24 @@ pub unsafe extern "C" fn cass_session_execute(
     let mut statement = statement_opt.statement.clone();
 
     #[cfg(cpp_integration_testing)]
-    statement_opt.record_hosts.then(|| {
+    let recording_listener = statement_opt.record_hosts.then(|| {
         let recording_listener =
             Arc::new(crate::integration_testing::RecordingHistoryListener::new());
         match statement {
             BoundStatement::Simple(ref mut unprepared) => {
-                unprepared.query.set_history_listener(recording_listener);
+                unprepared
+                    .query
+                    .set_history_listener(Arc::clone(&recording_listener) as _);
             }
             BoundStatement::Prepared(ref mut prepared) => {
+                // It's extremely interesting to me that this `as _` cast is required
+                // for the type checker to accept this code.
                 Arc::make_mut(&mut prepared.statement)
                     .statement
-                    .set_history_listener(recording_listener);
+                    .set_history_listener(Arc::clone(&recording_listener) as _);
             }
         };
+        recording_listener
     });
 
     let statement_exec_profile = statement_opt.exec_profile.clone();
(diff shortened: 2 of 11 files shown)


scylla-rust-wrapper/src/session.rs
```diff
@@ -295,6 +295,23 @@ pub unsafe extern "C" fn cass_session_execute(
     let request_timeout_ms = statement_opt.request_timeout_ms;
 
     let mut statement = statement_opt.statement.clone();
+
+    #[cfg(cpp_integration_testing)]
+    statement_opt.record_hosts.then(|| {
+        let recording_listener =
+            Arc::new(crate::integration_testing::RecordingHistoryListener::new());
+        match statement {
```
Lorak-mmk: 🤔 Something I've been wondering about: should we always use `Arc` for trait objects? `Box` would allow to get rid of interior mutability in some cases. Perhaps we need `Arc` e.g. because of `iter` which has to spawn a task?

scylla-rust-wrapper/src/session.rs
```diff
@@ -255,10 +259,16 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
+        Some(timeout_ms) => CassFuture::make_raw(
+            async move { request_with_timeout(timeout_ms, future).await },
+            #[cfg(cpp_integration_testing)]
+            None,
+        ),
+        None => CassFuture::make_raw(
+            future,
+            #[cfg(cpp_integration_testing)]
+            None,
+        ),
```
Lorak-mmk, line 273: batches don't need listener?

scylla-rust-wrapper/src/session.rs
```diff
@@ -297,19 +307,22 @@ pub unsafe extern "C" fn cass_session_execute(
...
-    statement_opt.record_hosts.then(|| {
+    let recording_listener = statement_opt.record_hosts.then(|| {
         let recording_listener =
             Arc::new(crate::integration_testing::RecordingHistoryListener::new());
         match statement {
             BoundStatement::Simple(ref mut unprepared) => {
-                unprepared.query.set_history_listener(recording_listener);
+                unprepared
+                    .query
+                    .set_history_listener(recording_listener.clone());
```
Lorak-mmk: Nit: `Arc::clone`

scylla-rust-wrapper/src/session.rs
```diff
@@ -297,19 +307,22 @@ pub unsafe extern "C" fn cass_session_execute(
...
             }
             BoundStatement::Prepared(ref mut prepared) => {
                 Arc::make_mut(&mut prepared.statement)
                     .statement
-                    .set_history_listener(recording_listener);
+                    .set_history_listener(recording_listener.clone());
             }
```
Lorak-mmk: ditto

scylla-rust-wrapper/src/session.rs
```diff
@@ -449,31 +468,35 @@ pub unsafe extern "C" fn cass_session_prepare_from_existing(
...
-            .prepare(query.query.clone())
-            .await
-            .map_err(|err| (err.to_cass_error(), err.msg()))?;
+    CassFuture::make_raw(
+        async move {
+            let query = match &statement {
+                BoundStatement::Simple(q) => q,
+                BoundStatement::Prepared(ps) => {
+                    return Ok(CassResultValue::Prepared(ps.statement.clone()));
+                }
```
Lorak-mmk: ditto.

I wish there was a clippy lint about this (maybe there is and I don't know it?)

scylla-rust-wrapper/src/session.rs
```diff
@@ -503,28 +526,32 @@ pub unsafe extern "C" fn cass_session_prepare_n(
...
+    CassFuture::make_raw(
+        async move {
+            let session_guard = cass_session.read().await;
+            if session_guard.is_none() {
+                return Err((
+                    CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
+                    "Session is not connected".msg(),
+                ));
+            }
+            let session = &session_guard.as_ref().unwrap().session;
```
Lorak-mmk: ♻️ Now that `make_raw` got another argument, maybe it would be cleaner to extract this async block out and assign it to a local var?

Makefile
```diff
@@ -292,4 +292,4 @@ endif
 	build/cassandra-integration-tests --version=${CASSANDRA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${CASSANDRA_NO_VALGRIND_TEST_FILTER}"
 
 run-test-unit: install-cargo-if-missing _update-rust-tooling
-	@cd ${CURRENT_DIR}/scylla-rust-wrapper; cargo test
+	@cd ${CURRENT_DIR}/scylla-rust-wrapper; RUSTFLAGS="--cfg cpp_rust_unstable --cfg cpp_integration_testing" cargo test
```
Lorak-mmk, line 295: Unrelated: I find this style of Makefile very weird. One reason is that it assumes users env (in this case it uses apt). Other is that is performs unrelated actions (installing cargo using a package manager) when I told it to run tests.

Its fine to have targets / scripts that install stuff, but they should not be in dependencies of normal targets. If they are needed by CI, they should be explicitly called by CI.

src/testing.cpp
```diff
@@ -51,7 +47,44 @@ String get_host_from_future(CassFuture* future) {
...
+    }
+    std::sort(attempted_hosts.begin(), attempted_hosts.end());
+    return attempted_hosts;
+    */
+
+    StringVec attempted_hosts;
+
+    char* const concatenated_attempted_addresses = testing_future_get_attempted_hosts(future);
+
+    std::string const& concatenated_addresses = concatenated_attempted_addresses;
```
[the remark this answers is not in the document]
Lorak-mmk: My C++ is extremely rusty now, but this looks like an actual good catch from copilot, right?
With the current version, `concatenated_attempted_addresses` is coerced into a std::string, to which we then take reference. Isn't that coerced string dropped later, causing possible UB?

src/testing.cpp
```diff
@@ -51,7 +47,44 @@ String get_host_from_future(CassFuture* future) {
...
+        if (std::getline(stream, address, '\n')) {
+            if (!address.empty()) {
+                attempted_hosts.push_back(address);
+            }
+        } else {
+            break; // No more addresses to read.
+        }
+    }
+    std::sort(attempted_hosts.begin(), attempted_hosts.end());
+
```
Lorak-mmk, line 84: It's surprising that this is what the code is expecting, I'd guess that it will want the hosts in original order.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/session.rs
wprzytula: This is a valid question. In the next commit, however, I directly leverage the `Arc`, because I clone it into `CassFuture`.

scylla-rust-wrapper/src/session.rs
wprzytula: The API in IT is only implemented for statements.

Makefile
wprzytula: You're more than welcome to improve this Makefile!
Especially if this makes it run on my Fedora, or Arch.

src/testing.cpp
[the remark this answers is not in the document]
wprzytula: GPT-4o states that the created temporary `std::string` would have its lifetime extended to the scope of the `const&`.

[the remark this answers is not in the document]
wprzytula: Not to play with the devil, I'm moving to plain `std::string`, though.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: I see this marked as resolved, but not changed in code, so unresolving. Not an important point (certainly not a blocker), so we can merge without it.

Lorak-mmk (approved):

Lorak-mmk (approved):

PR #326, opened 2025-06-24
argconv: Make FFI traits unsafe

wprzytula:
The FFI traits are inherently unsafe. In order to know which trait is correct to be implemented, manual analysis of the CPP Driver's code is required. Thus the traits should be marked as unsafe, indicating that special care and reasoning is needed when implementing them.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -340,7 +340,7 @@ mod origin_sealed {
 /// Implement this trait for types that are allocated by the driver via [`Box::new`],
 /// and then returned to the user as a pointer. The user is responsible for freeing
 /// the memory associated with the pointer using corresponding driver's API function.
-pub trait BoxFFI: Sized + origin_sealed::FromBoxSealed {
+pub unsafe trait BoxFFI: Sized + origin_sealed::FromBoxSealed {
     /// Consumes the Box and returns a pointer with exclusive ownership.
     /// The pointer needs to be freed. See [`BoxFFI::free()`].
     fn into_ptr<CM: CMutability>(self: Box<Self>) -> CassPtr<'static, Self, (Exclusive, CM)> {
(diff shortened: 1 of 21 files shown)


Lorak-mmk:
Could you explain why the traits are unsafe? Specifically, if I mistakenly implement such trait for some type, how can I write safe code that exhibits UB?

wprzytula:
> Could you explain why the traits are unsafe? Specifically, if I mistakenly implement such trait for some type, how can I write safe code that exhibits UB?

The simplest argument is existence of the following **safe** method:
```rust
pub(crate) trait ArcFFI: Sized + origin_sealed::FromArcSealed {
    /// Converts shared owned pointer back to owned Arc.
    fn from_ptr<CM: CMutability>(ptr: CassPtr<'static, Self, (Shared, CM)>) -> Option<Arc<Self>> {
        // SAFETY:
        // The only way to obtain a pointer with shared ownership ('static lifetime) is
        // ArcFFI::into_ptr(). It converts an owned Arc into the pointer. It is thus safe
        // to convert such pointer back to owned Arc.
        unsafe {
            ptr.to_raw().map(|p| {
                Arc::from_raw(p)
            })
        }
    }
```

If both the trait and the method are safe, and yet they perform an inherently unsafe operation (namely, treating an arbitrary pointer as a pointer to an Arc allocation), then this is bad - an obvious violation of the unsafety contracts in Rust.

Is this enough for you?

Lorak-mmk:
> If both the trait and the method are safe, and yet they perform an inherently unsafe operation (namely, treating an arbitrary pointer as a pointer to an Arc allocation), then this is bad - an obvious violation of the unsafety contracts in Rust.

This implication is not true. Exposing the safe API that uses `unsafe` underneath is the whole idea behind safety of Rust.
In this specific case, I'm not aware of a safe way to create an invalid CassPtr, that's why I'm asking for safe code that triggers UB.

wprzytula:
Quoting https://stackoverflow.com/a/31628759:

> A trait is marked unsafe to indicate that it is possible to violate memory safety by implementing it at all. This is commonly because the trait has invariants that other unsafe code relies on being upheld, and that these invariants cannot be expressed any other way.

The `ArcFFI` trait has invariants (the memory pointing to the valid Arc allocation) that unsafe code in the method I enclosed above relies on being upheld. Those invariants cannot be expressed any other way, because it's Rust knows nothing about ownership kind of the particular CPP Driver type. At the same time, the `FFI` trait with its associated type that specifies the ownership is just perfect for the use case. Yet, it's unsafe in that there is a specific external assumption that is asserted by the fact of a type having it implemented.

Lorak-mmk:
If there are invariants that require unsafe, then you should be able to demonstrate a snippet that doesn't use unsafe and triggers UB. Otherwise, there is no need for this PR.

> The ArcFFI trait has invariants (the memory pointing to the valid Arc allocation) that unsafe code in the method I enclosed above relies on being upheld

The code you linked has argument of type `CassPtr<'static, Self, (Shared, CM)>`
How can I create a value of such type (for a `Self` that implements `ArcFFI`) that is not backed by Arc? I don't know any way to do this.

wprzytula:
OK, in a discussion offline I agreed to consider these traits safe.

PR #327, opened 2025-06-24
Don't panic on `cass_session_get_client_id()` on disconnected sessions

wprzytula:
## Problem statement
`cass_session_get_client_id()` contained a bug: for disconnected sessions, it would panic on `unwrap()`, because the `client_id` was only stored on the `CassConnectedSession`, which is `None` when the session is disconnected.

## Solution
This PR fixes the bug by storing the `client_id` on the `CassSessionInner` struct, so that it is always available, even when
the session is disconnected.
Upon session creation (`cass_session_new()`), a random UUID v4 is generated and stored in the `CassSessionInner`. If the user sets a custom client id in the `CassCluster`, it will replace the random one upon `cass_session_connect()`.

## Testing

A unit test is added, which asserts the following:
1. We can get a client ID from a not-yet-connected session.
2. The client ID is inherited from the cluster when connecting.
3. We can still get the client ID after disconnecting the session, which is still the cluster's client ID, if set.

## Open for discussion
Note that we're using `RUNTIME.block_on()` to drive the future returned from `RwLock::read_owned()` to completion. This is necessary because `RwLock::blocking_read()` does not have an owned version.

I had a longer thought about `blocking_read()` (which we use in some places) vs `RUNTIME.block_on(read[_owned]())`, and I think that `blocking_read()` is strictly worse: it simply blocks the current thread until the lock is acquired, while `RUNTIME.block_on()` will also allow other tasks to run while waiting for the lock, which is not only more efficient, but also prevents deadlocks in some cases. In a follow-up PR, I'm thus going to replace all `blocking_read()` with `RUNTIME.block_on(read[_owned]())`.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -37,13 +37,13 @@ use tokio::sync::RwLock;
 // CassSession would need to be a newtype wrapper around this struct
 // instead of a type alias.
 #[expect(unnameable_types)]
-pub struct CassSessionInner {
+pub struct CassConnectedSession {
     session: Session,
     exec_profile_map: HashMap<ExecProfileName, ExecutionProfileHandle>,
     client_id: uuid::Uuid,
 }
 
-impl CassSessionInner {
+impl CassConnectedSession {
     pub(crate) fn resolve_exec_profile(
         &self,
         name: &ExecProfileName,
@@ -77,7 +77,7 @@ impl CassSessionInner {
     }
 
     fn connect(
-        session_opt: Arc<RwLock<Option<CassSessionInner>>>,
+        session_opt: Arc<RwLock<Option<CassConnectedSession>>>,
         cluster: &CassCluster,
         keyspace: Option<String>,
     ) -> CassOwnedSharedPtr<CassFuture, CMut> {
@@ -156,7 +156,7 @@ impl CassSessionInner {
             .await
             .map_err(|err| (err.to_cass_error(), err.msg()))?;
 
-        *session_guard = Some(CassSessionInner {
+        *session_guard = Some(CassConnectedSession {
             session,
             exec_profile_map,
             client_id,
@@ -185,7 +185,7 @@ impl CassSessionInner {
     }
 }
 
-pub type CassSession = RwLock<Option<CassSessionInner>>;
+pub type CassSession = RwLock<Option<CassConnectedSession>>;
 
 impl FFI for CassSession {
     type Origin = FromArc;
@@ -193,7 +193,7 @@ impl FFI for CassSession {
 
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn cass_session_new() -> CassOwnedSharedPtr<CassSession, CMut> {
-    let session = Arc::new(RwLock::new(None::<CassSessionInner>));
+    let session = Arc::new(RwLock::new(None::<CassConnectedSession>));
     ArcFFI::into_ptr(session)
 }
 
@@ -211,7 +211,7 @@ pub unsafe extern "C" fn cass_session_connect(
         return ArcFFI::null();
     };
 
-    CassSessionInner::connect(session_opt, cluster, None)
+    CassConnectedSession::connect(session_opt, cluster, None)
 }
 
 #[unsafe(no_mangle)]
@@ -240,7 +240,7 @@ pub unsafe extern "C" fn cass_session_connect_keyspace_n(
     };
     let keyspace = unsafe { ptr_to_cstr_n(keyspace, keyspace_length) }.map(ToOwned::to_owned);
 
-    CassSessionInner::connect(session_opt, cluster, keyspace)
+    CassConnectedSession::connect(session_opt, cluster, keyspace)
 }
 
 #[unsafe(no_mangle)]
@@ -274,10 +274,10 @@ pub unsafe extern "C" fn cass_session_execute_batch(
             ));
         }
 
-        let cass_session_inner = &session_guard.as_ref().unwrap();
-        let session = &cass_session_inner.session;
+        let cass_connected_session = session_guard.as_ref().unwrap();
+        let session = &cass_connected_session.session;
 
-        let handle = cass_session_inner
+        let handle = cass_connected_session
             .get_or_resolve_profile_handle(batch_exec_profile.as_ref())
             .await?;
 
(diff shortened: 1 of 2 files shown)


scylla-rust-wrapper/src/session.rs
```diff
@@ -73,7 +73,7 @@ impl CassConnectedSession {
     }
 
     fn connect(
-        session_opt: Arc<RwLock<Option<CassConnectedSession>>>,
+        session_opt: Arc<RwLock<CassSessionInner>>,
         cluster: &CassCluster,
         keyspace: Option<String>,
```
Lorak-mmk: The name `session_opt` is no longer correct after this change.

scylla-rust-wrapper/src/session.rs
Lorak-mmk, line 1: Commit: session: store client id on disconnected session too 

> and I think that
> `blocking_read()` is strictly worse: it simply blocks the current
> thread until the lock is acquired, while `RUNTIME.block_on()` will
> also allow other tasks to run while waiting for the lock

When blocking_read is executed in a user thread, this isn't really true, right? And I think most our usages fall into this case.

scylla-rust-wrapper/src/session.rs
```diff
@@ -81,15 +81,18 @@ impl CassConnectedSession {
         let exec_profile_map = cluster.execution_profile_map().clone();
         let host_filter = cluster.build_host_filter();
 
+        let mut session_guard = RUNTIME.block_on(session_opt.write_owned());
+
+        if let Some(cluster_client_id) = cluster.get_client_id() {
+            // If the user set a client id, use it instead of the random one.
+            session_guard.client_id = cluster_client_id;
+        }
+
```
Lorak-mmk: I don't see a point of doing it this way, instead of passing `cluster.get_client_id()` to `Self::connect_fut` and performing the assignment there. It would require less changes.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/session.rs
wprzytula: Why?

scylla-rust-wrapper/src/session.rs
wprzytula: > When blocking_read is executed in a user thread, this isn't really true, right?
Why?
1. If using _current_thread_ executor, using `blocking_read()` ends up with a deadlock.
2. If using _multi_thread_ executor, using `RUNTIME.block_on(read())` will make the thread join the executor pool and temporarily serve as an executor thread, which may be beneficial or not.

scylla-rust-wrapper/src/session.rs
wprzytula: This is absolutely necessary for correctness.
Consider the previous way of doing the assignment in the future. Then, we would have no control over when the assignment really takes place wrt to the end of the `cass_session_connect()` function. In such case, the unit test I wrote would fail, because having returned from `cass_session_connect()` would not guarantee that the client id had been already set.

wprzytula: A related bug (one about locking the RwLock only in the async block instead of doing it synchronously beforehand) exists in other places in the code, which I uncovered today. By playing with request execution delays I managed to create a reproducer that makes a test fail due to unfortunate interleavings. You'll see more about this once I open another PR fixing that.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: Before, there was Option in the type, hence _opt. Now there is not.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: > because having returned from cass_session_connect() would not guarantee that the client id had been already set.

I don't see why we need this guarantee. `cass_session_connect` returns CassFuture that performs some work, and does not guarantee anything about session before that future completes.

wprzytula: OK, I understand the reasoning you described. However, it still makes more sense to me to perform as big part of the connection configuration as possible in a synchronous manner, before the function returns. In fact, this is what CPP Driver does (with some irrelevant code removed):
```c++
Future::Ptr SessionBase::connect(const Config& config, const String& keyspace) {
  Future::Ptr future(new SessionFuture());

  ScopedMutex l(&mutex_);
  if (state_ != SESSION_STATE_CLOSED) {
    future->set_error(CASS_ERROR_LIB_UNABLE_TO_CONNECT,
                      "Already connecting, closing, or connected");
    return future;
  }

  if (config.is_client_id_set()) {
    client_id_ = config.client_id();
  }

  config_ = config.new_instance();
  connect_keyspace_ = keyspace;
  connect_future_ = future;
  state_ = SESSION_STATE_CONNECTING;

  if (config.use_randomized_contact_points()) {
    random_.reset(new Random());
  } else {
    random_.reset();
  }

  cluster_.reset();
  ClusterConnector::Ptr connector(
      new ClusterConnector(config_.contact_points(), config_.protocol_version(),
                           bind_callback(&SessionBase::on_initialize, this)));

  ClusterSettings settings(config_);
  settings.control_connection_settings.connection_settings.client_id = to_string(client_id_);
  settings.disable_events_on_startup = true;

  connector->with_listener(this)
      ->with_settings(settings)
      ->with_random(random_.get())
      ->with_metrics(metrics_.get())
      ->connect(event_loop_.get());

  return future;
}

```

Lorak-mmk: > However, it still makes more sense to me to perform as big part of the connection configuration as possible in a synchronous manner, before the function returns.

Why? It complicates the code.

wprzytula: Because it gives stronger guarantees, just this. And we are closer to the CPP-Driver's behaviour.
I don't see how this complicates the code.

wprzytula: Ping. I'd like to merge this, and I'm convinced the approach I've taken is correct.

wprzytula: Well, I'm no longer convinced. Reason: use in callbacks - this will panic due to #329 problem.

wprzytula:
I've observed that the CI fails in the specific way when b93e8b952be0cb590f40df9c96ef69cc4f84046e is missing.

wprzytula:
Rebased on master.

Lorak-mmk (approved):
Approving with the assumption that changes in the future will remove using block_on for locking.

wprzytula:
Rebased on master.

PR #330, opened 2025-06-28
Fix serial consistency handling

wprzytula:

## Problem overview

Serial consistency is handled incorrectly in both: 1) statement/batch, 2) exec profiles/cluster. The requirements that are not always met:
- `CASS_CONSISTENCY_ANY` shall result in **not setting any** serial consistency in the request frame;
- `CASS_CONSISTENCY_UNKNOWN` shall have the semantics of _"ignore me and use the default"_. CPP Driver allows it for statements/batches/exec profiles and forbids for cluster (because there would be no default to be used over cluster-level configuration).

## Solution

All changes are done to align with the CPP Driver behaviour.

The commits do the following:
1. **Support for serial consistency `ANY` in statements and batches**:
   Fixed the semantics in statements and batches (`cass_{statement,batch}_set_serial_consistency`):
   - `ANY` is allowed.
   - `UNKNOWN` is allowed and means that the execution profile's setting is used (or cluster's, if there's no execution profile set). Unfortunately, as the Rust driver misses corresponding option to unset consistency on statements, this is unsupported for now.

2. (Bonus) **Refactored consistency handling in statements and batches**:
  Similarly to what is done about serial consistency, the code is prepared for the Rust Driver's change introducing statements' API to unset consistency. 

3. **Fixes to serial consistency handling in execution profiles**:
  Fixed the semantics in execution profiles (`cass_execution_profile_set_serial_consistency`):
    - `ANY` is allowed.
   - `UNKNOWN` is allowed and means that the cluster's setting is used. This is unsupported for now and will be taken care in the follow-up.

4. **Fixes to serial consistency handling in cluster**:
   - Cluster forbids `CASS_CONSISTENCY_UNKNOWN`.

5. **Removal of `SerialConsistency` conversion**:
   - Eliminated the `impl TryFrom<CassConsistency> for SerialConsistency` implementation.
   - Addressed misuse risks due to differing semantics in conversion contexts (statements/batches/execution profiles vs cluster).

6. **No serial consistency as the default for the cluster**:
   - CPP Driver, as a default, sets **no serial consistency**.
   - The default is adjusted to conform.
   - **Note:** Rust driver purposefully set the default serial consistency to `LocalSerial` instead of `Any`. See the commit message for considerations. Let's discuss this.

### Discussion needed
The last commit: should we change the CPP-Driver's default of serial consistency from `Any` to `LocalSerial`?

### Summary:
These changes improve correctness of serial consistency handling, ensuring alignment with the CPP Driver's behavior.

### Testing

Coming soon, I'm working on an integration test for consistency/serial consistency set in all possible ways. This is going to be in a follow-up.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -3,18 +3,16 @@ use crate::argconv::{
     FFI, FromBox,
 };
 use crate::cass_error::CassError;
-use crate::cass_types::CassConsistency;
-use crate::cass_types::{CassBatchType, make_batch_type};
+use crate::cass_types::{CassBatchType, CassConsistency, make_batch_type};
 use crate::config_value::MaybeUnsetConfig;
 use crate::exec_profile::PerStatementExecProfile;
 use crate::retry_policy::CassRetryPolicy;
 use crate::statement::{BoundStatement, CassStatement};
 use crate::types::*;
 use crate::value::CassCqlValue;
-use scylla::statement::Consistency;
 use scylla::statement::batch::Batch;
+use scylla::statement::{Consistency, SerialConsistency};
 use scylla::value::MaybeUnset;
-use std::convert::TryInto;
 use std::sync::Arc;
 
 pub struct CassBatch {
@@ -115,13 +113,48 @@ pub unsafe extern "C" fn cass_batch_set_serial_consistency(
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     };
 
-    let serial_consistency = match serial_consistency.try_into().ok() {
-        Some(c) => c,
-        None => return CassError::CASS_ERROR_LIB_BAD_PARAMS,
+    // cpp-driver doesn't validate passed value in any way.
+    // If it is an incorrect serial-consistency value then it will be set
+    // and sent as-is.
+    // Before adapting the driver to Rust Driver 0.12 this code
+    // set serial consistency if a user passed correct value and set it to
+    // None otherwise.
+    // I think that failing explicitly is a better idea, so I decided to return
+    // an error.
+    let Ok(maybe_set_serial_consistency) =
+        MaybeUnsetConfig::<Option<SerialConsistency>>::from_c_value(serial_consistency)
+    else {
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     };
-    Arc::make_mut(&mut batch.state)
-        .batch
-        .set_serial_consistency(Some(serial_consistency));
+
+    match maybe_set_serial_consistency {
+        MaybeUnsetConfig::Unset => {
+            // The correct semantics for `CASS_CONSISTENCY_UNKNOWN` is to
+            // make batch not have any opinion at all about serial consistency.
+            // Then, the default from the cluster/execution profile should be used.
+            // Unfortunately, the Rust Driver does not support
+            // "unsetting" serial consistency from a batch at the moment.
+            //
+            // FIXME: Implement unsetting serial consistency in the Rust Driver.
+            // Then, fix this code.
+            //
+            // For now, we will throw an error in order to warn the user
+            // about this limitation.
+            tracing::warn!(
+                "Passed `CASS_CONSISTENCY_UNKNOWN` to `cass_batch_set_serial_consistency`. \
+                This is not supported by the CPP Rust Driver yet: once you set some serial consistency \
+                on a batch, you cannot unset it. This limitation will be fixed in the future. \
+                As a workaround, you can refrain from setting serial consistency on a batch, which \
+                will make the driver use the serial consistency set on execution profile or cluster level."
+            );
+            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+        }
+        MaybeUnsetConfig::Set(serial_consistency) => {
+            Arc::make_mut(&mut batch.state)
+                .batch
+                .set_serial_consistency(serial_consistency);
+        }
+    };
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/batch.rs
+++ b/scylla-rust-wrapper/src/batch.rs
@@ -5,11 +5,13 @@ use crate::argconv::{
 use crate::cass_error::CassError;
 use crate::cass_types::CassConsistency;
 use crate::cass_types::{CassBatchType, make_batch_type};
+use crate::config_value::MaybeUnsetConfig;
 use crate::exec_profile::PerStatementExecProfile;
 use crate::retry_policy::CassRetryPolicy;
 use crate::statement::{BoundStatement, CassStatement};
 use crate::types::*;
 use crate::value::CassCqlValue;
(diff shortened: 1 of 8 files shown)


Lorak-mmk:
> UNKNOWN is allowed and means that the execution profile's setting is used (or cluster's, if there's no execution profile set). Unfortunately, as the Rust driver misses corresponding option to unset consistency on statements, this is unsupported for now.

I don't understand this. First you say it is allowed, then you say it is unsupported. Which is it? What are the semantics in this PR?

Lorak-mmk:
> UNKNOWN is forbidden, because there's no source of the default setting over the execution profile.

Isn't a cluster such a source?

scylla-rust-wrapper/src/batch.rs
```diff
@@ -86,13 +88,40 @@ pub unsafe extern "C" fn cass_batch_set_serial_consistency(
...
+            // Then, the default from the cluster/execution profile should be used.
+            // Unfortunately, the Rust Driver does not support
+            // "unsetting" serial consistency from a batch at the moment.
+            //
+            // FIXME: Implement unsetting serial consistency in the Rust Driver.
+            // Then, fix this code.
+            //
+            // For now, we will throw an error in order to warn the user
+            // about this limitation.
+            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
```
Lorak-mmk: Maybe we should also print a warning with explanation, so that the user doesn't have to chase this GH PR?

scylla-rust-wrapper/src/statement.rs
```diff
@@ -671,6 +688,32 @@ fn get_consistency_from_cass_consistency(consistency: CassConsistency) -> Option
...
+        CassConsistency::CASS_CONSISTENCY_LOCAL_SERIAL => {
+            Ok(MaybeUnset::Set(Some(SerialConsistency::LocalSerial)))
+        }
+        CassConsistency::CASS_CONSISTENCY_SERIAL => {
+            Ok(MaybeUnset::Set(Some(SerialConsistency::Serial)))
+        }
+        _ => Err(()),
+    }
+}
+
```
Lorak-mmk: For me this function is very confusing. We have 3 layers of wrapping (Result, MaybeUnset, Option), yet no comment about which layer means what.
The usage of MaybeUnset (which was written for use in CQL) as a generic container is also surprising.
Lastly, using `MaybeUnset::Set` to indicate that consistency should be UNSET is... weird at least.

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -698,16 +699,27 @@ pub unsafe extern "C" fn cass_execution_profile_set_serial_consistency(
...
+            // does not expose API to unset serial consistency (enabling semantics:
+            // "ignore me and use the default profile's setting").
+        }
+        MaybeUnset::Set(maybe_serial_consistency) => {
+            // CASS_CONSISTENCY_ANY -> None
+            // other consistency -> Some()
+            profile_builder
+                .modify_in_place(|builder| builder.serial_consistency(maybe_serial_consistency));
+        }
+    }
```
Lorak-mmk: Why are you neither returning nor printing an error in case of UNKNOWN?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/exec_profile.rs
wprzytula: This is a bug that is here because only a follow-up fixes this fully, and I forgot that this current state is incorrect.

wprzytula:
> > UNKNOWN is allowed and means that the execution profile's setting is used (or cluster's, if there's no execution profile set). Unfortunately, as the Rust driver misses corresponding option to unset consistency on statements, this is unsupported for now.
> 
> I don't understand this. First you say it is allowed, then you say it is unsupported. Which is it? What are the semantics in this PR?

I'm sorry for this. Initially, I misunderstood semantics of `UNKNOWN` in execution profiles (I thought they forbid it similarly to the cluster). Later I fixed implementation and descriptions, but some were accidentally left stale. Now it should be consistent everywhere that execution profiles accept `UNKNOWN`.

scylla-rust-wrapper/src/statement.rs
wprzytula: CQL `UNSET` behaves according to the third value of the Sobociński logic, which is exactly what we need here. This is because `CASS_CONSISTENCY_UNKNOWN` is a value that means "ignore me and use the default".

CQL `NULL`, OTOH, behaves according to to the third value of Łukasiewicz logic. We represent CQL `NULL` as `Option` in the driver, which is why I believe `MaybeUnset` is a good choice for use out of CQL context, too, to carry the Sobociński semantics.

`Result`, of course, is to handle the case of a totally invalid value provided as consistency.

If you agree with this explanation, I'll write explanatory comments.

Lorak-mmk: You explained why it is correct, and I agree that it is. I argue, for the reasons outlined in my previous comment, that it is not intuitive. 
Imo a better option would be to introduce enum like ConsistencyConfig with options Unset (=ANY), Fallback (=UNKNOWN), Set/Some.
Maybe name it ConfigValue or something like this if it can be used in other places (it probably can).

wprzytula: `UnsettableConfigOption<T>` - WDYT?

Lorak-mmk: Why `Unsettable`? Doesn't that mean "impossible to set"?

wprzytula: In my idea, it's "able to be unset".

wprzytula: `MaybeUnsetConfigOption` ?

Lorak-mmk: Ok. Option may be redundant - MaybeUnsetConfig IMO works just as well and is shorter.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -40,6 +40,8 @@ use crate::cass_compression_types::CassCompressionType;
 // According to `cassandra.h` the defaults for
 // - consistency for statements is LOCAL_ONE,
 const DEFAULT_CONSISTENCY: Consistency = Consistency::LocalOne;
+// - serial consistency for statements is ANY, which corresponds to None in Rust Driver.
+const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = None;
 // - request client timeout is 12000 millis,
 const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_millis(12000);
```
Lorak-mmk, line 46: Let's open an issue about changing this to local serial.

Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
wprzytula: https://github.com/scylladb/cpp-rust-driver/issues/335

PR #331, opened 2025-06-29
prepared: Fix consistency bug

wprzytula:
### The bug

The found bug is quite severe: **any prepared statement would have its consistency set to `One` by default**. This is surprising, because `One` hasn't been the default consistency for prepared (or any other kind of) statements in CPP Driver since 2015 (as per the changelog)...

### The expected behaviour

Anyway, the CPP Driver's behaviour upon preparing is to set consistency in prepared statements to `UNKNOWN`, which means "use the consistency configured in the exec profile/cluster". This is the behaviour of Rust Driver, too: `PreparedStatement` has no consistency set by default.

### The fix

Thus, the fix is quite simple: we just remove the line that sets consistency to `One` in the `cass_session_prepare_n` function.

### Testing

Again, the IT for consistencies is being written. The test has already confirmed the bug's existence and the fix.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/session.rs
+++ b/scylla-rust-wrapper/src/session.rs
@@ -19,7 +19,6 @@ use scylla::client::session::Session;
 use scylla::client::session_builder::SessionBuilder;
 use scylla::cluster::metadata::ColumnType;
 use scylla::errors::ExecutionError;
-use scylla::frame::types::Consistency;
 use scylla::observability::metrics::MetricsError;
 use scylla::policies::host_filter::HostFilter;
 use scylla::response::PagingStateResponse;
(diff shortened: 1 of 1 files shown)


Lorak-mmk (approved):
> "use the consistency configured in the exec profile/cluster".

What is the cpp-driver default for that? I see that the PR that introduced this `set_consistency` line aimed to align the behavior with cpp-driver - so perhaps it just went about it the wrong way?

wprzytula:
> > "use the consistency configured in the exec profile/cluster".
> 
> What is the cpp-driver default for that? I see that the PR that introduced this `set_consistency` line aimed to align the behavior with cpp-driver - so perhaps it just went about it the wrong way?

### Per-cluster default consistency

The default per-cluster consistency has changed in time.
1. One. (2014?)
2. LocalQuorum.
3. LocalOne. (2015)

So for many years it's been LocalOne.

### Per-exec profile default consistency

It is unset (`CASS_CONSISTENCY_UNKNOWN`), so by default, the cluster setting prevails.

### Per-statement default consistency

It is unset (`CASS_CONSISTENCY_UNKNOWN`), so by default, the exec profile setting prevails.

## Our buggy implementation 

The `PreparedStatement` obtained from `Session::prepare()` had its statement-level consistency explicitly set to `One`, which made it override cluster-level and exec profile-level consistencies. **Only if** someone called `cass_statement_set_consistency()` after calling `cass_prepared_bind()` would the consistency be _different_ than `One`.

This is why I believe this is a severe bug.

Lorak-mmk:
>     One. (2014?)
>     LocalQuorum.
>     LocalOne. (2015)

> So for many years it's been LocalOne.

Do you have some links to commits / PRs of that?
I'm interested in their reasoning behind the changes. Its surprising to move from LocalQuorum to LocalOne

PR #332, opened 2025-06-29
Fix driver defaults - cluster & exec profiles

wprzytula:

Stacked on #341 (first 3 commits).

## Problem Overview

The Rust wrapper diverges from the CPP Driver in several areas of configuration, including:
1. default settings,
2. request timeouts,
3. execution profile behavior, and
4. load balancing policies.

These inconsistencies can lead to confusion for users migrating from the CPP Driver or relying on its documented behavior. Additionally, certain code patterns in the Rust wrapper could benefit from refactoring for improved clarity and maintainability.

## Solution

All changes are made to align the Scylla Rust wrapper with the CPP Driver's behavior while improving code organization and defensive programming practices.

The commits do the following:

1. **Style and typo fixes**:
   - Fixes minor style and typographical issues in `exec_profile.rs`.

1. **Paste CPP Driver defaults in a comment**:
   - Adds comments listing the CPP Driver's default settings and configuration code to ensure no defaults are accidentally omitted in the Rust wrapper.

1. **Use MonotonicTimestampGenerator by default**:
   - Ensures the default timestamp generator matches the CPP Driver's behavior, as well as most of the drivers'.

1. **Enable TCP keepalive by default**:
   - Aligns with the CPP Driver, but sets the keepalive interval to 2 seconds instead of 0 seconds due to libuv's rejection of 0-second configurations (and Rust driver's complaints that 1 sec is too short, whatever).

1. **Refactor `build_session_builder` as a method**:
   - Moves `build_session_builder` to be a method of `CassCluster`, as it perfectly makes sense.

1. **Set default policies explicitly**:
   - Explicitly sets default retry and speculative execution policies in the default execution profile builder to prevent future changes from breaking functionality.

1. **Describe load balancing policy divergence**:
   - Adds comments explaining the divergence in load balancing policy behaviour between the CPP Driver and the Rust wrapper. The Rust wrapper defaults to RoundRobin, requiring explicit local DC configuration for DC-aware policies.

1. **Use cluster defaults for unset settings in execution profiles**:
   - Modifies `CassExecProfile` to use the cluster's default profile for settings not explicitly set in the execution profile, ensuring consistency with the CPP Driver.

1. **Allow unsetting retry policy in execution profile**:
   - Passing null pointer as a retry policy to `cass_execution_profile_retry_policy` shall unset the retry policy, making the driver use the retry policy set in the cluster. This commit makes it true.

1. **Crate-expose `CassPtr::null(_mut)` constructors**:
   - For use in tests.

1.  **Unit test fetching unset settings from cluster**:
   - Verifies that settings that are unset on execution profile are fetched from the cluster.

**The remaining commits are not described here. I decided it's enough to read the commit messages.**

## Discussion Needed

Point 7: Should we reconsider the default load balancing policy to match the CPP Driver's behavior (DCAware with automatic local DC initialization)? While the current approach avoids potential pitfalls, it diverges from the CPP Driver's default.

## Summary

These changes improve the correctness, maintainability, and alignment of the Scylla Rust wrapper with the CPP Driver's behavior. They also enhance code clarity and defensive programming practices.

## Testing

I included one unit test for the execution profiles. The test verifies that each setting is inherited from the cluster iff it's unset in the execution profile.

Integration tests for consistency and serial consistency settings are planned as a follow-up.

## Pre-review Checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -240,14 +240,14 @@ impl<'a, T: Sized, P: Properties> CassPtr<'a, T, P> {
 
 /// Pointer constructors.
 impl<T: Sized, P: Properties> CassPtr<'_, T, P> {
-    fn null() -> Self {
+    pub fn null() -> Self {
         CassPtr {
             ptr: None,
             _phantom: PhantomData,
         }
     }
 
-    pub(crate) fn is_null(&self) -> bool {
+    pub fn is_null(&self) -> bool {
         self.ptr.is_none()
     }
 
@@ -274,7 +274,7 @@ impl<T: Sized, P: Properties> CassPtr<'_, T, P> {
 
 /// Constructors for to exclusive pointers.
 impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
-    fn null_mut() -> Self {
+    pub(crate) fn null_mut() -> Self {
         CassPtr {
             ptr: None,
             _phantom: PhantomData,
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -20,7 +20,7 @@ use scylla::client::{PoolSize, SelfIdentity, WriteCoalescingDelay};
 use scylla::frame::Compression;
 use scylla::policies::host_filter::HostFilter;
 use scylla::policies::load_balancing::LatencyAwarenessBuilder;
-use scylla::policies::retry::RetryPolicy;
+use scylla::policies::retry::{DefaultRetryPolicy, RetryPolicy};
 use scylla::policies::speculative_execution::SimpleSpeculativeExecutionPolicy;
 use scylla::policies::timestamp_generator::{MonotonicTimestampGenerator, TimestampGenerator};
 use scylla::routing::ShardAwarePortRange;
@@ -185,7 +185,9 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
     let default_execution_profile_builder = ExecutionProfileBuilder::default()
         .consistency(DEFAULT_CONSISTENCY)
         .serial_consistency(DEFAULT_SERIAL_CONSISTENCY)
-        .request_timeout(Some(DEFAULT_REQUEST_TIMEOUT));
+        .request_timeout(Some(DEFAULT_REQUEST_TIMEOUT))
+        .retry_policy(Arc::new(DefaultRetryPolicy::new()))
+        .speculative_execution_policy(None);
 
     /* Default config options - according to `cassandra.h`:
      * ```c++
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -66,6 +66,13 @@ const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_millis(5000);
 const DEFAULT_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(30);
 // - keepalive timeout is 60 secs
 const DEFAULT_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(60);
+// - TCP keepalive interval (actually: delay before the first keepalive
+//   probe is sent) is 0 sec in the CPP Driver. However, libuv started
+//   to reject such configuration on purpose since 1.49, so let's follow
+//   its reasoning and set it to more.
+//   1 sec could make sense, but Rust driver warns if it's not greater
+//   than 1 sec (for performance reasons), so let's set 2 secs..
+const DEFAULT_TCP_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(2);
 // - default local ip address is arbitrary
 const DEFAULT_LOCAL_IP_ADDRESS: Option<IpAddr> = None;
 // - default shard aware local port range is ephemeral range
@@ -284,6 +291,7 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
             .schema_agreement_timeout(DEFAULT_MAX_SCHEMA_WAIT_TIME)
             .schema_agreement_interval(DEFAULT_SCHEMA_AGREEMENT_INTERVAL)
             .tcp_nodelay(DEFAULT_SET_TCP_NO_DELAY)
+            .tcp_keepalive_interval(DEFAULT_TCP_KEEPALIVE_INTERVAL)
             .connection_timeout(DEFAULT_CONNECT_TIMEOUT)
             .pool_size(DEFAULT_CONNECTION_POOL_SIZE)
             .write_coalescing(DEFAULT_ENABLE_WRITE_COALESCING)
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -1275,24 +1275,19 @@ pub unsafe extern "C" fn cass_cluster_set_token_aware_routing_shuffle_replicas(
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn cass_cluster_set_retry_policy(
     cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
-    retry_policy: CassBorrowedSharedPtr<CassRetryPolicy, CMut>,
+    cass_retry_policy: CassBorrowedSharedPtr<CassRetryPolicy, CMut>,
 ) {
     let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
         tracing::error!("Provided null cluster pointer to cass_cluster_set_retry_policy!");
         return;
     };
 
-    let retry_policy: Arc<dyn RetryPolicy> = match ArcFFI::as_ref(retry_policy) {
-        Some(CassRetryPolicy::Default(default)) => Arc::clone(default) as _,
-        Some(CassRetryPolicy::Fallthrough(fallthrough)) => Arc::clone(fallthrough) as _,
-        Some(CassRetryPolicy::DowngradingConsistency(downgrading)) => Arc::clone(downgrading) as _,
-        Some(CassRetryPolicy::Logging(logging)) => Arc::clone(logging) as _,
-        #[cfg(cpp_integration_testing)]
-        Some(CassRetryPolicy::Ignoring(ignoring)) => Arc::clone(ignoring) as _,
-        None => {
-            tracing::error!("Provided null retry policy pointer to cass_cluster_set_retry_policy!");
-            return;
-        }
+    let maybe_unset_cass_retry_policy = ArcFFI::as_ref(cass_retry_policy);
+    let MaybeUnsetConfig::Set(retry_policy): MaybeUnsetConfig<_, Arc<dyn RetryPolicy>> =
+        MaybeUnsetConfig::from_c_value_infallible(maybe_unset_cass_retry_policy)
+    else {
+        tracing::error!("Provided null retry policy pointer to cass_cluster_set_retry_policy!");
+        return;
     };
 
     exec_profile_builder_modify(&mut cluster.default_execution_profile_builder, |builder| {
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -188,6 +188,12 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
         .request_timeout(Some(DEFAULT_REQUEST_TIMEOUT))
         .retry_policy(Arc::new(DefaultRetryPolicy::new()))
         .speculative_execution_policy(None);
+    // Load balancing is set in LoadBalancingConfig::build().
+    // The default load balancing policy is DCAware in CPP Driver.
+    // NOTE: CPP Driver initializes the local DC to the first node the client
+    // connects to. This is tricky, a possible footgun, and hard to be done with
+    // the current Rust driver implementation, which is why do don't use DC awareness
+    // by default.
 
     /* Default config options - according to `cassandra.h`:
      * ```c++
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -1524,14 +1524,14 @@ pub unsafe extern "C" fn cass_cluster_set_consistency(
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     };
 
-    let Ok(maybe_set_consistency) = MaybeUnsetConfig::<Consistency>::from_c_value(consistency)
+    let Ok(maybe_set_consistency) = MaybeUnsetConfig::<_, Consistency>::from_c_value(consistency)
     else {
         // Invalid consistency value provided.
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     };
 
     match maybe_set_consistency {
-        MaybeUnsetConfig::Unset => {
+        MaybeUnsetConfig::Unset(_) => {
             // `CASS_CONSISTENCY_UNKNOWN` is not supported in the cluster settings.
             return CassError::CASS_ERROR_LIB_BAD_PARAMS;
         }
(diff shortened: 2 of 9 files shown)


scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -47,23 +63,77 @@ impl CassExecProfile {
...
+
+        if !self.overrides.speculative_execution_policy {
+            self.modify_in_place(|builder| {
+                builder.speculative_execution_policy(
+                    default_profile
+                        .get_speculative_execution_policy()
+                        .map(Arc::clone),
+                )
+            });
+        }
```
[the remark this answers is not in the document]
wprzytula: Not worth it. Each setting has its own kind of "default".

Makefile
```diff
@@ -121,8 +121,7 @@ CASSANDRA_NO_VALGRIND_TEST_FILTER := $(subst ${SPACE},${EMPTY},AsyncTests.Integr
 endif
 
 ifndef CCM_COMMIT_ID
-	# TODO: change it back to master/next when https://github.com/scylladb/scylla-ccm/issues/646 is fixed.
-	export CCM_COMMIT_ID := 5392dd68
+	export CCM_COMMIT_ID := master
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk WDYT?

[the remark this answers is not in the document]
wprzytula: I'd think that we prefer to have the newest CCM in our tests in order to:
1. Automatically propagate bug fixes
2. Catch bugs in our CI as they are introduced to CCM.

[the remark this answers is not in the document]
Lorak-mmk: I strongly agree with using master.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -293,6 +293,7 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
             .keepalive_timeout(DEFAULT_KEEPALIVE_TIMEOUT)
             .local_ip_address(DEFAULT_LOCAL_IP_ADDRESS)
             .shard_aware_local_port_range(DEFAULT_SHARD_AWARE_LOCAL_PORT_RANGE)
+            .timestamp_generator(Arc::new(MonotonicTimestampGenerator::new()))
     };
```
Lorak-mmk, line 313: In cpp-driver I see:
```
  void set_timestamp_gen(TimestampGenerator* timestamp_gen) {
    if (timestamp_gen == NULL) return;
    timestamp_gen_.reset(timestamp_gen);
  }
```

Is it even possible to not use client-side timestamps there?
Hmm... I looked at the docs and I see this function: ```cass_timestamp_gen_server_side_new```
So that would imply that either `TimestampGenerator` interface in cpp-driver allows for both client-side and server-side timestamps, or there is some other place that controls it. 
Interestingly, docs for this function also include: "Note: This is the default timestamp generator."
So are you really sure that Monotonic is the default in cpp-driver and there is no other place in the code that unexpectedly make the driver use different one by default?

In general defaulting to client-side timestamps seems weird to me, I'd prefer to default to server-side.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk, line 1: Commit: "exec_profile: use cluster defaults for unset settings "

The behavior this commit fixes looks like a serious bug. Was that really the case before that cluster (=default EP) settings were ignored?

Lorak-mmk, line 1: "(whether it makes sense or not is another topic
for dicussion): if consistency is not set in the execution profile,
it is set to the hardcoded default consistency, not the one from
the cluster's default profile."

We are aiming for a "drop-in" replacement, but I think that as with most other things it is a matter of balance.
For example, we are not interested in replicating cpp-drivers bugs.
When migrating to this driver, people will have to pay more attention that with ordinary update, so I think it is a good idea to have a reasonable defaults in cpp-driver, instead of tying everything to cpp-driver.
Not that cpp-driver changed its default consistency multiple times, so why can't we when implementing new driver?

scylla-rust-wrapper/src/exec_profile.rs
```diff
@@ -748,23 +748,30 @@ pub unsafe extern "C" fn cass_execution_profile_set_retry_policy(
...
-        Some(CassRetryPolicy::Ignoring(ignoring)) => Arc::clone(ignoring) as _,
+    let maybe_retry_policy: Option<Arc<dyn RetryPolicy>> =
+        ArcFFI::as_ref(retry_policy).map(|rp| match rp {
+            CassRetryPolicy::Default(default) => Arc::clone(default) as Arc<dyn RetryPolicy>,
+            CassRetryPolicy::Fallthrough(fallthrough) => Arc::clone(fallthrough) as _,
+            CassRetryPolicy::DowngradingConsistency(downgrading) => Arc::clone(downgrading) as _,
+            CassRetryPolicy::Logging(logging) => Arc::clone(logging) as _,
+            #[cfg(cpp_integration_testing)]
+            CassRetryPolicy::Ignoring(ignoring) => Arc::clone(ignoring) as _,
+        });
```
Lorak-mmk: Why is the first cast needed? It was not present before.

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk, line 1: Comit: "argconv: crate-expose null(_mut) constructors "

> Both are safe to be used, because the unsafe code that CassPtr
> always safely dereferences the pointer, i.e., only after nullity checks.

" the unsafe code that CassPtr always safely dereferences the pointer," - that does not look like a correct sentence.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/exec_profile.rs
wprzytula: Previously we had `Arc<dyn RetryPolicy>`. Now we have `Option<Arc<dyn RetryPolicy>>`. The presence of an `Option` wrapper somehow alters the type-inferring logic.

scylla-rust-wrapper/src/cluster.rs
wprzytula: They were ignored iff an execution profile was set for the executed statement.
This is the Rust Driver's semantics, that's where the bug comes from.

Lorak-mmk: So in Rust Driver the cluster-level default will be ignored when EP is set on a statement, even if this EP doesn't have some value set?

scylla-rust-wrapper/src/cluster.rs
wprzytula: > Is it even possible to not use client-side timestamps there?

Yes, by passing an instance of the server-side timestamp generator, which may be obtained by `cass_timestamp_gen_server_side_new`. This API is fully analogous to retry policy API, where you similarly cannot pass a NULL - some retry policy is always needed, which is also the case with timestamp generators. You can also change the used RP/timestamp generator.

> So are you really sure that Monotonic is the default in cpp-driver and there is no other place in the code that unexpectedly make the driver use different one by default?

CPP Driver, by default, **does use client-side timestamps**.

Excerpt from `topics/basics/client_side_timestamps/README.md`:

```md
# Client-side timestamps

**Note**: Cassandra 2.1+ is required.

Cassandra uses timestamps to serialize write operations. That is, values with a
more current timestamp are considered to be the most up-to-date version of that
information. By default, timestamps are assigned by the driver on the
client-side. This behavior can be overridden by configuring the driver to use a
timestamp generator or assigning a timestamp directly to a [`CassStatement`] or
[`CassBatch`].
```

Lorak-mmk: So there is bug in documentation, right?

scylla-rust-wrapper/src/cluster.rs
wprzytula: Yes, because in Rust Driver there is no notion of _unset_ setting in an exec profile. This is what I'm referring to with `MaybeUnset::Unset` - a condition that someone left a setting unset, which is equivalent to the semantics of `UNSET` in CQL (Sobociński logic).

scylla-rust-wrapper/src/cluster.rs
wprzytula: What bug? I don't see any.

Perhaps you misparse the `gen` part in `cass_timestamp_gen_server_side_new` as `generate` instead of `generator`.

Lorak-mmk: Documentation for `cass_timestamp_gen_server_side_new` says that it is the default generator.
Documentation for `cass_timestamp_gen_monotonic_new` does not say it is the default.

wprzytula: > Interestingly, docs for this function also include: "Note: This is the default timestamp generator."

Ah, I've missed this.

This is the commit that changed the default timestamp generator to Monotonic:
https://github.com/datastax/cpp-driver/commit/c705ea54a562a657248c7ff7b987b35c2fda2cc9
They indeed forgot to update the docs in `cassandra.h`.

See the related [Datastax Jira issue](https://datastax-oss.atlassian.net/jira/software/c/projects/CPP/issues/CPP-413?jql=project%20%3D%20%22CPP%22%20AND%20%28textfields%20~%20%22CPP-413%22%20OR%20issuekey%20%3D%20%22CPP-413%22%29%20ORDER%20BY%20created%20DESC).

Lorak-mmk: This issue says "For consistency with other drivers, the cpp-driver should enable client timestamps by default in the next non-patch release."

Do you know if this is really the case? I was under the impression that drivers generally default to server-side timestamps.

wprzytula: After we checked other drivers, we already know that **most of the drivers default to client-side timestamp generation**.

scylla-rust-wrapper/src/cluster.rs
wprzytula: > Not that cpp-driver changed its default consistency multiple times, so why can't we when implementing new driver?

Well, but the last such change occurred 10 years ago...
Since then, users could have become very used to the driver's behaviour and defaults.

wprzytula:
Rebased on master.
Now I'm going to use the new Rust Driver API to fix statement and batch.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: Ok, we can keep this default, but the documentation needs to be updated.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -188,6 +188,12 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
         .request_timeout(Some(DEFAULT_REQUEST_TIMEOUT))
         .retry_policy(Arc::new(DefaultRetryPolicy::new()))
         .speculative_execution_policy(None);
+    // Load balancing is set in LoadBalancingConfig::build().
+    // The default load balancing policy is DCAware in CPP Driver.
+    // NOTE: CPP Driver initializes the local DC to the first node the client
+    // connects to. This is tricky, a possible footgun, and hard to be done with
+    // the current Rust driver implementation, which is why do don't use DC awareness
+    // by default.
```
Lorak-mmk, line 196: Maybe something like this should be described in user-facing documentation?

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: Ok, but this change makes sense and fixes a possible footgun (I'm talking about fallback to cluster execution profile). I don't think it will break any workflows.

wprzytula:
Fixes thanks to new API exposed by Rust Driver:
1. request timeout is now properly set in batches.
1. (serial) consistency can now be unset in statements/batches.

wprzytula:
Rebased on master.

scylla-rust-wrapper/src/config_value.rs
Lorak-mmk, line 1: Commit: "parametrize MaybeUnsetConfigValue with a lifetime"

I don't understand why this change is necessary - commit message does not explain it well enough.
In struct T<X>, X may very well be a type with non-static lifetime. So what is the additional lifetime for?

scylla-rust-wrapper/src/config_value.rs
```diff
@@ -130,3 +133,31 @@ impl MaybeUnsetConfigValue<'static> for RequestTimeout {
...
+                CassRetryPolicy::Logging(logging) => Arc::clone(logging) as _,
+                #[cfg(cpp_integration_testing)]
+                CassRetryPolicy::Ignoring(ignoring) => Arc::clone(ignoring) as _,
+            })
+            .expect(
+                "None passed as CassRetryPolicy, which should not happen due to \
+                MaybeUnsetConfigValue construction invariants: `from_set_c_value` should not be called."
+            )
+        )
+    }
```
Lorak-mmk: Ok, now I see what problem you were trying to solve.
CValue is an associated type, so it needs to be specified per implementation. You added a lifetime to have a separate trait per lifetime, each implementing conversion from reference with this lifetime.

It looks to me like CValue should be a type parameter instead of associated type. The heuristic I know for thinking about this is input types vs output types. Input type - think arguments for Add operation. Output type - type of result of add operation.

In this case, CValue is input type (type we convert from). There is nothing preventing conversion from multiple different CValue types. It is possible you will have to split this trait in 2 to achieve this (one for conversion, one for final interface).

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cluster.rs
wprzytula: What piece of documentation do you mean?

scylla-rust-wrapper/src/cluster.rs
wprzytula: Let's do it the following way:
1. I'll open an issue about making this consistent (but diverging from the CPP Driver).
2. We'll merge this PR as-is.
3. I'll open another PR that addresses the issue.
4. We'll merge that one as well.

Rationale: I want to keep bug fixes (making the behaviour correct wrt expectations from the CPP Driver) separate from improvements (making the behaviour more correct than the CPP Driver). Let the points of divergence be as discoverable and standing out as possible.

scylla-rust-wrapper/src/config_value.rs
wprzytula: It seems to me you're right. Let me explore this path (moving from associated type to type parameter) and come back with the results.

wprzytula:
Rebased on master.

scylla-rust-wrapper/src/config_value.rs
wprzytula: I've succeeded. Thanks for this suggestion!

wprzytula:
Made `CValue` a type parameter instead of an associated type on `MaybeUnsetConfigValue`.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: ok

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: Probably cassandra.h. Is there any other user-facing documentation?

Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
wprzytula: I guess we'll have a Scylla Documentation, as for the other drivers.

wprzytula: Nowhere in `cassandra.h` there is any notice about this behaviour. I mean, `cassandra.h` says nothing about the peculiar behaviour of CPP Driver that the first contact point's DC is considered local.

wprzytula: Only the changelog mentions it:
```markdown
1.0.0-rc1
===========
Dec 22, 2014

Features
--------
* DC-aware load balancing policy is now the default and uses the DC of the first
  connected contact point. The DC-aware policy now has settings to control the
  number of remote hosts used in each DC (after exhausting the hosts in the
  local DC) and whether to ignore local consistency levels.
```

wprzytula: Once we have the Scylla Documentation for this driver, we should mention this divergence.

Issue #319, opened 2025-06-06
Prevent API regressions


wprzytula:
## Motivation
#318 revealed a bug: two functions were missing `#[no_mangle]` attribute, which effectively made them absent from the API.

### How can we prevent bugs arising involving deletion of a function belonging to the API?

Note: in the case of materialized views API, those functions were there, but with mangled names; so effectively, they were missing from the API.

### Idea

1. Create a C source file that calls **ALL** functions belonging to the driver's public API.
2. Divide the full API of the original CPP Driver into two categories:
    - supported or quasi-supported. 
      - by quasi-supported I mean functions that we implement, but they "do nothing" (because they are similarly no-ops in CPP Driver, or it's otherwise acceptable that we make them no-ops in CPP-Rust Driver).
    - nonsupported (not implemented).
3. For the nonsupported group, create another C source file with stub implementations of all nonsupported functions.
4. Assert that combination `(test_file.c, nonsupported_stubs.c, cpp-rust-driver.a)` links successfully.
    - for supported functions, we make sure that they are indeed exposed by the driver. Otherwise linker shouts at unresolved references.
    - for nonsupported functions, we make sure that they are indeed **NOT** exposed by the driver. Otherwise linker shouts at multiple symbols with the same name.

Lorak-mmk:
> For the nonsupported group, create another C source file with stub implementations of all nonsupported functions.

I think we already have something like this, no? Maybe it does not contain all the missing functions, but we could add them.


Long time ago I had a bit different idea. We could have an  attribute macro, lets call it `#[driver_export]`.
In the most basic form, it would apply all the needed attributes (`#[unsafe(no_mangle)]`, `pub extern "C"`). But it could potentially do more:
- Track which functions were exported and save that in `target`. Then we can have a script to generate a report about this.
- Check that signatures match with cassandra.h, including the pointer types.

PR #333, opened 2025-06-30
Expose public rust API & enable API lints

wprzytula:
Ref: #319

Stacked on: #330
Start review from 3368e28 - `collection,tuple: expose missed functions`.

## Problem Overview

The Scylla Rust wrapper aims to be a drop-in replacement for the CPP Driver, but several areas required improvement to ensure better alignment with the CPP Driver's public API and usability in Rust. These include:
- missing public API functions
  - some functions are unimplemented,
  - some implemented functions are just stubs and they do nothing,
  - some correctly implemented functions are not `pub`.
- inconsistent module organization,
- accessibility issues for integration testing.

Additionally, some types and functions were erroneously exposed or restricted.

### Big Picture

A new module - `api` - is introduced to re-export public Rust (`extern "C"`) API of the driver.
Thanks to that, we can enable new lints: `unnameable_types` and `unreachable_pub`, which are a good measure to safeguard against regressions in spirit of #319. On the way, a number of misconfiguration cases were found and fixed. 

### Per-commit

The per-commit changes are as follows:

1. **Privatize remaining types**:
   - Cleans up the codebase by privatizing types that were mistakenly exposed, ensuring a more consistent and maintainable API.

1. **Expose missed functions**:
   - Functions such as `cass_collection_data_type`, `cass_tuple_data_type`, and others were erroneously module-private, preventing their use outside their modules. This commit ensures these functions are publicly accessible, aligning with the CPP Driver's API.

1. **Hide the `CassIterator` enum from the public API**:    
   - The variants of the `CassIterator` should not be exposed to the public API, as they are an implementation detail of the Scylla Rust wrapper. A wrapper type is added to achieve this.

1. **Pubify `ArcFFI` trait**:
   - It's going to be useful for integration tests.

1. **Add `expect(unnameable_types)` in sealed traits**:
   - Like the Rust driver did. This particular use case depends on type unnameability.

1. **Group all public C API in one module**:
    - Introduces a new `api` module that re-exports all public C API functions from `cassandra.h`, grouped by their order in the header file. This improves visibility into implemented and missing functions, facilitates integration testing, and ensures the Rust wrapper is a drop-in replacement for the CPP Driver.
   - Incorrectly implemented functions are appropriately commented.
   - Unimplemented functions are commented out.

1. **Include public types in the API**:
   - Re-exports public types alongside public functions in the `api` module. This change ensures that public API functions can be used seamlessly in Rust, particularly for integration testing and other use cases.

1. **Enable `unnameable_types` and `unreachable_pub` lints**:
   - This has successfully caught some issues in the codebase.

## Discussion Needed

Is the format I chose for `api` module clear enough? Does it make unimplemented/incorrectly implemented functions discoverable enough?

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -76,6 +76,9 @@ macro_rules! make_c_str {
 pub(crate) use make_c_str;
 
 mod sealed {
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[expect(unnameable_types)]
     pub trait Sealed {}
 }
 
@@ -330,8 +333,19 @@ impl<T: Sized> CassPtr<'_, T, (Exclusive, CMut)> {
 }
 
 mod origin_sealed {
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[expect(unnameable_types)]
     pub trait FromBoxSealed {}
+
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[expect(unnameable_types)]
     pub trait FromArcSealed {}
+
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[expect(unnameable_types)]
     pub trait FromRefSealed {}
 }
 
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -414,7 +414,7 @@ pub trait BoxFFI: Sized + origin_sealed::FromBoxSealed {
 /// The data should be allocated via [`Arc::new`], and then returned to the user as a pointer.
 /// The user is responsible for freeing the memory associated
 /// with the pointer using corresponding driver's API function.
-pub(crate) trait ArcFFI: Sized + origin_sealed::FromArcSealed {
+pub trait ArcFFI: Sized + origin_sealed::FromArcSealed {
     /// Creates a pointer from a valid reference to Arc-allocated data.
     /// Holder of the pointer borrows the pointee.
     #[allow(clippy::needless_lifetimes)]
--- a/scylla-rust-wrapper/src/lib.rs
+++ b/scylla-rust-wrapper/src/lib.rs
@@ -54,7 +54,7 @@ macro_rules! include_bindgen_generated {
 }
 
 /// All numeric types are defined here.
-pub(crate) mod types {
+pub mod types {
     #![allow(non_camel_case_types)]
     // for `cass_false` and `cass_true` globals.
     #![allow(non_upper_case_globals)]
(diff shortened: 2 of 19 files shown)


scylla-rust-wrapper/src/argconv.rs
Lorak-mmk, line 1: Commit: " argconv: pubify ArcFFI trait "

I don't get how will it be useful, and why this one but not others.
Why should the integration tests even use anything other than C api?

Lorak-mmk, line 1: Commit: " argconv: expect(unnameable_types) in sealed traits "

This commit will introduce lint errors, since you expect this lint and it is not enabled.
Imo this should be merged with commit that enables this lint.

scylla-rust-wrapper/src/lib.rs
```diff
@@ -8,6 +8,7 @@ use tokio::runtime::Runtime;
 
 #[macro_use]
 mod binding;
+pub mod api;
 // pub, because doctests defined in `argconv` module need to access it.
 pub mod argconv;
 pub(crate) mod batch;
```
Lorak-mmk, line 14: `api` should be the only public mod now. I assume `argconv` will be made pub(crate) in further commits.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/argconv.rs
wprzytula: This is not true. `expect` passes if the lint would be triggered if enabled. I confirmed it empirically.

wprzytula: 1. The others (`BoxFFI`, `RefFFI`) are already public.
2. Functions of those `???FFI` traits are required for Rust code that uses `CassPtr`-based C API to type-check. Consider that `CassPtr` is not `Clone`. Then, it would be usable only once, because the C API accepts it by move. The only way to prevent it is borrowing it, using `borrow()` and `borrow_mut()` methods. These are provided by the FFI traits I mentioned.

scylla-rust-wrapper/src/lib.rs
wprzytula: It's a viable choice, but in such case `api` must re-export FFI abstractions (`CassPtr`, its properties and traits) from `argconv`. Do you believe it's the good direction?

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Huh, I didn't know this. Thanks!

scylla-rust-wrapper/src/lib.rs
Lorak-mmk: Let's keep it as is

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Ok

Lorak-mmk (approved):

wprzytula:
Rebased on master.

wprzytula:
> ## Discussion Needed
> 
> Is the format I chose for `api` module clear enough? Does it make unimplemented/incorrectly implemented functions discoverable enough?

@Lorak-mmk Please share your thoughts on this.

Lorak-mmk:
Yes I think it is fine

PR #336, opened 2025-07-06
Run examples in CI

### LINKED ISSUE #305: CI: run examples
We should run (working) examples in CI.

wprzytula:
## What's done

1. Stubs were added for functions that are called by examples. This is to enable building all examples.
1. libuv is now linked to all examples. This is because some examples need libuv. Note that it might make more sense to only link libuv against the particular examples that require it, but it seemed to me to be too much fuss.
1. Makefile has now targets to build all examples and run a specified list of examples.
   - One ScyllaDB node is run in Docker for the purpose of running examples.
1. All examples were checked.
   - successfully running examples were enabled;
   - failing examples were annotated with a comment why they fail.
1. A CI step that builds and runs examples is added to the _build, lint and test_ workflow.

Fixes: #305

## Pre-review checklist


examples/tuple/tuple.c
```diff
@@ -177,7 +177,8 @@ CassError select_from_tuple(CassSession* session) {
         const CassValue* value = cass_iterator_get_value(item);
 
         if (!cass_value_is_null(value)) {
-          if (cass_value_type(value) == CASS_VALUE_TYPE_VARCHAR) {
+          if (cass_value_type(value) == CASS_VALUE_TYPE_VARCHAR ||
+              cass_value_type(value) == CASS_VALUE_TYPE_TEXT) {
             const char* text;
             size_t text_length;
```
Lorak-mmk: When is `CASS_VALUE_TYPE_VARCHAR` used, and when `CASS_VALUE_TYPE_TEXT`?
If the new driver always returns `CASS_VALUE_TYPE_TEXT`, then perhaps value of `CASS_VALUE_TYPE_VARCHAR` should be made identical to it to preserve backward compatibility?

examples/drop_examples_keyspace/drop_examples_keyspace.c
Lorak-mmk, line 1: Commit: examples: add drop_examples_keyspace example 

The other alternative is to make examples also work with existing keyspaces (`CREATE KEYSPACE / TABLE IF NOT EXISTS). Imo it would be cleaner, not requiring this additional weird target.

tests/examples_cluster/docker-compose.yml
Lorak-mmk, line 1: I don't really understand why use docker here. You have already existing workflows in CI that use CCM. You could do the same instead of introducing another method.

CMakeLists.txt
```diff
@@ -236,7 +240,25 @@ add_subdirectory(scylla-rust-wrapper)
...
+    get_filename_component(EXAMPLE_NAME ${EXAMPLE_DIR} NAME)
+
+    # Check whitelist
+    if(IS_DIRECTORY ${EXAMPLE_DIR})
+      if(CASS_EXAMPLES_WHITELIST STREQUAL "" OR EXAMPLE_NAME IN_LIST CASS_EXAMPLES_WHITELIST)
+        message(STATUS "Adding example subdirectory ${EXAMPLE_DIR}, example name ${EXAMPLE_NAME}")
+        add_subdirectory(${EXAMPLE_DIR})
+      endif()
+    endif()
+  endforeach()
```
Lorak-mmk: I understand there is a need to split responsibilities between Makefile and CMake, but it could be a little cleaner: 
`CASS_EXAMPLES_WHITELIST` is a variable used by CMake, so it should be declared there, with a default value.
Then, Makefile can either overwrite it with examples that user chose, or not touch it at all.

CMakeLists.txt
```diff
@@ -240,6 +240,17 @@ add_subdirectory(scylla-rust-wrapper)
...
+  if(NOT DEFINED CASS_EXAMPLES_WHITELIST) # If someone runs CMake straight, not from the Makefile.
+    execute_process(
+      COMMAND make -f ../Makefile -p -n # Print the makefile's variables without executing it.
+      COMMAND grep -E "^SCYLLA_EXAMPLES_TO_RUN :="  # Find the list of examples to run.
+      COMMAND awk -F ":= " "{print $2}" # Extract the value after the ":=", which is a whitespace-separated list.
+      OUTPUT_VARIABLE SCYLLA_EXAMPLES_TO_RUN
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    string(REPLACE " " ";" CASS_EXAMPLES_WHITELIST ${SCYLLA_EXAMPLES_TO_RUN})
+    string(APPEND CASS_EXAMPLES_WHITELIST drop_examples_keyspace)
```
Lorak-mmk: Oh no, let's not make this build system more cursed please.
CMake should be the build system, and makefile should contain shortcuts for useful stuff.

Lorak-mmk (changes requested):

examples/drop_examples_keyspace/drop_examples_keyspace.c
wprzytula: They would also have to drop the keyspace beforehand, because they often write the same data to the same keyspace/table. I'm still not convinced it's worth modyfing the examples.

CMakeLists.txt
wprzytula: I wanted to put the definition of `CASS_EXAMPLES_WHITELIST` in Makefile in order to be consistent with the test filter. It makes sense to have a list of enables/disabled tests/examples in one place.

examples/tuple/tuple.c
wprzytula: Those values are taken straight from the CQL protocol. Do you suggest diverging from CQL for compatibility purposes, by making `VARCHAR` value a CQL `TEXT` value?

CMakeLists.txt
wprzytula: WDYT? Do you still believe the list should be in CMakeLists.txt?

CMakeLists.txt
wprzytula: I'm open for ideas how to improve this.

wprzytula:
Rebased on master.

CMakeLists.txt
Lorak-mmk: I think so, because Cmake is the place that makes use of this list (this is different thab the tests, where the makefile itself uses this lidt, right?).

wprzytula: This inconsistency makes me very sad tbh.

Lorak-mmk: I don't see inconsistency, because those are totally different things.
Test filter is a parameter to test binary - nothing to do with cmake.
Example list is a build system thing (which targets we want to build). Think of it as defining a list of files or dependencies in CMake.

CMakeLists.txt
Lorak-mmk: When you move this variable to CMake, this won't be neccesary anymore, right?

CMakeLists.txt
wprzytula: If I put the list of examples to CMake (build system), how will I choose the list of examples to **run**? AFAIU, it's Make that is responsible for choosing a subset of examples to run.

tests/examples_cluster/docker-compose.yml
wprzytula: Which workflows in CI use CCM? Well, I agree that integration tests use CCM. However, they are calling it through the bridge layer, from C++. Here I'd have to use it straight from the CLI. Why is it better than a ready docker-compose taken from Rust Driver?

examples/tuple/tuple.c
Lorak-mmk: What I aim for is compatibility. Doesn't the need for this change prove that users may also encounter this issue in their code?

wprzytula: It does. That's why I'm asking for confirmation. Are we allowed to modify a constant that is related to the CQL protocol, in order to preserve backward compatibility in case of a returned query result?

Lorak-mmk: As discussed offline, this indicates a bug in cpp-rust-driver: NativeType::Text should be mapped to `CASS_VALUE_TYPE_VARCHAR`

wprzytula: Opened #344.

wprzytula:
Rebased on master, dropping the commit that adjusts the examples from VARCHAR to TEXT.

CMakeLists.txt
wprzytula: Resolved by having only a "build all examples" capability in CMake, as previously.

wprzytula:
Rebased on master.
CMake is reverted to its original form, in which it only supports building all examples.

scylla-rust-wrapper/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ else()
   set(CMAKE_Rust_FLAGS "--cfg cpp_rust_unstable")
 endif()
 
-if(CASS_BUILD_INTEGRATION_TESTS)
+if(CASS_BUILD_INTEGRATION_TESTS OR CASS_BUILD_EXAMPLES)
   set(CMAKE_Rust_FLAGS "${CMAKE_Rust_FLAGS} --cfg cpp_integration_testing")
 endif()
```
Lorak-mmk, line 33: We need to make sure this flag is not set in published binaries - we don't want stubs there.

Lorak-mmk (approved):

scylla-rust-wrapper/CMakeLists.txt
wprzytula: I've read the packaging logic carefully and my conclusion is that the used build command is correct wrt not including stubs in the generated package.

For redhat derivates, this line is important (`dist/redhat/scylla-cpp-rust-driver.spec`):
```sh
%cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RustFLAGS="-Cforce-frame-pointers=yes --cap-lints=warn"
```

For debian, it's similar (`dist/debian/debian/rules`):
```sh
dh_auto_configure -- -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RustFLAGS="--cap-lints warn -C linker=$(DEB_HOST_GNU_TYPE)-gcc -C link-arg=-Wl,-Bsymbolic-functions -C link-arg=-Wl,-z,relro"
```

I'm suspicious about the `--cap-lints warn` part. Why would anyone want to include this? Essentially, this converts all errors to warnings, effectively making the compilation process pass even if it would fail otherwise.
[I've talked to ChatGPT on this topic](https://chatgpt.com/share/687e3468-3f60-8009-b16e-1daf3429ac1f).

PR #342, opened 2025-07-10
codewide: delete unimplemented function stubs

wprzytula:
There exist several function implementations from `cassandra.h` in the wrapper that just non-working stubs, not correct implementations. This is contrary to the politics of the unimplemented functions, which says that unimplemented functions should not be present in the wrapper at all, in order to trigger linker errors if the user tries to use them in their applications.

This commit removes those stubs, so that the wrapper does not pretend to implement those functions. Users will correctly start getting linker errors if they try to use those functions.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/api.rs
+++ b/scylla-rust-wrapper/src/api.rs
@@ -60,7 +60,7 @@ pub mod cluster {
         cass_cluster_set_blacklist_filtering_n,
         cass_cluster_set_client_id,
         // cass_cluster_set_cloud_secure_connection_bundle, UNIMPLEMENTED
-        cass_cluster_set_cloud_secure_connection_bundle_n, // FIXME: shouldn't this be unimplemented?
+        // cass_cluster_set_cloud_secure_connection_bundle_n, UNIMPLEMENTED, stub present
         // cass_cluster_set_cloud_secure_connection_bundle_no_ssl_lib_init, UNIMPLEMENTED
         // cass_cluster_set_cloud_secure_connection_bundle_no_ssl_lib_init_n, UNIMPLEMENTED
         cass_cluster_set_coalesce_delay,
@@ -79,7 +79,7 @@ pub mod cluster {
         cass_cluster_set_credentials_n,
         cass_cluster_set_execution_profile,
         cass_cluster_set_execution_profile_n,
-        cass_cluster_set_exponential_reconnect, // FIXME: Should be unimplemented!
+        // cass_cluster_set_exponential_reconnect, UNIMPLEMENTED, stub present
         // cass_cluster_set_host_listener_callback, UNIMPLEMENTED
         cass_cluster_set_latency_aware_routing,
         cass_cluster_set_latency_aware_routing_settings,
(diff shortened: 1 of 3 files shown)


Lorak-mmk (approved):

Lorak-mmk:
What has changed since my approval?

wprzytula:
> What has changed since my approval?

It was merely a rebase on master.

Lorak-mmk:
This doesn't look like a rebase, was it really? https://github.com/scylladb/cpp-rust-driver/compare/eab966a7226b6422a6845bd7a1978f4af273503b..894cebc3b704bd410292d9a94e5d54de62187d0b

scylla-rust-wrapper/src/api.rs
```diff
@@ -940,4 +932,18 @@ pub mod integration_testing {
...
+    /// and at the same time the functions are not yet implemented in the wrapper.
+    #[rustfmt::skip]
+    pub use crate::integration_testing::stubs::{
+        CassCustomPayload,
+        cass_cluster_set_cloud_secure_connection_bundle_n,
+        cass_cluster_set_exponential_reconnect,
+        cass_cluster_set_use_randomized_contact_points,
+        cass_custom_payload_new,
+        cass_future_custom_payload_item,
+        cass_future_custom_payload_item_count,
```
Lorak-mmk: why skip rustfmt?

scylla-rust-wrapper/src/integration_testing.rs
```diff
@@ -361,3 +361,95 @@ pub unsafe extern "C" fn testing_retry_policy_ignoring_new()
...
+        CassError::CASS_OK
+    }
+
+    #[unsafe(no_mangle)]
+    pub extern "C" fn cass_future_custom_payload_item_count(
+        _future: CassBorrowedExclusivePtr<CassFuture, CMut>,
+    ) -> size_t {
+        0
+    }
+}
```
Lorak-mmk, line 455: Can't this whole module be only compiled when testing?

scylla-rust-wrapper/src/api.rs
wprzytula: Because it groups multiple functions in one line, whereas I prefer to have them one per line.

scylla-rust-wrapper/src/integration_testing.rs
wprzytula: It is, because it's a child module of the `integration_testing` module, which itself is `#[cfg(cpp_integration_testing)]`.

wprzytula:
> This doesn't look like a rebase, was it really? https://github.com/scylladb/cpp-rust-driver/compare/eab966a7226b6422a6845bd7a1978f4af273503b..894cebc3b704bd410292d9a94e5d54de62187d0b

Indeed it wasn't. I got mixed up in chronology.

scylla-rust-wrapper/src/api.rs
Lorak-mmk: Then please describe this in a comment, don't disable rustfmt without explanation.

scylla-rust-wrapper/src/integration_testing.rs
Lorak-mmk: Ah, I didn't notice that.

Lorak-mmk (approved):

PR #343, opened 2025-07-10
Introduce an IT for consistency

wprzytula:
This PR introduces the first integration test for the CPP-Rust Driver that is written in Rust.

The test is based on an analogous one that is present in the Rust Driver. It finally convinces me that the consistencies are handled correctly.

## Pre-review checklist

- ~~[] I have implemented Rust unit tests for the features/changes introduced.~~
- ~~[] I have enabled appropriate tests in `Makefile` in `{SCYLLA,CASSANDRA}_(NO_VALGRIND_)TEST_FILTER`.~~
- ~~[] I added appropriate `Fixes:` annotations to PR description.~~

### DIFF
--- a/scylla-rust-wrapper/tests/integration/utils.rs
+++ b/scylla-rust-wrapper/tests/integration/utils.rs
@@ -1,0 +1,203 @@
+use bytes::BytesMut;
+use futures::Future;
+use libc::c_char;
+use scylla_cpp_driver::api::error::{CassError, cass_error_desc};
+use scylla_cpp_driver::api::future::{
+    CassFuture, cass_future_error_code, cass_future_error_message, cass_future_free,
+    cass_future_wait,
+};
+use scylla_cpp_driver::argconv::{CMut, CassOwnedSharedPtr, ptr_to_cstr, ptr_to_cstr_n};
+use scylla_cpp_driver::types::size_t;
+use std::collections::HashMap;
+use std::ffi::CString;
+use std::net::SocketAddr;
+use std::str::FromStr;
+use std::sync::Arc;
+
+use scylla_proxy::{
+    Condition, Node, Proxy, ProxyError, Reaction as _, RequestFrame, RequestOpcode,
+    RequestReaction, RequestRule, ResponseFrame, RunningProxy, ShardAwareness,
+};
+
+pub(crate) fn setup_tracing() {
+    let _ = tracing_subscriber::fmt::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_writer(tracing_subscriber::fmt::TestWriter::new())
+        .try_init();
+}
+
+pub(crate) async fn test_with_3_node_dry_mode_cluster<F, Fut>(
+    shard_awareness: ShardAwareness,
+    test: F,
+) -> Result<(), ProxyError>
+where
+    F: FnOnce([String; 3], RunningProxy) -> Fut,
+    Fut: Future<Output = RunningProxy>,
+{
+    let proxy1_uri = format!("{}:9042", scylla_proxy::get_exclusive_local_address());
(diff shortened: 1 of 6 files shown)


wprzytula:
Rebased on master and appeased clippy.

scylla-rust-wrapper/tests/integration/utils.rs
```diff
@@ -0,0 +1,226 @@
...
+            unsafe { ptr_to_cstr(cass_error_desc(errcode)) },
+            unsafe { ptr_to_cstr_n(message, message_len) },
+        );
+    }
+    unsafe {
+        assert_cass_error_eq!(cass_future_error_code(fut.borrow()), CassError::CASS_OK);
+    }
+    unsafe { cass_future_free(fut) };
+}
+
```
Lorak-mmk: Is there a reason why this function reads the code twice? First in line 80, then in 92. It looks redundant, as does the whole second check of error code.

scylla-rust-wrapper/tests/integration/utils.rs
```diff
@@ -0,0 +1,226 @@
...
+        Consistency::Three => CassConsistency::CASS_CONSISTENCY_THREE,
+        Consistency::Quorum => CassConsistency::CASS_CONSISTENCY_QUORUM,
+        Consistency::All => CassConsistency::CASS_CONSISTENCY_ALL,
+        Consistency::LocalQuorum => CassConsistency::CASS_CONSISTENCY_LOCAL_QUORUM,
+        Consistency::EachQuorum => CassConsistency::CASS_CONSISTENCY_EACH_QUORUM,
+        Consistency::LocalOne => CassConsistency::CASS_CONSISTENCY_LOCAL_ONE,
+        Consistency::Serial => CassConsistency::CASS_CONSISTENCY_SERIAL,
+        Consistency::LocalSerial => CassConsistency::CASS_CONSISTENCY_LOCAL_SERIAL,
+    }
+}
```
Lorak-mmk: Why do you need this? I see we have `impl From<Consistency> for CassConsistency`.

scylla-rust-wrapper/tests/integration/utils.rs
```diff
@@ -0,0 +1,226 @@
...
+
+pub(crate) fn drop_metadata_queries_rules() -> impl IntoIterator<Item = RequestRule> {
+    [RequestRule(
+        Condition::ConnectionRegisteredAnyEvent.and(Condition::or(
+            Condition::RequestOpcode(RequestOpcode::Query),
+            Condition::RequestOpcode(RequestOpcode::Prepare),
+        )),
+        RequestReaction::forge().server_error(),
+    )]
+}
```
Lorak-mmk: Is there a reason for not addind EXECUTE here? Lack of it makes it more brittle - it won't work if you set it after driver connects (because it will have already prepared the queries by then).

Lorak-mmk (changes requested):

scylla-rust-wrapper/tests/integration/utils.rs
wprzytula: This is vestigial.

scylla-rust-wrapper/tests/integration/utils.rs
```diff
@@ -0,0 +1,207 @@
...
+        assert_eq!(
+            ___x,
+            $error,
+            "expected \"{}\", instead got \"{}\"",
+            ptr_to_cstr(cass_error_desc($error)).unwrap(),
+            ptr_to_cstr(cass_error_desc(___x)).unwrap()
+        );
+    }};
+}
+pub(crate) use assert_cass_error_eq;
```
Lorak-mmk: Why is this a macro? It should work perfectly fine being a function that takes two `CassError` arguments.

Lorak-mmk (approved):

PR #344, opened 2025-07-14
types: fix CQL TEXT and VARCHAR mapping

wprzytula:
Rust Driver unifies both VARCHAR and TEXT CQL types into `NativeType::Text`. CPP Driver, in accordance to the CQL protocol, has separate types for VARCHAR and TEXT. Even worse, Rust Driver even does not handle CQL TEXT correctly! It errors out on TEXT type...

As the DBs (Cassandra and ScyllaDB) seem to send the VARCHAR type in the protocol, we will assume that the `NativeType::Text` is actually a VARCHAR type.

The issue was discovered while running the examples, which were expecting VARCHAR type, but were getting TEXT type instead.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cass_types.rs
+++ b/scylla-rust-wrapper/src/cass_types.rs
@@ -270,7 +270,13 @@ fn native_type_to_cass_value_type(native_type: &NativeType) -> CassValueType {
         Float => CassValueType::CASS_VALUE_TYPE_FLOAT,
         Int => CassValueType::CASS_VALUE_TYPE_INT,
         BigInt => CassValueType::CASS_VALUE_TYPE_BIGINT,
-        Text => CassValueType::CASS_VALUE_TYPE_TEXT,
+        // Rust Driver unifies both VARCHAR and TEXT into NativeType::Text.
+        // CPP Driver, in accordance to the CQL protocol, has separate types for VARCHAR and TEXT.
+        // Even worse, Rust Driver even does not handle CQL TEXT correctly! It errors out on TEXT
+        // type...
+        // As the DBs (Cassandra and ScyllaDB) seem to send the VARCHAR type in the protocol,
+        // we will assume that the NativeType::Text is actually a VARCHAR type.
+        Text => CassValueType::CASS_VALUE_TYPE_VARCHAR,
         Timestamp => CassValueType::CASS_VALUE_TYPE_TIMESTAMP,
         Inet => CassValueType::CASS_VALUE_TYPE_INET,
         SmallInt => CassValueType::CASS_VALUE_TYPE_SMALL_INT,
--- a/tests/src/integration/tests/test_schema_metadata.cpp
+++ b/tests/src/integration/tests/test_schema_metadata.cpp
@@ -22,7 +22,7 @@
 
 class SchemaMetadataTest : public Integration {
 public:
-  SchemaMetadataTest() { 
+  SchemaMetadataTest() {
     is_schema_metadata_ = true;
     // Materialized views do not work with tablets.
     disable_tablets_ = true;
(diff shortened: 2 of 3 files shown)


scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -270,7 +270,13 @@ fn native_type_to_cass_value_type(native_type: &NativeType) -> CassValueType {
...
         BigInt => CassValueType::CASS_VALUE_TYPE_BIGINT,
-        Text => CassValueType::CASS_VALUE_TYPE_TEXT,
+        // Rust Driver unifies both VARCHAR and TEXT into NativeType::Text.
+        // CPP Driver, in accordance to the CQL protocol, has separate types for VARCHAR and TEXT.
+        // Even worse, Rust Driver even does not handle CQL TEXT correctly! It errors out on TEXT
+        // type...
+        // As the DBs (Cassandra and ScyllaDB) seem to send the VARCHAR type in the protocol,
+        // we will assume that the NativeType::Text is actually a VARCHAR type.
+        Text => CassValueType::CASS_VALUE_TYPE_VARCHAR,
         Timestamp => CassValueType::CASS_VALUE_TYPE_TIMESTAMP,
```
Lorak-mmk, line 280: "Even worse," suggests this is a bug. I don't think it is. The TEXT type might be implemented in the protocol, but if implementations of this protocol don't use it, then what is the point of supporting it? We wouldn't be able to even write a proper integration test for it in the driver.

Lorak-mmk (approved):

scylla-rust-wrapper/src/cass_types.rs
wprzytula: > but if implementations of this protocol don't use it, then what is the point of supporting it

If any implementation ever decides to use it, it will be hindered by the Rust Driver not supporting it, even though it is a well-defined part of the protocol. I consider such limitation an obstacle.

Lorak-mmk: This is a chicken-and-egg problem. Personally I think that the server-side is the one that should be implemented first, because:
- it is typically easier to implement driver-side changes, so it decreases iteration time
- it is simply how it was usually done (with e.g. metadata mark, tablets feedback, lwt extension, sharding extension).

wprzytula: I refuse to compare the examples that you provided (which are custom extensions to the protocol) with the core parts of the protocol, which `TEXT` belongs to.

Lorak-mmk: They are comparable in the sense that in both cases you have something that is unimplemented on both sides. Some side needs to be first. There is no point in having code in the driver that no server supports, or plans to support.

wprzytula:
@Lorak-mmk CI seems to be failing because Gor partly adjusted the testing code to Rust Driver's Varchar -> Text lossy conversion: 6b643af43839a0e2bc6f69ec3463e9cebebe5e75. I guess I need to revert those changes. Do you agree?

scylla-rust-wrapper/src/cass_types.rs
wprzytula: > They are comparable in the sense that in both cases you have something that is unimplemented on both sides. > Some side needs to be first.

OTOH, any side may assume that the other side supports the full core CQL protocol which it claims to support. The TEXT type belongs to the core part of CQL protocol v4, so any driver that claims support for it should correctly recognize TEXT.

> There is no point in having code in the driver that no server supports, or plans to support.

Perhaps this is right.

Lorak-mmk: > OTOH, any side may assume that the other side supports the full core CQL protocol which it claims to support.

True. Eh, I'm not sure what is the best solution here. One thing I do know is that this discussion of course does not influence this PR: it is correct regardless of the changes in Rust Driver.

Lorak-mmk:
sure - probably not all changes, but those regarding TEXT/VARCHAR, right?

wprzytula:
Reverted the patch that Gor wrote atop the IT framework in https://github.com/scylladb/cpp-rust-driver/commit/6b643af43839a0e2bc6f69ec3463e9cebebe5e75.

Lorak-mmk (approved):

wprzytula:
Rewritten the `test_schema_metadata.cpp` test that Gor wrote from scratch, so that it expects VARCHAR.

PR #345, opened 2025-07-15
A bunch of minor refactors - type safety, readability, robustness

wprzytula:
This is a bunch of small independent refactors that aim to increase type safety, readability and robustness.

## Pre-review checklist


scylla-rust-wrapper/src/statement.rs
```diff
@@ -426,7 +426,7 @@ pub unsafe extern "C" fn cass_statement_set_paging_state_token(
         return CassError::CASS_ERROR_LIB_NULL_VALUE;
     }
 
-    let paging_state_usize: usize = paging_state_size.try_into().unwrap();
+    let paging_state_usize: usize = paging_state_size as _;
     let paging_state_bytes =
         unsafe { slice::from_raw_parts(paging_state as *const u8, paging_state_usize) };
     statement_from_raw.paging_state = PagingState::new_from_raw_bytes(paging_state_bytes);
```
Lorak-mmk: Commit: "don't panic on u64 != usize wrt type size "

Why? It is better to panic than to silently produce incorrect value.

scylla-rust-wrapper/src/statement.rs
```diff
@@ -422,7 +422,6 @@ pub unsafe extern "C" fn cass_statement_set_paging_state_token(
     };
 
     if paging_state.is_null() {
-        statement_from_raw.paging_state = PagingState::start();
         return CassError::CASS_ERROR_LIB_NULL_VALUE;
     }
```
Lorak-mmk: Commit: "don't unset paging state if null ptr passed "

WDYM by "Note: CPP Driver does not check the passed pointer for null at all."? How will cpp-driver behave if null is passed?

scylla-rust-wrapper/src/cass_types.rs
```diff
@@ -898,11 +898,11 @@ pub unsafe extern "C" fn cass_data_type_add_sub_value_type_by_name_n(
...
-        CassBatchType::CASS_BATCH_TYPE_LOGGED => Some(BatchType::Logged),
-        CassBatchType::CASS_BATCH_TYPE_UNLOGGED => Some(BatchType::Unlogged),
-        CassBatchType::CASS_BATCH_TYPE_COUNTER => Some(BatchType::Counter),
-        _ => None,
+        CassBatchType::CASS_BATCH_TYPE_LOGGED => Ok(BatchType::Logged),
+        CassBatchType::CASS_BATCH_TYPE_UNLOGGED => Ok(BatchType::Unlogged),
+        CassBatchType::CASS_BATCH_TYPE_COUNTER => Ok(BatchType::Counter),
+        _ => Err(()),
     }
 }
```
Lorak-mmk, line 908: Nit: typically the original value is returned in the error in such cases, right? In this case it does not matter because the type is Copy (I think).

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/statement.rs
wprzytula: Hmm, right.

scylla-rust-wrapper/src/statement.rs
wprzytula: It calls a `String` constructor:

```c++
CassError cass_statement_set_paging_state_token(CassStatement* statement, const char* paging_state,
                                                size_t paging_state_size) {
  statement->set_paging_state(String(paging_state, paging_state_size));
  return CASS_OK;
}
```

scylla-rust-wrapper/src/cass_types.rs
wprzytula: Perhaps typically in other libraries yes, but in the wrapper it's common to see `type Error = ()`.

scylla-rust-wrapper/src/statement.rs
Lorak-mmk: Which will do what on null?

wprzytula: `String` is defined as a type alias:
`typedef std::basic_string<char, std::char_traits<char>, internal::Allocator<char> > String;`

So we need to look for a corresponding constructor...
https://en.cppreference.com/w/cpp/string/basic_string/basic_string

I believe it corresponds to:
```c++
basic_string( const CharT* s, size_type count, const Allocator& alloc = Allocator());
```
which is described as follows:
```
6) Constructs a string with the contents of the range [s, s + count).
If [s, s + count) is not a valid range, the behavior is undefined.
```

So, I expect this to be UB in CPP Driver.

Lorak-mmk: Ok. So what is the proper way to unset paging state in cpp-driver?

wprzytula: There's no such way as far as I can see.

wprzytula: Moreover, page size and paging state are independent:

```c++

  if (page_size() > 0) {
    flags |= CASS_QUERY_FLAG_PAGE_SIZE;
  }

  if (!paging_state().empty()) {
    flags |= CASS_QUERY_FLAG_PAGING_STATE;
  }
```

So even with `page_size` set to `-1` (which means no paging), if the paging state is set, the whole remaining response, beginning with the given paging state, will come in one shot.

Lorak-mmk: The snippet above suggests that setting paging state to empty string unsets it (the flag won't be sent).
It seems reasonable to me to allow null to make unsetting easier

wprzytula:
Altered the commit with changes to `cass_statement_set_paging_state_token` so that null strings and empty strings are both now correct arguments that unset the paging state.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -64,7 +64,7 @@ impl TryFrom<CassLogLevel> for Level {
             CassLogLevel::CASS_LOG_WARN => Level::WARN,
             CassLogLevel::CASS_LOG_ERROR => Level::ERROR,
             CassLogLevel::CASS_LOG_CRITICAL => Level::ERROR,
-            _ => return Err("Error while converting CassLogLevel to Level".to_string()),
```
Lorak-mmk, line 67: Did you check callers of this to verify if they should be adjusted as well? I'd expect this change to be accompanied by changes to messages in callers.

wprzytula: Yes, there was only one caller (detected by temporary deletion of the conversion impl).
The caller only calls `unwrap_or()`.

Lorak-mmk (approved):

PR #352, opened 2025-07-28
errors: Make Rust errors conversion to `CassError` more appropriate

### LINKED ISSUE #133: Make `QueryError` -> `CassError` conversion more accurate

wprzytula:
Fixes: #133

This PR:
1. Merges `execution_error.rs` into `cass_error.rs`, as both contain related functionality that should live together.
2. Adjusts conversion of multiple Rust Driver's error variants to `CassError`, by choosing a better-fitting `CassError` variant. I tried to provide clear explanation why I believe the new variant is a better fit.

## Pre-review checklist


scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -147,13 +147,13 @@ impl ToCassError for DbError {
...
+                CassError::CASS_ERROR_LIB_MESSAGE_ENCODE
+            }
+            BadQuery::PartitionKeyExtraction => CassError::CASS_ERROR_LIB_MESSAGE_ENCODE,
             BadQuery::SerializationError(e) => e.to_cass_error(),
-            BadQuery::TooManyQueriesInBatchStatement(_) => CassError::CASS_ERROR_LAST_ENTRY,
+            BadQuery::TooManyQueriesInBatchStatement(_) => CassError::CASS_ERROR_LIB_MESSAGE_ENCODE,
             // BadQuery is non_exhaustive
-            // For now, since all other variants return LAST_ENTRY,
-            // let's do it here as well.
             _ => CassError::CASS_ERROR_LAST_ENTRY,
```
Lorak-mmk, line 157: Here we need to handle non-exhaustiveness of error types from the driver.
Maybe we could utilize unstable `non_exhaustive_omitted_patterns_lint` from rustc?

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -257,7 +257,7 @@ impl ToCassError for SerializationError {
             // It means that our custom `UnknownNamedParameterError` was returned.
             CassError::CASS_ERROR_LIB_NAME_DOES_NOT_EXIST
         } else {
-            CassError::CASS_ERROR_LAST_ENTRY
+            CassError::CASS_ERROR_LIB_INVALID_DATA
         }
     }
```
Lorak-mmk, line 262: Commit: "cass_error: SerializationError -> INVALID_DATA "

When is LIB_INVALID_DATA appropriate vs LIB_MESSAGE_ENCODE?

wprzytula: My vague understanding of it is that when the particular CQL value is of bad type or corrupted, then `INVALID_DATA` is appropriate. `MESSAGE_ENCODE` is more about CQL protocol limitations, such as too big number of values, too many statements in a batch etc.

scylla-rust-wrapper/src/cass_error.rs
wprzytula: If you're suggesting a switch to a nightly compiler, I'm against. From my experience with TockOS, working with nightly is a pain which I'd definitely prefer to avoid.

Lorak-mmk: > From my experience with TockOS, working with nightly is a pain

Isn't it more precise to say "working with specific old version of nightly is a pain"?

wprzytula: I don't see how working with newer versions of nightly is not a pain.

Lorak-mmk: Could you explain why working with nightly is a pain?

wprzytula: Consider the following:
- nightly versions are often (_nightly_) changing;
- you need to keep CI in sync with your local version, as well as local version of all contributors;
- Is your idea to pin a specific nightly version, like a specific night that all contributors and CI shall use, and bump it once upon a time?
- Isn't nightly more prone to bugs and unstable behaviour? Can we afford this in a production-critical software?
- Are nightly features documented well-enough to rely on them? Won't it be more pain than gain when they change and we need to adjust the code frequently?

Lorak-mmk (approved):

PR #353, opened 2025-07-29
add API to specify metadata request timeout

### LINKED ISSUE #172: Decouple schema fetch queries from server-side timeouts
Rust driver part should be completed under https://github.com/scylladb/scylla-rust-driver/issues/1052

Here we need only expose configuration option.

Migrated to Jira: [DRIVER-196](https://scylladb.atlassian.net/browse/DRIVER-196)

[DRIVER-196]: https://scylladb.atlassian.net/browse/DRIVER-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

wprzytula:
This adds a new API function to the driver: `cass_cluster_set_metadata_request_serverside_timeout`. This function allows users to set a server-side timeout for metadata queries, which is particularly useful for large schemas where the default timeout may be too tight. The default timeout is set to 2 seconds (as in Rust Driver), and users can pass 0 to use the cluster's default timeout. This simply exposes the existing Rust Driver's API.

Fixes: #172

## Pre-review checklist

Closes [DRIVER-196](https://scylladb.atlassian.net/browse/DRIVER-196)

[DRIVER-196]: https://scylladb.atlassian.net/browse/DRIVER-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -3124,6 +3124,27 @@ CASS_EXPORT void
 cass_cluster_set_compression(CassCluster* cluster,
                              CassCompressionType compression_type);
 
+/**
+ * Sets the server-side timeout for metadata queries.
+ *
+ * It means that all metadata queries will be set the given timeout
+ * no matter what timeout is set as a cluster default.
+ * This prevents timeouts of schema queries when the schema is large
+ * and the default timeout is configured as tight.
+ *
+ * <b>Default:</b> 2 seconds.
+ *
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] timeout_ms Request timeout in milliseconds. Pass 0 to use cluster default timeout.
+ */
+CASS_EXPORT void
+cass_cluster_set_metadata_request_serverside_timeout(CassCluster* cluster,
+                             cass_duration_t timeout_ms);
+
+
+
 /***********************************************************************************
  *
  * Session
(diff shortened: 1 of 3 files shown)


include/cassandra.h
```diff
@@ -3097,6 +3097,27 @@ CASS_EXPORT void
...
+ * @public @memberof CassCluster
+ *
+ * @param[in] cluster
+ * @param[in] timeout_ms Request timeout in milliseconds. Pass 0 to use cluster default timeout.
+ */
+CASS_EXPORT void
+cass_cluster_set_metadata_request_serverside_timeout(CassCluster* cluster,
+                             cass_duration_t timeout_ms);
+
+
```
Lorak-mmk, line 3146: Is there a dedicated setting for metadata client-side timeout?
If so, can we make it that:
- Default for server-side timeout is client-side timeout (perhaps with a slight reduction)
- There is a warning if users sets server-side higher than client-side?

If not, what client-side timeout is used? Can the above points apply to it?

wprzytula: > Is there a dedicated setting for metadata client-side timeout?

No.

wprzytula: > If not, what client-side timeout is used?

As with the other settings configured on execution profiles and on statements directly, the hierarchy is as follows:

statement > execution profile > cluster

wprzytula: > Can the above points apply to it?

I'm not sure I understand.

Lorak-mmk (approved):

PR #354, opened 2025-07-31
implement `cass_cluster_set_num_threads_io`

### LINKED ISSUE #259: Implement cass_cluster_set_num_threads_io

wprzytula:
Fixes: #259

# What's done

This PR does the necessary refactors and implements `cass_cluster_set_num_threads_io`.

## Semantics

The semantics of the function is slightly different than in the CPP Driver:
- in CPP Driver, there are always `n+1` worker threads, where `n` is the argument passed to `cass_cluster_set_num_threads_io`. One worker thread serves the control connection and other background Session tasks, whereas the remaining `n` worker threads serve requests.
- in CPP-Rust Driver, the `n` argument is directly passed to `worker_threads()` config of the tokio `multi_thread` runtime. Those threads execute all tasks that are spawned in the wrapper and in the Rust Driver itself.

### 0 worker threads
This is an extension compared to the CPP Driver. 0 is handled specially: it uses `current_thread` tokio executor undeneath. Limitations and pitfalls of it are described in the function's documentation in `cassandra.h`.

## `CassFuture` refactors

In order to support holding the runtime in each `CassFuture` more cleanly, the `CassFuture` is split into two variants:
- `Resolvable`, which is like the `CassFuture` so far, but it also holds `Arc<Runtime>` it needs to use to poll the future,
- `Immediate`, which is created as an instantly ready future and thus does not need to store a runtime.

As a bonus, I cleaned up the logic in `CassFuture` a bit and provided more documentation and comments.

## Testing

I ran all examples with 0 worker threads. All but `callbacks` worked. `callbacks`, as expected, hung.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -7,6 +7,7 @@ use crate::load_balancing::{
     CassHostFilter, DcRestriction, LoadBalancingConfig, LoadBalancingKind,
 };
 use crate::retry_policy::CassRetryPolicy;
+use crate::runtime::RUNTIMES;
 use crate::ssl::CassSsl;
 use crate::timestamp_generator::CassTimestampGen;
 use crate::types::*;
@@ -82,7 +83,11 @@ const DRIVER_NAME: &str = "ScyllaDB Cpp-Rust Driver";
 const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 pub struct CassCluster {
-    runtime: Arc<tokio::runtime::Runtime>,
+    /// Number of threads in the tokio runtime thread pool.
+    ///
+    /// Specified with `cass_cluster_set_num_threads_io`.
+    /// If not set, the default tokio runtime is used.
+    num_threads_io: Option<usize>,
 
     session_builder: SessionBuilder,
     default_execution_profile_builder: ExecutionProfileBuilder,
@@ -101,8 +106,20 @@ pub struct CassCluster {
 }
 
 impl CassCluster {
-    pub(crate) fn get_runtime(&self) -> &Arc<tokio::runtime::Runtime> {
-        &self.runtime
+    /// Gets the runtime that has been set for the cluster.
+    /// If no runtime has been set yet, it creates a default runtime
+    /// and makes it cached in the global `Runtimes` instance.
+    pub(crate) fn get_runtime(&self) -> Arc<tokio::runtime::Runtime> {
+        let mut runtimes = RUNTIMES.lock().unwrap();
+
+        if let Some(num_threads_io) = self.num_threads_io {
+            // If the number of threads is set, we create a runtime with that number of threads.
+            runtimes.n_thread_runtime(num_threads_io)
+        } else {
+            // Otherwise, we use the default runtime.
+            runtimes.default_runtime()
+        }
+        .unwrap_or_else(|err| panic!("Failed to create an async runtime: {err}"))
     }
 
     pub(crate) fn execution_profile_map(&self) -> &HashMap<ExecProfileName, CassExecProfile> {
@@ -184,12 +201,6 @@ impl CassCluster {
 
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster, CMut> {
-    let Ok(default_runtime) = tokio::runtime::Runtime::new()
-        .inspect_err(|e| tracing::error!("Failed to create async runtime: {}", e))
-    else {
-        return CassPtr::null_mut();
-    };
-
     let default_execution_profile_builder = ExecutionProfileBuilder::default()
         .consistency(DEFAULT_CONSISTENCY)
         .serial_consistency(DEFAULT_SERIAL_CONSISTENCY)
@@ -321,7 +332,7 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
     };
 
     BoxFFI::into_ptr(Box::new(CassCluster {
-        runtime: Arc::new(default_runtime),
+        num_threads_io: None,
 
         session_builder: default_session_builder,
         port: 9042,
@@ -1554,25 +1565,7 @@ pub unsafe extern "C" fn cass_cluster_set_num_threads_io(
         return CassError::CASS_ERROR_LIB_BAD_PARAMS;
     };
 
-    let runtime_res = match num_threads {
-        0 => tokio::runtime::Builder::new_current_thread()
-            .enable_all()
-            .build(),
-        n => tokio::runtime::Builder::new_multi_thread()
-            .worker_threads(n as usize)
-            .enable_all()
-            .build(),
-    };
-
-    let runtime = match runtime_res {
-        Ok(runtime) => runtime,
-        Err(err) => {
-            tracing::error!("Failed to create async runtime: {}", err);
-            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
-        }
-    };
-
-    cluster.runtime = Arc::new(runtime);
+    cluster.num_threads_io = Some(num_threads as usize);
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -83,6 +83,8 @@ const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[derive(Clone)]
 pub struct CassCluster {
+    runtime: Arc<tokio::runtime::Runtime>,
+
     session_builder: SessionBuilder,
     default_execution_profile_builder: ExecutionProfileBuilder,
     execution_profile_map: HashMap<ExecProfileName, CassExecProfile>,
@@ -100,6 +102,10 @@ pub struct CassCluster {
 }
 
 impl CassCluster {
+    pub(crate) fn get_runtime(&self) -> &Arc<tokio::runtime::Runtime> {
+        &self.runtime
+    }
+
     pub(crate) fn execution_profile_map(&self) -> &HashMap<ExecProfileName, CassExecProfile> {
         &self.execution_profile_map
     }
@@ -179,6 +185,12 @@ impl CassCluster {
 
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster, CMut> {
+    let Ok(default_runtime) = tokio::runtime::Runtime::new()
+        .inspect_err(|e| tracing::error!("Failed to create async runtime: {}", e))
+    else {
+        return CassPtr::null_mut();
+    };
+
     let default_execution_profile_builder = ExecutionProfileBuilder::default()
         .consistency(DEFAULT_CONSISTENCY)
         .serial_consistency(DEFAULT_SERIAL_CONSISTENCY)
@@ -310,6 +322,8 @@ pub unsafe extern "C" fn cass_cluster_new() -> CassOwnedExclusivePtr<CassCluster
     };
 
     BoxFFI::into_ptr(Box::new(CassCluster {
+        runtime: Arc::new(default_runtime),
+
         session_builder: default_session_builder,
         port: 9042,
         contact_points: Vec::new(),
@@ -1527,6 +1541,39 @@ pub unsafe extern "C" fn cass_cluster_set_execution_profile_n(
     };
 
     cluster.execution_profile_map.insert(name, profile);
+
+    CassError::CASS_OK
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cass_cluster_set_num_threads_io(
+    cluster: CassBorrowedExclusivePtr<CassCluster, CMut>,
+    num_threads: cass_uint32_t,
+) -> CassError {
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster) else {
+        tracing::error!("Provided null cluster pointer to cass_cluster_set_num_threads_io!");
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
+    let runtime_res = match num_threads {
+        0 => tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build(),
+        n => tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(n as usize)
+            .enable_all()
+            .build(),
+    };
+
+    let runtime = match runtime_res {
+        Ok(runtime) => runtime,
+        Err(err) => {
+            tracing::error!("Failed to create async runtime: {}", err);
+            return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+        }
+    };
+
+    cluster.runtime = Arc::new(runtime);
 
     CassError::CASS_OK
 }
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -81,7 +81,6 @@ const DEFAULT_SHARD_AWARE_LOCAL_PORT_RANGE: ShardAwarePortRange =
 const DRIVER_NAME: &str = "ScyllaDB Cpp-Rust Driver";
 const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-#[derive(Clone)]
 pub struct CassCluster {
     runtime: Arc<tokio::runtime::Runtime>,
 
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -9,6 +9,7 @@ use futures::future;
 use std::future::Future;
 use std::mem;
 use std::os::raw::c_void;
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Condvar, Mutex, OnceLock};
 use tokio::task::JoinHandle;
 use tokio::time::Duration;
@@ -57,7 +58,18 @@ struct CassFutureState {
     join_handle: Option<JoinHandle<()>>,
 }
 
-pub struct CassFuture {
+enum FutureKind {
+    /// Future that must be resolved by the tokio runtime.
+    Resolvable { fut: ResolvableFuture },
+
+    /// Future that is immediately ready with the result.
+    Immediate {
+        res: CassFutureResult,
+        callback_set: AtomicBool,
+    },
+}
+
+struct ResolvableFuture {
     /// Mutable state of the future that requires synchronized exclusive access
     /// in order to ensure thread safety of the future execution.
     state: Mutex<CassFutureState>,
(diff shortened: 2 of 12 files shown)


scylla-rust-wrapper/src/future.rs
```diff
@@ -437,15 +439,12 @@ pub unsafe extern "C" fn cass_future_error_message(
...
-                Ok(CassResultValue::QueryError(err)) => err.msg(),
-                Err((_, s)) => s.msg(),
-                _ => "".to_string(),
-            });
+    future.with_waited_result(|result: &CassFutureResult| {
+        let msg = future.err_string.get_or_init(|| match result {
+            Ok(CassResultValue::QueryError(err)) => err.msg(),
+            Err((_, s)) => s.msg(),
+            _ => "".to_string(),
+        });
```
Lorak-mmk: We could avoid the allocations in Err and _ cases, right?
I think it should be as easy as inverting nesting of get_or_init and match:
```rust
    let msg = match value {
        Ok(CassResultValue::QueryError(err)) => future.err_string.get_or_init(|| err.msg()),
        Err((_, s)) => s,
        _ => "",
    };

    unsafe { write_str_to_c(msg, message, message_length) };
```

but I did not carefully read this code, so it may contain bugs.

Lorak-mmk: It is interesting that according to rust analyzer, `msg` is `&'static str` in my example. I think it should be `&str` (so with a lifetime of the future) - but it should still be a correct lifetime here.

scylla-rust-wrapper/src/integration_testing.rs
```diff
@@ -280,14 +280,27 @@ pub unsafe extern "C" fn testing_future_get_attempted_hosts(
...
 #[test]
 fn test_future_get_attempted_hosts() {
     use scylla::observability::history::HistoryListener as _;
 
+    let runtime = runtime_for_test();
+
     let listener = Arc::new(RecordingHistoryListener::new());
-    let future = CassFuture::new_from_future(std::future::pending(), Some(listener.clone()));
+    let future =
+        CassFuture::new_from_future(runtime, std::future::pending(), Some(listener.clone()));
```
Lorak-mmk, line 303: Commit: `future: ditch closure-based awaiting `

It looks like this change was mistakenly put in this commit. Signature of `new_from_future` did not change, so surely this commit does not compile now.

Lorak-mmk: AI could possibly be a great tool to catch issues like this, if only someone thought to make it review commits instead of all changes lumped together.

Well, Github doesn't even have a proper UI for per-commit review, so they obviously don't care about such unimportant things like good commit history

scylla-rust-wrapper/src/future.rs
```diff
@@ -230,25 +275,38 @@ impl CassFuture {
...
-    fn with_waited_state_timed<T>(
-        &self,
-        f: impl FnOnce(&mut CassFutureState) -> T,
+    fn waited_result_timed<'s>(
+        &'s self,
         timeout_duration: Duration,
-    ) -> Result<T, FutureError> {
-        let mut guard = self.state.lock().unwrap();
+    ) -> Result<&'s CassFutureResult, FutureError> {
+        let resolvable_fut = match self.kind {
```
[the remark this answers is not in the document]
Lorak-mmk: I agree with this - such method would help the reader parse all the places where you panic on  one kind.

scylla-rust-wrapper/src/future.rs
```diff
@@ -396,7 +432,17 @@ pub unsafe extern "C" fn cass_future_set_callback(
...
+        FutureKind::Resolvable(ref resolvable) => {
+            // Safety: `callback` is a valid pointer to a function that matches the signature.
+            unsafe { resolvable.set_callback(future_raw.borrow(), callback, data) }
+        }
+        FutureKind::Immediate(_) => {
+            let bound_cb = BoundCallback { cb: callback, data };
+            bound_cb.invoke(future_raw.borrow());
+            CassError::CASS_OK
+        }
+    }
```
Lorak-mmk: That doesn't look correct. Futures had the semantics, both here and in cpp driver, that callback is called at least one per future.
Now you are allowing multiple calls in case of Immediate.

scylla-rust-wrapper/src/session.rs
```diff
@@ -279,13 +279,18 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
-        let Some(connected_session) = session_guard.connected.as_ref() else {
-            return Err((
-                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                "Session is not connected".msg(),
-            ));
-        };
+        let connected_session = session_guard
+            .connected
+            .as_ref()
+            .expect("This should have been handled synchronously!");
```
Lorak-mmk, line 319: Why do you need to recreate this var? The closure is move, so `connected_session` created with `let Some` before should be enough.

scylla-rust-wrapper/src/session.rs
```diff
@@ -366,15 +378,13 @@ pub unsafe extern "C" fn cass_session_execute(
...
-            return Err((
-                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                "Session is not connected".msg(),
-            ));
-        };
-        let session = &cass_connected_session.session;
+        let connected_session = session_guard
+            .connected
+            .as_ref()
+            .expect("This should have been handled synchronously!");
```
Lorak-mmk, line 413: ditto

scylla-rust-wrapper/src/session.rs
```diff
@@ -511,12 +528,10 @@ pub unsafe extern "C" fn cass_session_prepare_from_existing(
...
-            let Some(connected_session) = session_guard.connected.as_ref() else {
-                return Err((
-                    CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                    "Session is not connected".msg(),
-                ));
-            };
+            let connected_session = session_guard
+                .connected
+                .as_ref()
+                .expect("This should have been handled synchronously!");
```
Lorak-mmk, line 567: ditto

scylla-rust-wrapper/src/session.rs
```diff
@@ -566,15 +581,20 @@ pub unsafe extern "C" fn cass_session_prepare_n(
...
-        let Some(connected_session) = session_guard.connected.as_ref() else {
-            return Err((
-                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                "Session is not connected".msg(),
-            ));
-        };
+        let connected_session = session_guard
+            .connected
+            .as_ref()
+            .expect("This should have been handled synchronously!");
```
Lorak-mmk, line 632: ditto

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -83,6 +83,8 @@ const DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[derive(Clone)]
 pub struct CassCluster {
+    runtime: Arc<tokio::runtime::Runtime>,
+
```
Lorak-mmk: So now we'll always be using per-cluster runtime? No way to use the global one?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1531,6 +1545,39 @@ pub unsafe extern "C" fn cass_cluster_set_execution_profile_n(
...
+    let runtime_res = match num_threads {
+        0 => tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build(),
+        n => tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(n as usize)
+            .enable_all()
+            .build(),
+    };
```
Lorak-mmk: Its a bit wasteful that the flow is:
- Create default runtime (which involves spawning threads) in cluster new
- Drop it here and create another one

Can't we build runtime in connect?

scylla-rust-wrapper/src/future.rs
```diff
@@ -646,11 +661,13 @@ mod tests {
     #[ntest::timeout(100)]
     fn cass_future_thread_safety() {
         const ERROR_MSG: &str = "NOBODY EXPECTED SPANISH INQUISITION";
+        let runtime = runtime_for_test();
         let fut = async {
             tokio::time::sleep(Duration::from_millis(10)).await;
             Err((CassError::CASS_OK, ERROR_MSG.into()))
```
Lorak-mmk, line 706: I wanted to suggest using `tokio::test` and somehow getting current runtime object, but I see there is not such functionality in Tokio. Surprising.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/session.rs
wprzytula: `connected_session` borrows from `session_guard`, which lives on the stack, not in the `move` block. Therefore it's not possible to move it into.

scylla-rust-wrapper/src/cluster.rs
wprzytula: Yes. Why would we want to have and use a global runtime?

scylla-rust-wrapper/src/cluster.rs
wprzytula: > Can't we build runtime in connect?

This is a good idea, which however requires to lazily build the runtime and store it in the cluster upon call to `cass_session_connect` - because we don't want to build a runtime for every session separately. It should belong to the cluster and be shared with sessions.

This would unfortunately not be in line with C API's type hint conveyed by the `CassCluster const*` pointer in the signature of `cass_session_connect`. In order to allow concurrent connecting multiple sessions, we would have to use interior mutability. No problem with that, we can always employ `ArcSwap` or simply a `Mutex` to guard the runtime inside `CassCluster`.
Actually, `Mutex` looks like the best choice, because we want to block all racing threads but one so that only one attempts to create a runtime. An alternative would be to have an `ArcSwapOption` that operates with `compare_and_swap()` with threads racing to set their created runtime, but this would be wasteful.

scylla-rust-wrapper/src/future.rs
wprzytula: Yep, rust-analyzer exhibits a bug here.

scylla-rust-wrapper/src/future.rs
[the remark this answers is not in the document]
wprzytula: There are exactly two places in the code when I believe such method increases readability instead of decreasing it. They both are in the `CassFuture::new_from_future` constructor. Do you still believe it is worth introducing this?

scylla-rust-wrapper/src/future.rs
wprzytula: You mean "at most once per future"?

wprzytula: I haven't checked cpp-driver yet, but cpp-rust-driver on `master` allows executing a callback indefinitely many times. Once a future has completed, each call to `cass_future_set_callback` on that future will synchronously execute the callback being set.

scylla-rust-wrapper/src/session.rs
Lorak-mmk: Ahh, I see it now. It is quite sad there is no better solution for this pattern :(

scylla-rust-wrapper/src/future.rs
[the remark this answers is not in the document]
Lorak-mmk: For some reason I thought there were more. In that case we can skip this.

scylla-rust-wrapper/src/session.rs
wprzytula: Yep, unfortunately.

scylla-rust-wrapper/src/future.rs
Lorak-mmk: > You mean "at most once per future"?

Yes. Eh, not the first mistake I see in my review comments from yesterday. Sorry for that!

wprzytula: So I don't see how the previous semantics differs from the current one.

Lorak-mmk: You're right, cpp-rust-driver does allow multiple callbacks after finishing. The code is a bit tricky to understand - it does check if callback is already set but:
- If it is not, and future is already resolved, it will be immediately called and NOT set on the future.
- If it is, then future resolves, and takes the callback out before calling it, making it unset.

Now what abut cpp-driver?
```cpp
bool Future::set_callback(Future::Callback callback, void* data) {
  ScopedMutex lock(&mutex_);
  if (callback_) {
    return false; // Callback is already set
  }
  callback_ = callback;
  data_ = data;
  if (is_set_) {
    // Run the callback if the future is already set
    lock.unlock();
    callback(CassFuture::to(this), data);
  }
  return true;
}
```

This always sets the callback - not only when future is unresolved. Also `Future::internal_set` which I think is called when future is resolved:
```cpp
void Future::internal_set(ScopedMutex& lock) {
  is_set_ = true;
  if (callback_) {
    Callback callback = callback_;
    void* data = data_;
    lock.unlock();
    callback(CassFuture::to(this), data);
    lock.lock();
  }
  // Broadcast after we've run the callback so that threads waiting
  // on this future see the side effects of the callback.
  uv_cond_broadcast(&cond_);
}
```

it does not unset the callback. 
So it looks like cpp-driver does have "at most once" semantics, while we do not. We should probably align with them in that aspect, right?

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: For apps that use > 1 session (perhaps with different configs) it may make sense to share the runtime instead of creating hundreds of threads.

wprzytula: Mhm. For such niche use cases, we can advise creating only a single `CassCluster` and mutating it accordingly for each session.

scylla-rust-wrapper/src/future.rs
wprzytula: Good observation. Yes, I agree that we should align. I'll start by introducing a fix to the logic _before_ my changes, and then modify my changes to keep the semantics unchanged.

scylla-rust-wrapper/src/cluster.rs
wprzytula: The same problem arises if someone creates multiple `CassCluster`, each followed by a call to `cass_cluster_set_num_threads_io` **with the same argument**. Then, runtimes are multiplicated even though they are identical.

A full solution to that problem would be IMO to have a global `Map<usize, Arc<Runtime>>` that would prevent `CassCluster`s from creating redundant runtimes. Will you approve such a solution?

wprzytula: Aside from the Map, there would be a `OnceLock<Runtime>` used to store the lazy-initialized default runtime.

Lorak-mmk: > Will you approve such a solution?

I will even approve the current solution if you want. The issue I mentioned is probably niche - we can just open an issue to discuss it in the future.

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: As with the comment about shared runtime - I'm fine with keeping the PR as-is, and maybe discussing this at a later time.

scylla-rust-wrapper/src/cluster.rs
wprzytula: I'm quite convinced that the Map solution is a good one. Let me try it out.

wprzytula: I implemented this solution.

scylla-rust-wrapper/src/cluster.rs
wprzytula: I implemented this using `Mutex`-guarded global runtime cache.

scylla-rust-wrapper/src/future.rs
wprzytula: Done.

scylla-rust-wrapper/src/session.rs
```diff
@@ -602,7 +658,7 @@ pub unsafe extern "C" fn cass_session_free(session_raw: CassOwnedSharedPtr<CassS
     };
 
     let close_fut = CassConnectedSession::close_fut(session_opt);
-    close_fut.with_waited_result(|_| ());
+    close_fut.waited_result();
```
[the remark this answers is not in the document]
wprzytula: If `close_fut` fails, this means that the session was already closed. This isn't anything to warn about.

scylla-rust-wrapper/src/runtime.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Weak pointers are used to make runtimes dropped once all CassCluster instances
+    // that reference them are freed.
+    default_runtime: Option<Weak<Runtime>>,
+    // This is Option to allow creating a static instance of Runtimes.
+    // (`HashMap::new` is not `const`).
+    n_thread_runtimes: Option<HashMap<usize, Weak<Runtime>>>,
+}
+
+pub(crate) static RUNTIMES: std::sync::Mutex<Runtimes> = {
+    std::sync::Mutex::new(Runtimes {
```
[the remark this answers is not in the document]
wprzytula: Accessing this Mutex will only happen upon calls to `cass_session_connect`, which are rare. Not an issue.

scylla-rust-wrapper/src/future.rs
```diff
@@ -96,66 +131,79 @@ impl CassFuture {
...
+            },
         });
         let cass_fut_clone = Arc::clone(&cass_fut);
-        let join_handle = RUNTIME.spawn(async move {
+        let join_handle = runtime.spawn(async move {
+            let resolvable_fut = match cass_fut_clone.kind {
+                FutureKind::Resolvable {
+                    fut: ref resolvable,
+                } => resolvable,
+                _ => unreachable!("CassFuture has been created as Resolvable"),
```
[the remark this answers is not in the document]
wprzytula: Unfortunately, I couldn't come up with anything better.

scylla-rust-wrapper/src/future.rs
```diff
@@ -96,66 +131,79 @@ impl CassFuture {
...
-            cass_fut_clone.wait_for_value.notify_all();
+            resolvable_fut.wait_for_value.notify_all();
         });
         {
-            let mut lock = cass_fut.state.lock().unwrap();
+            let resolvable_fut = match cass_fut.kind {
+                FutureKind::Resolvable {
+                    fut: ref resolvable,
+                } => resolvable,
+                _ => unreachable!("CassFuture has been created as Resolvable"),
```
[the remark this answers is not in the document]
wprzytula: ditto

scylla-rust-wrapper/src/session.rs
```diff
@@ -279,18 +303,24 @@ pub unsafe extern "C" fn cass_session_execute_batch(
...
-        if session_guard.connected.is_none() {
-            return Err((
-                CassError::CASS_ERROR_LIB_NO_HOSTS_AVAILABLE,
-                "Session is not connected".msg(),
-            ));
-        }
+        let connected_session = session_guard
+            .connected
+            .as_ref()
+            .expect("This should have been handled synchronously!");
```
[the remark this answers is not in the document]
wprzytula: This message is like `unreachable!()` - user should never see this.

scylla-rust-wrapper/src/session.rs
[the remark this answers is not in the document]
wprzytula: Answered before - if `close_fut` errors out, this means that the session was already closed. So no need to warn here.

Lorak-mmk (changes requested):
As discussed offline, please fix commit message and comments in the commit that introduces runtime cache.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -100,6 +106,22 @@ pub struct CassCluster {
...
+        let mut runtimes = RUNTIMES.lock().unwrap();
+
+        if let Some(num_threads_io) = self.num_threads_io {
+            // If the number of threads is set, we create a runtime with that number of threads.
+            runtimes.n_thread_runtime(num_threads_io)
+        } else {
+            // Otherwise, we use the default runtime.
+            runtimes.default_runtime()
+        }
+        .expect("Failed to create an async runtime")
```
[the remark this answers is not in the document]
wprzytula: This is actually a fair point. Will fix.

Lorak-mmk (approved):

PR #363, opened 2025-09-09
Fix panic upon runtime drop from async context

### LINKED ISSUE #362: Runtime may be dropped from an async context, causing a panic
A [CI run](https://github.com/scylladb/cpp-rust-driver/actions/runs/17442213987/job/49528268513?pr=361) revealed a bug in the tokio runtime management: **`Runtime` might be dropped from an async context**, which is forbidden (as it's a blocking operation), **so the worker thread panics**.

Investigation showed that the problem manifests when a tokio task which drives a future stored in `CassFuture` becomes the last shared owner of the tokio `Runtime`, which is shared behind an `Arc`.

In order to provoke such situation, I crafted a reproducer. Its core idea is to drop all other handles to the `Runtime`, i.e., all entities that own `Arc<Runtime>` on the main thread. This way the tokio task becomes the last owner, so it attempts to drop the `Runtime` and causes the panic. 

The bug can be reproduced locally with high probability (more than half of the runs) with the following code:

```rust
    #[tokio::test]
    async fn runtime_is_never_dropped_in_async_context() {
        setup_tracing();

        test_with_one_proxy(
            runtime_is_never_dropped_in_async_context_do,
            mock_init_rules(),
        )
        .with_current_subscriber()
        .await;
    }

    fn runtime_is_never_dropped_in_async_context_do(
        node_addr: SocketAddr,
        proxy: RunningProxy,
    ) -> RunningProxy {
        unsafe {
            let ip = node_addr.ip().to_string();
            let mut cluster_raw = cass_cluster_new();
            let (c_ip, c_ip_len) = str_to_c_str_n(ip.as_str());

            assert_cass_error_eq!(
                cass_cluster_set_contact_points_n(cluster_raw.borrow_mut(), c_ip, c_ip_len),
                CassError::CASS_OK
            );
            let session_raw = cass_session_new();
            cass_future_wait_check_and_free(cass_session_connect(
                session_raw.borrow(),
                cluster_raw.borrow().into_c_const(),
            ));

            let session_close_fut = cass_session_close(session_raw.borrow());
            // Crucial: intentionally not waiting to increase likelihood of the tokio task dropping the runtime.
            // cass_future_wait(session_close_fut.borrow());
            cass_future_free(session_close_fut);

            cass_cluster_free(cluster_raw);
            cass_session_free(session_raw);
        }
        proxy
    }
}
```

wprzytula:
Fixes: #362

The solution is here, as described in the issue.

I also wrote an automated regression test.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/runtime.rs
+++ b/scylla-rust-wrapper/src/runtime.rs
@@ -4,10 +4,48 @@
 
 use std::{
     collections::HashMap,
+    mem::ManuallyDrop,
     sync::{Arc, Weak},
 };
 
-use tokio::runtime::Runtime;
+/// This is a wrapper to ensure that the runtime is properly shut down even
+/// if dropped in the async context. This is done by using a custom
+/// Drop implementation that calls `shutdown_background`, as opposed
+/// to the default `Drop` implementation of `Runtime` that calls
+/// `shutdown` which panics if called from the async context.
+#[repr(transparent)]
+pub(crate) struct Runtime {
+    inner: ManuallyDrop<tokio::runtime::Runtime>,
+}
+
+impl From<tokio::runtime::Runtime> for Runtime {
+    fn from(inner: tokio::runtime::Runtime) -> Self {
+        Self {
+            inner: ManuallyDrop::new(inner),
+        }
+    }
+}
+
+impl std::ops::Deref for Runtime {
+    type Target = tokio::runtime::Runtime;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl Drop for Runtime {
+    fn drop(&mut self) {
+        // SAFETY: We are executing the drop logic only once, here.
+        // Drop::drop is guaranteed to be called only once, and we are taking
+        // the inner value out of the ManuallyDrop exactly once here.
+        let runtime = unsafe { ManuallyDrop::take(&mut self.inner) };
+
+        // We use `shutdown_background()` instead of the usual Runtime Drop impl to avoid panicking
+        // if the runtime is dropped in an async context.
+        runtime.shutdown_background();
+    }
+}
 
 /// Manages tokio runtimes for the application.
 ///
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/src/runtime.rs
```diff
@@ -4,10 +4,46 @@
...
+            inner: ManuallyDrop::new(inner),
+        }
+    }
+}
+
+impl Runtime {
+    pub(crate) fn inner(&self) -> &tokio::runtime::Runtime {
+        &self.inner
+    }
+}
```
Lorak-mmk: Maybe implementing a trait like Deref (or maybe AsRef) would be a good idea here, instead of calling `inner()` everywhere?

wprzytula: I did it initially, but then I concluded it's an antipattern to abuse Deref for something that **is not** a _smart pointer_.

[`Deref` guidelines](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut) do allow using it in this case, though.

Lorak-mmk: > [Deref guidelines](if(connections[i].source) 

That looks like a copy-paste error.

IIRC Deref guidelines changed at some point. Before they did say it should only be implemented for smart pointers. Now they are not as strict. For example, they say:

> The same advice applies to both deref traits. In general, deref traits should be implemented if:
>     - a value of the type transparently behaves like a value of the target type;
>     - the implementation of the deref function is cheap; and
>     - users of the type will not be surprised by any deref coercion behavior.

I think all of those conditions are met here.

wprzytula: > IIRC Deref guidelines changed at some point.

This would explain a lot. In this case I'll implement `Deref` and get rid of those `inner()` calls.

Lorak-mmk (approved):

wprzytula:
I think I solved the issue with address collisions between main process and subprocesses. It was caused by violation of uniqueness of addresses that the proxy node listen on, which was caused by existence of multiple address generators (in different processes). The idea is to generate addresses only in the main process and pass them to subprocesses dynamically. This is done by a new macro, `rusty_fork_test_with_proxy`, which is based on `rusty_fork_test` macro and extended to pass the address for the proxy to listen on.

@Lorak-mmk please take a look.

wprzytula:
Once you approve the new approach, I'll amend the commit that introduces the test, so that all tests pass on all commits.

Lorak-mmk (commented):
I like the new approach

wprzytula:
Amended the commits. Ready for final review.

Lorak-mmk (approved):

PR #369, opened 2025-09-22
change default serial consistency to `LOCAL_SERIAL`

wprzytula:
Although the C++ Driver default is `ANY`, we believe `LOCAL_SERIAL` is a safer default: `ANY` is not valid for LWTs, and sending it will result in errors from the server.

Fixes: https://github.com/scylladb/cpp-rust-driver/issues/335

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -40,8 +40,9 @@ use crate::cass_compression_types::CassCompressionType;
 // According to `cassandra.h` the defaults for
 // - consistency for statements is LOCAL_ONE,
 pub(crate) const DEFAULT_CONSISTENCY: Consistency = Consistency::LocalOne;
-// - serial consistency for statements is ANY, which corresponds to None in Rust Driver.
-const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = None;
+// - serial consistency for statements is LOCAL_SERIAL. This is different from CPP Driver's ANY - see
+//   https://github.com/scylladb/cpp-rust-driver/issues/335 for context.
+const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = Some(SerialConsistency::LocalSerial);
 // - request client timeout is 12000 millis,
 const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_millis(12000);
 // - fetching schema metadata is true
--- a/include/cassandra.h
+++ b/include/cassandra.h
@@ -1700,7 +1700,7 @@ cass_cluster_set_consistency(CassCluster* cluster,
 /**
  * Sets default serial consistency level of statement.
  *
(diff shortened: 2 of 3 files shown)


Lorak-mmk (approved):

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -41,7 +41,10 @@ use crate::cass_compression_types::CassCompressionType;
 // - consistency for statements is LOCAL_ONE,
 pub(crate) const DEFAULT_CONSISTENCY: Consistency = Consistency::LocalOne;
 // - serial consistency for statements is ANY, which corresponds to None in Rust Driver.
```
dkropachev: I see that it makes sense when you know history of this options, but if you just read this comment as brand new user it does not make much sense.

wprzytula: Which specific comment do you refer to? How can I improve it?

dkropachev: After this commit it will look like this:
```
// - serial consistency for statements is ANY, which corresponds to None in Rust Driver.
//   However, we deem it safer to use LOCAL_SERIAL as the default, since executing an LWT
//   with ANY serial consistency is not valid and the DB will reject such requests.
//   See https://github.com/scylladb/cpp-rust-driver/issues/335 for context.
const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = Some(SerialConsistency::LocalSerial);
```

Why don't update docstring for `cass_cluster_set_serial_consistency` in `cassandra.h` to:
```
/**
 * Sets default serial consistency level of statement.
 *
 * <b>Default:</b> CASS_CONSISTENCY_LOCAL_SERIAL
 *
 * @public @memberof CassCluster
 *
 * @param[in] cluster
 * @param[in] consistency
 * @return CASS_OK if successful, otherwise an error occurred.
 */
CASS_EXPORT CassError
cass_cluster_set_serial_consistency(CassCluster* cluster,
                                    CassConsistency consistency);
```

And have simple comment for it in `cluster.rs` similar to others:
```
// - serial consistency for statements is LOCAL_SERIAL
const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = Some(SerialConsistency::LocalSerial);
```

wprzytula: Ah, got it. The intention behind this was to have `cassandra.h` which is the original one, taken from CPP Driver without modifications - so we can claim drop-in compatibility.

But that's incorrect - we can't ship a header file with an API and its implementation that differs from behaviour described in that header.

So I agree, I'll update `cassandra.h` as you suggest.

wprzytula: Applied your suggestion.

Lorak-mmk (approved):

dkropachev (approved):

PR #375, opened 2025-09-24
Restructure modules & move some session tests to integration

wprzytula:
# What's done 

This PR improves the module structure of the driver by two actions:

## Module refactor

Some moduled are moved, grouped, extracted or annihilated, so that the structure makes more sense.

## Integration tests extraction

Some tests from `session` module, which only use the public API, are extracted to the `integration` module as integration tests.

### Makefile

`rustup` no longer updates *all* toolchains upon running make commands, just the `stable` toolchain.

### `api.rs`

`CassUuid`, which must have been accidentally omitted, is now exposed.

### `integration`

The helper function to test with the proxy has now more convenient semantics:
- it no longer accepts `ShardAwareness`, which is no-op for the dry mode proxy;
- it accepts _initial request rules_ as the parameter, as a shorthand to configure the proxy;
- it ensures the closure is run under `tokio::task::spawn_blocking`, which is required not to experience panics upon blocking in the async context; the closure itself is no longer `async`, which is good - the C API is not `async` either.

## Pre-review checklist


wprzytula:
Rebased on master.

Lorak-mmk (approved):

PR #379, opened 2025-09-30
Switch to CPack for packaging

### LINKED ISSUE #376: Packaging issues before release
1. Supported distros? 
2. CICD Pipeline as part of release or separate one.
3. Backport (to prior releases) binaries for some distros ? (have workflow for that ?)
4. Maybe change packager?
5. Package naming ?
6. Packages testing
7. Distribution for windows
8. Distribution for macos

## Supported distros ##
1. Have a policy
2. Come up with list of distros based on policy

## Packager ##
1. Decide if we switch to another packager
2. Better to have it integrated into cmake
3. Cross compilation support for macos and windows ?

## Package naming ##

Question: 
Do we add build date and hash id to the package, like so: `libscylla-cpp-rs-driver1_1.0.0-20240328.f143d09a2057_noble_amd64.deb` ?

### .deb
`regular package` - `libscylla-cpp-rs-driver<major>_<version>_<os>_<arch>.deb` - `libscylla-cpp-rs-driver1_1.0.0_noble_amd64.deb`
`debug symbols pacakge` - `libscylla-cpp-rs-driver<major>-dbgsym_<version>_<os>_<arch>.deb` - `libscylla-cpp-rs-driver1-dbgsym_1.0.0_noble_amd64.deb`
`dev pacakge` - `libscylla-cpp-rs-driver-dev_<version>_<os>_<arch>.deb` - `libscylla-cpp-rs-driver-dev_1.0.0_noble_amd64.deb`

### .rpm
`regular package` - `scylla-cpp-rs-driver-<version>.<os>.<arch>.rpm` - `scylla-cpp-rs-driver-1.0.0.fc42.amd64.rpm`
`debug symbols pacakge` - `scylla-cpp-rs-driver-debuginfo-<version>.<os>.<arch>.rpm` - `scylla-cpp-rs-driver-debuginfo-1.0.0.fc42.amd64.rpm`
`dev pacakge` - `scylla-cpp-rs-driver-devel-<version>.<os>.<arch>.rpm` - `scylla-cpp-rs-driver-devel-1.0.0.fc42.amd64.rpm`

dkropachev:
## Summary

This PR modernizes the driver's packaging infrastructure by replacing the custom shell scripts with CMake's CPack.
The new approach is more maintainable, cross-platform, and integrates better with CI/CD pipelines.

## Key Changes

Packaging Infrastructure:
- Replace custom dist/debian/ and dist/redhat/ build scripts with CPack configuration
- Add CPack generators for all platforms: DEB, RPM (Linux), productbuild/DragNDrop (macOS), WIX (Windows)
- Add component-based installation separating runtime libraries from development files
- Add cmake/RunMacProductbuild.cmake helper for macOS PKG builds

## CI/CD Workflows:
- Add build-cpack-packages.yml - reusable workflow for building packages on Ubuntu, macOS, and Windows
- Add release-upload-packages.yml - automated package upload to GitHub releases on tag push
- Update pkg.yml to delegate to new workflows

## Smoke Test Application:
- Add packaging/smoke-test-app/ with a C application that validates installed packages
- Connects to a cluster and queries system.local to verify the driver works
- Includes its own CPack configuration for testing the full install -> build -> run cycle
- Add docker-compose.yml for containerized RPM testing with Fedora

## Makefile Targets:
- make build-package - auto-detects OS and builds appropriate packages
- make test-package - end-to-end package testing for all formats
- make test-package-{deb,rpm,pkg,dmg,msi} - format-specific testing

## Bug Fixes:
- Fix error code handling in cass_error.rs
- Fix CMake cargo invocation and OpenSSL path detection

## Deleted

- dist/debian/ - custom deb build scripts and templates
- dist/redhat/ - custom RPM spec and build scripts
- dist/homebrew/ - homebrew formula (can be re-added later if needed)

## Test Plan

- Verify make build-package works on Linux (creates DEB + RPM)
- Verify make test-package-deb installs and runs smoke test successfully
- Verify make test-package-rpm works in Fedora container
- Verify CI workflow `build-cpack-packages.yml` passes on all platforms
- Verify release workflow uploads packages correctly on tag push

## Pre-review checklist

- [x] I have enabled appropriate tests in `Makefile` in `{SCYLLA,CASSANDRA}_(NO_VALGRIND_)TEST_FILTER`.
- [ ] I added appropriate `Fixes:` annotations to PR description.


CMakeLists.txt
```diff
@@ -243,3 +243,62 @@ endif()
...
+# Packaging (CPack)
+#------------------------
+
+set(_CPACK_PACKAGE_NAME "scylla-cpp-driver")
+set(CPACK_PACKAGE_NAME ${_CPACK_PACKAGE_NAME})
+set(CPACK_PACKAGE_VENDOR "ScyllaDB")
+set(CPACK_PACKAGE_CONTACT "ScyllaDB <info@scylladb.com>")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY
+  "ScyllaDB C++ driver backed by the Rust core driver")
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/scylladb/cpp-rust-driver")
```
wprzytula: :wrench:

wprzytula:
```suggestion
set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/scylladb/cpp-rs-driver")
```

CMakeLists.txt
```diff
@@ -243,3 +243,62 @@ endif()
...
+elseif(APPLE)
+  configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE"
+    "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt"
+    COPYONLY)
+  set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt")
+  set(CPACK_GENERATOR "productbuild;DragNDrop")
+  set(CPACK_PACKAGE_FILE_NAME
+    "${_CPACK_PACKAGE_NAME}-${PROJECT_VERSION_STRING}-macos")
+  set(CPACK_PRODUCTBUILD_IDENTIFIER "com.scylladb.cpp-rust-driver")
```
wprzytula: :question: Sometimes you write `cpp-driver` and sometimes `cpp-rust-driver`. Is the discrepancy intended?

:wrench: `cpp-rust-driver` -> `cpp-rs-driver`

MAINTENANCE.md
```diff
@@ -35,9 +35,9 @@ Description of this PR should consist of just release notes. **NOTE: Preferably
...
+11. Build binary packages with CPack on each target platform (Linux, macOS, Windows). The README provides the exact command sequence (`cmake -S`, `cmake --build`, `cpack -G <generator>`). If you do not have access to the platforms locally, trigger a run of the reusable workflow `.github/workflows/build-cpack-packages.yml` from another workflow and download the resulting artifacts.
+12. Go to https://github.com/scylladb/cpp-rust-driver/releases , click the `Draft new release` button and follow the procedure to create a new release on GitHub. Use the release notes as its description.
+13. After the release is published, the `Attach Packages to Release` workflow (`.github/workflows/release-upload-packages.yml`) automatically builds fresh packages using CPack and uploads every artifact to the release. Verify that the workflow finished successfully and that RPM/DEB/MSI/PKG/DMG files are attached.
```
wprzytula: :question: What is the intended difference between building packages in point 11 and 13?

README.md
```diff
@@ -424,48 +424,47 @@ Now, use `--gtest_filter` to run certain integration tests:
...
+From the `build` directory you can invoke `cpack` with the generator that
+matches your target platform:
+
+* **Linux (Debian/Ubuntu/Fedora/Rocky):**
+  ```shell
+  cpack -G DEB -C Release   # produces .deb in build/
+  cpack -G RPM -C Release   # produces .rpm in build/
+  ```
+  Debian packages set shlibdeps automatically; RPM packages target the
+  `Applications/Databases` group and use release number `1` by default.
```
wprzytula: :question: What does it mean? It's all Greek to me.

README.md
```diff
@@ -424,48 +424,47 @@ Now, use `--gtest_filter` to run certain integration tests:
...
+
+* **macOS (pkg / dmg):**
+  ```shell
+  cpack -G productbuild -C Release  # generates .pkg
+  cpack -G DragNDrop -C Release     # generates .dmg
+  ```
+  Both outputs bundle the compiled libraries alongside headers and pkg-config
+  manifests.
```
wprzytula: :question: This also uses naming that is unknown to me.

README.md
```diff
@@ -424,48 +424,47 @@ Now, use `--gtest_filter` to run certain integration tests:
...
+
+* **Windows (MSI):**
+  ```powershell
+  cpack -G WIX -C Release  # requires WiX Toolset 3.11+
```
wprzytula: > `requires WiX Toolset 3.11+`

How do I obtain it? Is it explained anywhere?

README.md
```diff
@@ -424,48 +424,47 @@ Now, use `--gtest_filter` to run certain integration tests:
...
+  ```
+  The MSI installs into `Program Files\ScyllaDB\Scylla CPP Driver` and reuses a
+  stable upgrade GUID for seamless upgrades.
```
wprzytula: :muscle: I can see you solved the problem with upgrades on Windows.

src/testing.cpp
```diff
@@ -47,44 +47,44 @@ String get_host_from_future(CassFuture* future) {
...
-        } else {
-            break; // No more addresses to read.
-        }
+  AddressVec attempted_addresses = response_future->attempted_addresses();
+  for (const auto& address : attempted_addresses) {
+      attempted_hosts.push_back(address.to_string());
+  }
+  std::sort(attempted_hosts.begin(), attempted_hosts.end());
+  return attempted_hosts;
+  */
```
wprzytula, line 65: :bow:  Thanks for fixing this.

:thinking: No CI step failed due to this code being inproperly formatted. What command did you use to format the code? Please add it as a step to the CI.

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -134,7 +134,11 @@ impl ToCassError for DbError {
             DbError::AlreadyExists { .. } => CassError::CASS_ERROR_SERVER_ALREADY_EXISTS,
             DbError::Unprepared { .. } => CassError::CASS_ERROR_SERVER_UNPREPARED,
             DbError::Other(num) => {
-                CassError((CassErrorSource::CASS_ERROR_SOURCE_SERVER.0 << 24) | *num as u32)
+                // On windows enums are i32, so we need to cast it
+                #[allow(clippy::unnecessary_cast)]
+                let source = CassErrorSource::CASS_ERROR_SOURCE_SERVER.0 as u32;
+                let code = (source << 24) | (*num as u32);
+                CassError(code as _)
```
wprzytula: Quite an unexpected catch.

.github/workflows/build-cpack-packages.yml
```diff
@@ -0,0 +1,118 @@
...
+        default: Release
+      extra-cmake-flags:
+        description: Additional flags passed to CMake configure step
+        type: string
+        default: ""
+      save-artifacts:
+        description: Save built packages as artifacts
+        type: boolean
+        default: false
+    secrets: {}
```
wprzytula, line 33: :memo: It's a pity we need to be that WET instead of DRY. Chat GPT, however, hasn't come up with a better solution for enabling both manual dispatch and being called from another workflow, both with the same parameters and configuration.

Makefile
```diff
@@ -1,5 +1,21 @@
 EMPTY :=
 SPACE := ${EMPTY} ${EMPTY}
+SHELL = bash
+
+SHELL = bash
```
wprzytula: :question: Why the duplication?

Makefile
```diff
@@ -1,5 +1,21 @@
...
+ifeq ($(OS),Windows_NT)
+    SHELL := pwsh.exe
```
wprzytula: :wrench: I believe it's an error to mix eager binding `:=` operator with late binding operator `=` for the same variable. Do you agree?

.github/workflows/build-cpack-packages.yml
```diff
@@ -47,12 +47,50 @@ jobs:
...
+          shopt -s nullglob
+          DRIVER_PACKAGES=(build/*.deb)
+          if [ "${#DRIVER_PACKAGES[@]}" -eq 0 ]; then
+            echo "No driver DEB packages produced"
+            exit 1
+          fi
+          sudo dpkg -i "${DRIVER_PACKAGES[@]}"
+          sudo apt-get install -f -y
+
+      - name: Build smoke-test application package
```
wprzytula: :pick: Let's add a missing `(DEB and RPM)` suffix, so this step's name isconsistency with other steps (`(DEB)` only) is explicit.

wprzytula (commented):
In general, the approach looks correct. Not knowing CPack, I'm not sure though.

CMakeLists.txt
dkropachev: done

CMakeLists.txt
dkropachev: changed

MAINTENANCE.md
dkropachev: Updated, please take a look.

README.md
dkropachev: Added sections to explain how to get it.

README.md
dkropachev: updated, please take a look

README.md
dkropachev: updated, please take a look

Makefile
dkropachev: Just, copy-paste, fixed

Makefile
dkropachev: Nope, it is lazy vs immediate assignment, in this context no difference.

.github/workflows/build-cpack-packages.yml
dkropachev: We can have two files, which is, IMHO, worse than this.

.github/workflows/build-cpack-packages.yml
dkropachev: Done

src/testing.cpp
dkropachev: Don't remember at this point, unfortunately.

dkropachev:
@wprzytula, finaly, I am done with it, please take a look.

Makefile
```diff
@@ -359,14 +446,198 @@ run-examples-scylla: build-examples
...
+
+# Windows MSI package testing
+test-package-msi: .windows-setup-wix build-package
+	@echo "=== Testing MSI packages ==="
+	$(MAKE) -C $(SMOKE_TEST_DIR) install-driver-dev
+	$(MAKE) -C $(SMOKE_TEST_DIR) install-driver
+	$(MAKE) -C $(SMOKE_TEST_DIR) build-package
+	$(MAKE) -C $(SMOKE_TEST_DIR) install-app
+	$(MAKE) -C $(SMOKE_TEST_DIR) test-app-package
+	@echo "=== MSI package test completed successfully ==="
```
wprzytula, line 626: :question: Why is Fedora testing so different from all others? I assume we're running on Ubuntu, so it's simple to test on Ubuntu. But what about testing MacOS and Windows? Those have simple instructions here, too, which are similar to Ubuntu's.

wprzytula:
@dkropachev Please address Copilot's comments.

Makefile
dkropachev: There is no `Fedora` testing, I only tested rpm on Ubuntu, which is not great, but better than nothing, my asumption was that there is no fedora runner availalbe, but now I see that there is one, let me updated cicd to have rpm tested properly.

packaging/smoke-test-app/Makefile
```diff
@@ -52,21 +52,27 @@ define VALIDATE_SINGLE_PACKAGE
 	package=$${packages[0]}
 endef
 
+# Use sudo only when not running as root (e.g., inside containers that run as root)
+define MAYBE_SUDO
+	if [ "$$(id -u)" -eq 0 ]; then SUDO=""; else SUDO="sudo"; fi
+endef
```
wprzytula:
```suggestion
# Use sudo only when not running as root (which is unnecessary e.g., inside containers that run as root)
define MAYBE_SUDO
	if [ "$$(id -u)" -eq 0 ]; then SUDO=""; else SUDO="sudo"; fi
endef
```

wprzytula:
@dkropachev Please ask me for re-review once the code is ready for it.

Makefile
dkropachev: take a look now

dkropachev:
@wprzytula , it is ready

packaging/smoke-test-app/Makefile
dkropachev: fixed

packaging/smoke-test-app/CMakeLists.txt
```diff
@@ -0,0 +1,292 @@
...
+elseif(APPLE)
+  set(CPACK_GENERATOR "productbuild;DragNDrop")
+  configure_file(
+    "${CPP_DRIVER_ROOT}/LICENSE"
+    "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt"
+    COPYONLY)
+  set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt")
+  set(CPACK_PACKAGE_FILE_NAME
+    "${_SMOKE_APP_PACKAGE_BASENAME}-${PROJECT_VERSION_STRING}-macos")
+  set(CPACK_PRODUCTBUILD_IDENTIFIER "com.scylladb.cpp-rs-driver.smoke-app")
```
wprzytula, line 263: :wrench: Sometimes you are using `CPP Driver` naming, sometimes `cpp-rs-driver`. Please make them consistent / justify the discrepancies.

wprzytula (approved):
Awesome work! Thank you! :rocket:

dkropachev:
@wprzytula , do we need to involve @Lorak-mmk here ?

packaging/smoke-test-app/CMakeLists.txt
dkropachev: The naming follows the same pattern as the main driver's `CMakeLists.txt` (lines 352 and 370):

- **Human-facing display names** (Windows install directories): "CPP Driver" with spaces for readability in UI
- **Technical identifiers** (macOS bundle IDs, URLs, package names): "cpp-rs-driver" with hyphens per reverse-DNS and package naming conventions

This is intentional and consistent with the main driver packaging configuration.

cmake/RunMacProductbuild.cmake
```diff
@@ -0,0 +1,59 @@
...
+  set(_cpack_executable "${CPACK_EXECUTABLE}")
+endif()
+
+string(REGEX MATCH "^(.+)_([0-9].*)$" _matched "${CPACK_PACKAGE_FILE_NAME}")
+if(NOT _matched)
+  message(FATAL_ERROR "Unexpected CPACK_PACKAGE_FILE_NAME format: ${CPACK_PACKAGE_FILE_NAME}")
+endif()
+set(_base_name "${CMAKE_MATCH_1}")
+set(_suffix "${CMAKE_MATCH_2}")
+
```
Lorak-mmk, line 53: What is this name? The regex is <any characters, 1+ times>, then underscore, then a single digit, then arbitrary characters (0+) again. Is that correct? Could you comment on what this name is and why the format is like that?

cmake/RunMacProductbuild.cmake
Lorak-mmk, line 1: 🤔 I don't know CPack, but this file looks very generic overall. It's just some vars validation, and calling an executable. I wonder why it is specific to Mac

CMakeLists.txt
```diff
@@ -250,3 +262,167 @@ endif()
...
+# Dev component: headers, static libraries, pkg-config files, and development symlinks
+cpack_add_component(${SCYLLA_DRIVER_DEV_COMPONENT_NAME}
+  DISPLAY_NAME "Scylla C++ Driver Development"
+  DESCRIPTION "Headers, static libraries, and development files for the Scylla C++ Driver")
+
+if(WIN32)
+  set(CPACK_GENERATOR "WIX")
+  set(CPACK_PACKAGE_INSTALL_DIRECTORY "ScyllaDB\\Scylla CPP Driver")
+  # Stable GUID keeps WiX upgrades working between versions
+  set(CPACK_WIX_UPGRADE_GUID "3E09F4F4-1C4A-4B8B-973D-0CBA64D5E78F")
```
Lorak-mmk: How was it generated?

Makefile
Lorak-mmk, line 1: Commit: Add package building targets to Makefile

It would be great to make formatting changes in separate commit. They make it more difficult to tell what are the important parts of the diff here.

tests/src/integration/tests/test_dc_aware_policy.cpp
```diff
@@ -189,7 +189,7 @@ CASSANDRA_INTEGRATION_TEST_F(DcAwarePolicyTest, AllowRemoteDcsForLocalCl) {
   initialize();
 
   Statement statement_with_nonlocal_consistency = select_statement("1");
-  statement_with_nonlocal_consistency.set_consistency(CASS_CONSISTENCY_TWO);
+  statement_with_nonlocal_consistency.set_consistency(CASS_CONSISTENCY_ONE);
 
   Statement statement_with_local_consistency = select_statement("1");
```
Lorak-mmk, line 194: This change is not explained anywhere. "Fix test_dc_aware_policy.cpp test assertions" Is not enough. Was this test failing? Why exactly was the old value wrong?

scylla-rust-wrapper/src/cass_error.rs
```diff
@@ -133,7 +133,11 @@ impl ToCassError for DbError {
...
```
Lorak-mmk: I don't understand this.
`num` in `DbError::Other` is always i32
`CassErrorSource::CASS_ERROR_SOURCE_SERVER.0` is c_uint, which according to docs must be unsigned, and is almost always i32.
What exactly becomes `i32` on Windows that is not `i32` on Linux?
Your change is to cast `CassErrorSource::CASS_ERROR_SOURCE_SERVER.0` to u32, so I assume the comment is about that, but `c_uint`  can never be i32.

tests/src/integration/tests/test_heartbeat.cpp
```diff
@@ -88,17 +88,16 @@ CASSANDRA_INTEGRATION_TEST_F(HeartbeatTests, HeartbeatFailed) {
...
-         elapsed_time() < 60000) {
-    session_.execute_async(SELECT_ALL_SYSTEM_LOCAL_CQL); // Simply execute statements ignore any
-                                                         // error that can occur from paused node
-    metrics = session_.metrics();
+  while (logger_.count() == 0 && elapsed_time() < 60000) {
+    msleep(100);
   }
-  EXPECT_LT(metrics.stats.total_connections, initial_connections);
   EXPECT_GE(logger_.count(), 1u);
 }
```
Lorak-mmk: > Fix test_heartbeat.cpp timing-sensitive tests

How exactly was this test timing-sensitive? The before state looks really solid to me, and your change removed useful assertions (that checked there are actually less connections after failure).

Lorak-mmk (changes requested):

CMakeLists.txt
dkropachev: randomly generated, it is an app id so that installed knew how to find installed app when new updates comes.
Added a comment.

tests/src/integration/tests/test_heartbeat.cpp
dkropachev: ### Here is how it was timing-sensitive:
1. Race between heartbeat timeout and query execution: The loop sent execute_async queries to all nodes (including the paused one). These in-flight queries could keep the connection to node 2 alive or cause
  connection state changes that interfered with the heartbeat mechanism detecting the failure.
2. Connection count is not a reliable signal: The total_connections metric can fluctuate for reasons unrelated to heartbeat - e.g., the driver may establish new connections as part of reconnection attempts
  even as old ones are dropped. The test checked total_connections < initial_connections, but the driver could have already started reconnecting, bringing the count back up.
3. No targeted interaction with node 2: The original test never explicitly sent a request to node 2 before pausing it, so there was no guarantee that a connection to node 2 was actively used and being
  heartbeated.

### The new version on other hand:
  - Explicitly sends a query to node 2 (ensuring there's an active connection)
  - Pauses node 2
  - Waits for the heartbeat timeout log message (the actual behavior under test)
  - Uses logger_.add_critera("Timed out while waiting for response to keepalive request") which directly validates heartbeat failure detection

The assertion on connection count was indirectly testing heartbeat failure. 
The log message is a direct signal that the heartbeat mechanism detected the failure, making the test both more reliable and more precise about what it validates. 
That said, if we want to keep the connection-count assertion too, it could be added back as a secondary check after the log message is observed.

cmake/RunMacProductbuild.cmake
dkropachev: The regex `^(.+)_([0-9].*)$` splits `CPACK_PACKAGE_FILE_NAME` (e.g. `scylla_cpp_driver_0.1.0_macos`) into:
- Group 1 (`_base_name`): `scylla_cpp_driver` — the component/package base name
- Group 2 (`_suffix`): `0.1.0_macos` — version + platform suffix

It anchors on the last underscore followed by a digit (the start of the version number) to split the name reliably. The purpose is so per-component packages can be generated by replacing the base name portion (e.g. `scylla_cpp_driver-dev_0.1.0_macos`).

Added a comment in the code explaining this.

cmake/RunMacProductbuild.cmake
dkropachev: CPack's `productbuild` generator (macOS `.pkg`) doesn't natively support building separate per-component packages in a single run — it bundles everything into one `.pkg`. Other generators (DEB, RPM) handle this via `CPACK_DEB_COMPONENT_INSTALL` / `CPACK_RPM_COMPONENT_INSTALL`, but productbuild has no equivalent.

This script works around that by invoking `cpack -G productbuild` once per component, overriding `CPACK_COMPONENTS_ALL` and `CPACK_PACKAGE_FILE_NAME` on each invocation. That's why it's Mac-specific.

Added a header comment in the file explaining this.

tests/src/integration/tests/test_dc_aware_policy.cpp
dkropachev: Yes, this test was failing. The test uses a 2-DC cluster (2 nodes each) with `SimpleStrategy` replication (default RF), then stops both local DC nodes. With `CASS_CONSISTENCY_TWO`, the query needs 2 replicas to respond, but after stopping DC1, there may not be enough replicas available in DC2 to satisfy CL=TWO. The query would fail with an unavailable error rather than succeeding via the remote DC — which is not what the test intends to validate.

The purpose of `statement_with_nonlocal_consistency` is simply to use a **non-local** consistency level (i.e., not `LOCAL_*`) to prove that non-local CLs are routed to remote DCs. `CASS_CONSISTENCY_ONE` serves that purpose equally well and doesn't risk failing due to insufficient replicas.

Updated the commit message to explain this.

scylla-rust-wrapper/src/cass_error.rs
dkropachev: You're right, the comment was misleading. The issue is in **bindgen's output**, not in the Rust `c_uint` type itself. `CassErrorSource` is generated by bindgen from the C header's `enum CassErrorSource_`. On Linux/GCC, bindgen represents C enums as newtype wrappers around `c_uint` (`u32`). On Windows/MSVC, the underlying type of C enums is `int` (signed), so bindgen generates the wrapper around `c_int` (`i32`). Thus `CassErrorSource::CASS_ERROR_SOURCE_SERVER.0` is `u32` on Linux but `i32` on Windows.

The `as u32` cast normalizes this, and `#[allow(clippy::unnecessary_cast)]` suppresses the clippy warning on Linux where it's already `u32`.

Fixed the comment to say: "Bindgen generates C enum wrappers with platform-dependent underlying types: c_uint (u32) on Linux, c_int (i32) on Windows/MSVC. Cast to u32 to normalize."

Makefile
dkropachev: done, splitted

tests/src/integration/tests/test_dc_aware_policy.cpp
Lorak-mmk: The test is setting the following:
```
    number_dc1_nodes_ = 2;
    number_dc2_nodes_ = 2;
```

In `Integration::SetUp` we have `replication_strategy_ = default_replication_strategy();`
Definition of `default_replication_strategy`:
```
std::string Integration::default_replication_strategy() {
  // Determine the replication strategy
  std::stringstream replication_strategy_s;
  if (number_dc2_nodes_ > 0) {
    replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': " << number_dc1_nodes_ << ", "
                           << "'dc2': " << number_dc2_nodes_;
  } else {
    replication_strategy_s << "'NetworkTopologyStrategy', 'dc1': ";

    // Ensure the replication factor has not been overridden or already set
    if (replication_factor_ == 0) {
      replication_factor_ = default_replication_factor();
    }
    replication_strategy_s << replication_factor_;
  }

  // Return the default replication strategy
  std::string replication_strategy = replication_strategy_s.str();
  return format_string(REPLICATION_STRATEGY, replication_strategy.c_str());
}
```
`number_dc2_nodes_` is 2, so we'll go into first branch of the `if`, and produce strategy with 2 replicas in DC1 and 2 replicas in DC2. After shutting down DC1, `CASS_CONSISTENCY_TWO` should still work correctly.

tests/src/integration/tests/test_heartbeat.cpp
Lorak-mmk: > Race between heartbeat timeout and query execution: The loop sent execute_async queries to all nodes (including the paused one). These in-flight queries could keep the connection to node 2 alive or cause
connection state changes that interfered with the heartbeat mechanism detecting the failure.

WDYM? How can sending requests to dead node keep its connection alive? If anything, it will make driver detect the dead node faster, before the normal heartbeat detects it.

> Connection count is not a reliable signal: The total_connections metric can fluctuate for reasons unrelated to heartbeat - e.g., the driver may establish new connections as part of reconnection attempts
even as old ones are dropped. The test checked total_connections < initial_connections, but the driver could have already started reconnecting, bringing the count back up.

Not true afaik, connection count in metrics will only increase when a connection is opened. It cannot be opened if the node is pasued.


So:
- I do see the point of removing the `execute_async` in a loop: we want the heartbeat (not user queries) to detect paused node.
- I think we should keep the metrics, not the logger, in the loop condition. What we want is the connections to actually be dropped, not just a log message about it emitted. Metrics are the most direct way to check that this happened.
- We should also keep the assert about connection count after the loop.
- Sending initial request to node 2 is fine, as an additional hardening mechanism.

dkropachev: > > Race between heartbeat timeout and query execution: The loop sent execute_async queries to all nodes (including the paused one). These in-flight queries could keep the connection to node 2 alive or cause
> > connection state changes that interfered with the heartbeat mechanism detecting the failure.
> 
> WDYM? How can sending requests to dead node keep its connection alive? If anything, it will make driver detect the dead node faster, before the normal heartbeat detects it.
> 
> > Connection count is not a reliable signal: The total_connections metric can fluctuate for reasons unrelated to heartbeat - e.g., the driver may establish new connections as part of reconnection attempts
> > even as old ones are dropped. The test checked total_connections < initial_connections, but the driver could have already started reconnecting, bringing the count back up.
> 
> Not true afaik, connection count in metrics will only increase when a connection is opened. It cannot be opened if the node is pasued.
> 
> So:
> 
> * I do see the point of removing the `execute_async` in a loop: we want the heartbeat (not user queries) to detect paused node.
> * I think we should keep the metrics, not the logger, in the loop condition. What we want is the connections to actually be dropped, not just a log message about it emitted. Metrics are the most direct way to check that this happened.
> * We should also keep the assert about connection count after the loop.
> * Sending initial request to node 2 is fine, as an additional hardening mechanism.

Np, let's move it out to another PR and continue discusstion there.

tests/src/integration/tests/test_dc_aware_policy.cpp
dkropachev: let's move it to another PR

scylla-rust-wrapper/src/cass_error.rs
Lorak-mmk: I think your code works for both i32 and u32. Can you add the following block, to make sure we don't have any surprises on other platforms?
```
                const {
                    // On some systems enums will be signed, and on some unsigned,
                    // and we need to deal with that.
                    // What we do assume, is that on all supported systems those enums will
                    // have a size of 4 bytes. If this is not true, we wan't to fail early (at
                    // compile time).
                    assert!(std::mem::size_of::<CassError>() == 4);
                    assert!(std::mem::size_of::<CassErrorSource>() == 4);
                }
```

tests/src/integration/tests/test_heartbeat.cpp
dkropachev: done, please check

tests/src/integration/tests/test_dc_aware_policy.cpp
dkropachev: done, please check.

scylla-rust-wrapper/src/cass_error.rs
dkropachev: done

scylla-rust-wrapper/src/testing/utils.rs
```diff
@@ -185,7 +185,7 @@ pub(crate) async fn test_with_one_proxy(
 /// It is also possible to specify a timeout which is applied to the test in
 /// the block, by adding the following line straight before the test:
 ///
-/// ```rust,no_compile
+/// ```no_rust
 /// #![rusty_fork(timeout_ms = 1000)]
```
Lorak-mmk, line 256: Commit: "Fix invalid rustdoc code block attributes"

Commit message should mostly be about "why was this done? why was this done this way?", and less about "what was done?".

In this commit, I have no idea why you made the change, and why did you decide on `no_rust`. Not really a helpful message.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/testing/utils.rs
dkropachev: Added explanation what it fixes.

scylla-rust-wrapper/src/testing/utils.rs
```diff
@@ -185,7 +185,7 @@ pub(crate) async fn test_with_one_proxy(
 /// It is also possible to specify a timeout which is applied to the test in
 /// the block, by adding the following line straight before the test:
 ///
-/// ```rust,no_compile
+/// ```no_rust
```
Lorak-mmk: Please use `no_run`. See https://rust-lang.github.io/rust-clippy/master/index.html#test_attr_in_doctest

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/testing/utils.rs
dkropachev: after rebase this changes are gone.

Lorak-mmk (approved):

wprzytula (approved):

PR #382, opened 2025-10-01
Categorize unimplemented functions

### LINKED ISSUE #373: Unimplemented functions in `cassandra.h` - what we do about them
## Current situation

There are quite a number of functions declared in `cassandra.h` which we don't implement in the wrapper. Currently, calling them in the user code will result in linker errors (unresolved symbols).

## Problem

Linker errors because of unimplemented functions may be confusing. Also, they give no insight to the user whether the function is not implemented _yet_ or it's not even planned to be ever implemented.

## Suggested solution

Divide the unimplemented functions into two categories:
1. TODO.
2. Will never be implemented.

Functions that are TODO shall have it clearly stated in the docstring, so that their unimplementedness is visible to the user, along with the intention to eventually have it implemented.

Functions that we don't plan to ever implement **shall be deleted** from `cassandra.h`, not to confuse users. This makes users experience compiler errors instead of linker errors, which is normally less unexpected.

wprzytula:
This PR treats the unimplemented functions according to their categorization in #373.
- not planned functions are moved from `cassandra.h` to `cassandra_deleted_functions.h`. The new header is only included in tests for them to compile and correctly link against the Rust stubs  (which are present only for tests).
- planned functions have a warning added about them not being implemented yet.

Bonus:
Default ScyllaDB version for tests is bumped to 2025.3.

Fixes: #373

## Pre-review checklist


examples/execution_profiles/execution_profiles.c
```diff
@@ -205,7 +204,7 @@ int main(int argc, char* argv[]) {
 
   /* Create a keyspace and table for the execution profile example  */
   execute_query(session, "CREATE KEYSPACE IF NOT EXISTS examples WITH replication = { \
-                  'class': 'SimpleStrategy', 'replication_factor': '3' \
+                  'class': 'NetworkTopologyStrategy', 'replication_factor': '3' \
```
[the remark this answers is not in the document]
wprzytula: This is incorrect.

examples/schema_meta/schema_meta.c
```diff
@@ -323,9 +321,6 @@ void print_meta_field(const CassIterator* iterator, int indent) {
   size_t name_length;
   const CassValue* value;
 
-  cass_iterator_get_meta_field_name(iterator, &name, &name_length);
-  value = cass_iterator_get_meta_field_value(iterator);
-
```
Lorak-mmk, line 328: I don't think you can just remove those lines. Sure, the code compiles, but it no longer makes sense. Now you pass NULL `value` to `print_schema_value`, so `print_meta_field` will always print `null` as value.

Lorak-mmk (changes requested):

wprzytula:
Rebased on master.

examples/schema_meta/schema_meta.c
wprzytula: Solved.

examples/schema_meta/schema_meta.c
Lorak-mmk, line 1: I don't think we understood each other when it comes to this example.
I understand that we won't implement metadata raw string accessors because:
- It is difficult (requires duplicating work already done by Rust Driver, or exposing some new APIs from Rust Driver)
- They have no good use case and it is better to use normal metadata API that we implemented in the driver.

Now I see that in this example you removed usage of raw string metadata API, but did not replace it with normal metadata API. Why? It should be replaced, if not for other reason then to prove that our decision to not implement raw API was correct.

Lorak-mmk (changes requested):

examples/schema_meta/schema_meta.c
wprzytula: I can't even empirically verify that the raw metadata API can be replaced with use of existing metadata API functions, because those functions are not implemented! Indexes, functions and aggregates aren't supported by the Rust Driver.
What's the use of re-implementing those examples if they are still not runnable anyway?
@Lorak-mmk

wprzytula: As discussed, I reverted the changes to `schema_meta.c` example and moved the metadata raw accessors API from _Deleted_ to _Planned_ section.

Lorak-mmk (approved):

PR #390, opened 2025-11-26
Update scylla-rust-driver to v1.4.1

wprzytula:
## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/statements/prepared.rs
+++ b/scylla-rust-wrapper/src/statements/prepared.rs
@@ -41,7 +41,7 @@ impl CassPrepared {
             .collect();
 
         let result_metadata = Arc::new(CassResultMetadata::from_column_specs(
-            statement.get_result_set_col_specs(),
+            statement.get_current_result_set_col_specs().get(),
         ));
 
         Self {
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/statements/prepared.rs
```diff
@@ -41,7 +41,7 @@ impl CassPrepared {
             .collect();
 
         let result_metadata = Arc::new(CassResultMetadata::from_column_specs(
-            statement.get_result_set_col_specs(),
+            statement.get_current_result_set_col_specs().get(),
```
Lorak-mmk, line 44: ⚠️🔧 This is absolutely not enough. The method was deprecated because the result metadata can now change. This code processes **during construction** it into CassResultMetadata, which is immutable, and later uses it for deserialization. If the metadata changes in the meantime, you're going to deserialize with wrong metadata.

Also there is comment above which will need to be updated after this is fixed.

scylla-rust-wrapper/tests/integration/consistency.rs
```diff
@@ -178,7 +179,9 @@ fn check_consistencies(
     mut request_rx: UnboundedReceiver<(RequestFrame, Option<TargetShard>)>,
 ) -> UnboundedReceiver<(RequestFrame, Option<TargetShard>)> {
     let (request_frame, _shard) = request_rx.blocking_recv().unwrap();
-    let deserialized_request = request_frame.deserialize().unwrap();
+    let deserialized_request = request_frame
+        .deserialize(&ProtocolFeatures::default())
+        .unwrap();
     assert_eq!(deserialized_request.get_consistency().unwrap(), consistency);
```
Lorak-mmk, line 185: ⚠️🔧 This is deserializing EXECUTE frames, right? In that case you need to fetch real protocol features and use them. With the result metadata extension, execute gets a new field, so deserialization will fail if this test is executed with recent Scylla version.

Lorak-mmk (changes requested):

Lorak-mmk:
closing in favor of https://github.com/scylladb/cpp-rs-driver/pull/391

PR #393, opened 2025-12-02
Implement cass_cluster_set_resolve_timeout

### LINKED ISSUE #258: Implement DNS resolution timeout (`cass_cluster_set_resolve_timeout`)
It waits for https://github.com/scylladb/scylla-rust-driver/pull/1113.

Lorak-mmk:
DNS timeouts were recently added to Rust Driver, which allows us to implement the missing function here.
This PR bumps Rust Driver to the commit that contains this new feature, and adds the function.

I did not enable any tests, because I found no tests that use this. I also don't see how I could write them.

Fixes: https://github.com/scylladb/cpp-rs-driver/issues/258

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -45,6 +45,8 @@ pub(crate) const DEFAULT_CONSISTENCY: Consistency = Consistency::LocalOne;
 const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = Some(SerialConsistency::LocalSerial);
 // - request client timeout is 12000 millis,
 const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_millis(12000);
+// - dns resolve timeout is 2000 millis. Note that this is different from Rust Driver default (5000).
+const DEFAULT_DNS_RESOLVE_TIMEOUT: Duration = Duration::from_millis(2000);
 // - fetching schema metadata is true
 const DEFAULT_DO_FETCH_SCHEMA_METADATA: bool = true;
 // - schema agreement timeout is 10000 millis,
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -45,6 +45,8 @@ pub(crate) const DEFAULT_CONSISTENCY: Consistency = Consistency::LocalOne;
 const DEFAULT_SERIAL_CONSISTENCY: Option<SerialConsistency> = Some(SerialConsistency::LocalSerial);
 // - request client timeout is 12000 millis,
 const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_millis(12000);
+// - dns resolve timeout is 2000 millis. Note that this is different from Rust Driver default (5000).
+const DEFAULT_DNS_RESOLVE_TIMEOUT: Duration = Duration::from_millis(2000);
```
wprzytula, line 49: :thinking: It seems tight, don't you think?

wprzytula: :question: What happens to the driver when DNS resolve timeout elapses?

wprzytula: Is there any fallback and recovery, or will the driver just error out?

Lorak-mmk: > 🤔 It seems tight, don't you think?

Why? Normally DNS query should not take more than a few milliseconds.

Lorak-mmk: > ❓ What happens to the driver when DNS resolve timeout elapses?
> Is there any fallback and recovery, or will the driver just error out?

Driver gets a list of contact points from the user (either dns names, or ip addresses). DNS names need to be resolved. If some fail (because of timeout or misconfiguration), they just won't be present in final resolved contact point list.

If this happends during initial resolution (session setup), then `NewSessionError::FailedToResolveAnyHostname` will be returned.
If this happens in our "emergency" path (failed to establish control connection to any know peer, trying to re-resolve contact points), then we will simply not resolve those contact points (and so not attempt a connection there). Driver will attempt again during next metadata refresh.

wprzytula:
@Lorak-mmk please open an issue about the failed test likely being flaky.

wprzytula (approved):

Lorak-mmk:
Opened https://github.com/scylladb/cpp-rs-driver/issues/395

PR #396, opened 2025-12-04
Implement contact point shuffling

### LINKED ISSUE #241: Implement `cass_cluster_set_use_randomized_contact_points`
It waits for `ControlConnectionPolicy` on rust-driver side. See: https://github.com/scylladb/scylla-rust-driver/issues/1279

Lorak-mmk:
It is done in a simple way - by just shuffling the list before passing it to Rust Driver. This matches the cpp-driver implementation.

A possibly better way is to implement ContronConnectionPolicy in Rust Driver, and use it here to make shuffling possible for all hosts (not only contact points) and re-shuffle each time CC is broken. This is a greater effort, so for now lets go with cpp-driver impl.

I wanted to enable `CASSANDRA_INTEGRATION_TEST_F(ControlConnectionFourNodeClusterTests, RandomizedContactPoints)`, but it assumes that round robin policy is actually round robin, which is not the case for our fork. I also fail to see how this test verifies contact point randomization.

After merging, I'll open and issue about improving this mechanism in the futre using CCPolicy.

Fixes: https://github.com/scylladb/cpp-rs-driver/issues/241

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/cluster.rs
+++ b/scylla-rust-wrapper/src/cluster.rs
@@ -14,6 +14,8 @@ use crate::timestamp_generator::CassTimestampGen;
 use crate::types::*;
 use openssl::ssl::SslContextBuilder;
 use openssl_sys::SSL_CTX_up_ref;
+use rand::seq::SliceRandom;
+use rand::thread_rng;
 use scylla::client::execution_profile::ExecutionProfileBuilder;
 use scylla::client::session_builder::SessionBuilder;
 use scylla::client::{PoolSize, SelfIdentity, WriteCoalescingDelay};
@@ -104,6 +106,7 @@ pub struct CassCluster {
     auth_password: Option<String>,
 
     client_id: Option<uuid::Uuid>,
+    shuffle_contact_points: bool,
 }
 
 impl CassCluster {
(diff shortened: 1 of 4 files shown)


scylla-rust-wrapper/src/cluster.rs
```diff
@@ -539,6 +548,21 @@ pub unsafe extern "C" fn cass_cluster_set_use_schema(
...
+pub unsafe extern "C" fn cass_cluster_set_use_randomized_contact_points(
+    cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
+    enabled: cass_bool_t,
+) -> CassError {
+    let Some(cluster) = BoxFFI::as_mut_ref(cluster_raw) else {
+        tracing::error!("Provided null cluster pointer to cass_cluster_set_use_schema!");
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
+    cluster.shuffle_contact_points = enabled == cass_true;
```
wprzytula, line 561: ❓ Wouldn't it be more robust to check for `enabled != cass_false`? This would make `enabled > 1`, e.g. `2`, result in Rust `true` instead of `false`, which is more expected.

wprzytula (approved):

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: I can change that, but why is that more expected? Any value outside of true / false is undefined / unspecified, right?

wprzytula: Well, probably.

PR #398, opened 2025-12-11
cluster: implement `cass_cluster_set_host_listener_callback`

### LINKED ISSUE #276: Implement cass_cluster_set_host_listener_callback

wprzytula:
The host listener feature allows users to register a callback function
that will be invoked when certain events occur on the cluster's hosts,
such as when a host is added, removed, goes up, or goes down. This is
useful for monitoring the state of the cluster and reacting to changes
in real-time.

The host_listener example is unfortunately not suitable for automated
running in the CI, as it is designed to run indefinitely, listening
for host events.

### Testing

I confirmed in manual tests that CPP-RS Driver's behaviour is in line
with CPP Driver's behaviour for host listener callbacks, by running
both drivers against a ccm-controlled ScyllaDB cluster and observing
that both drivers received the same host events when run host_listener.c
example.

Fixes: #276

## Pre-review checklist


Makefile
```diff
@@ -152,7 +152,7 @@ SCYLLA_EXAMPLES_TO_RUN := \
 	uuids \
 
 	# auth <- unimplemented `cass_cluster_set_authenticator_callbacks()`
-	# host_listener <- unimplemented `cass_cluster_set_host_listener_callback()`
+	# host_listener <- never terminates by design; loops forever listening to events.
 	# logging <- unimplemented `cass_cluster_set_host_listener_callback()`
```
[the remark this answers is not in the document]
wprzytula: Good observation. Not related to this PR though.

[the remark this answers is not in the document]
wprzytula: Low urgency.

wprzytula:
Made Cargo.toml target Rust driver's commit on main which contains the HostListener functionality.

scylla-rust-wrapper/src/testing/integration.rs
```diff
@@ -516,15 +515,6 @@ pub(crate) mod stubs {
...
 
-    #[unsafe(no_mangle)]
-    pub extern "C" fn cass_cluster_set_host_listener_callback(
-        _cluster_raw: CassBorrowedExclusivePtr<CassCluster, CMut>,
-        _callback: CassHostListenerCallback,
-        _data: *mut c_void,
-    ) -> CassError {
-        CassError::CASS_OK
-    }
-
```
Lorak-mmk, line 527: Not related to this PR: Can you remind me why we have both `scylla-rust-wrapper/src/testing/integration.rs` and `src/testing_unimplemented.cpp`? I noticed you had to remove the function from both.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -1660,6 +1663,26 @@ pub unsafe extern "C" fn cass_cluster_set_metadata_request_serverside_timeout(
...
+        tracing::error!(
+            "Provided null cluster pointer to cass_cluster_set_host_listener_callback!"
+        );
+        return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+    };
+
+    cluster.session_builder.config.host_listener = callback.map(|_| {
+        Arc::new(CCallbackBasedHostListener::new(callback, data)) as Arc<dyn HostListener>
+    });
+
```
Lorak-mmk: Here you ignore the arg passed to the closure that you give to map. This arg is a non-option C function. If you used it, you could skip checking Option in `CCallbackBasedHostListener::on_event`.

scylla-rust-wrapper/src/host_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl CCallbackBasedHostListener {
+    pub(crate) fn new(callback: CassHostListenerCallback, data: *mut c_void) -> Self {
+        Self { callback, data }
+    }
+}
+
+unsafe impl Send for CCallbackBasedHostListener {}
+unsafe impl Sync for CCallbackBasedHostListener {}
+
```
Lorak-mmk, line 32: This is interesting, and I didn't notice that when looking at Rust Driver impl. 
In Rust Driver, we only call on_event in the worker. That means there should be no problem with making on_event take &mut self and/or not requiring `Sync` (we still need `Send` because of Tokio ofc).
With such change, you could avoid using Mutex in rust Driver tests.

scylla-rust-wrapper/src/host_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+                    .copy_from_slice(&v6_addr.octets());
+
+                CassInet {
+                    address,
+                    address_length: CassInetLength::CASS_INET_V6 as u8,
+                }
+            }
+        }
+    }
+}
```
Lorak-mmk, line 122: As I mentioned on Slack, I'm fine with generating just one file using bindgen.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/host_listener.rs
wprzytula: Can we solve this separately? Refactoring bindgen would be quite a piece of work, and the workaround here is OK, just duplicates code.

Lorak-mmk: Sure. Can you open an issue?

scylla-rust-wrapper/src/testing/integration.rs
wprzytula: `testing_unimplemented.cpp` is used ONLY for C++ integration tests (`/tests/src/integration`).
`integration.rs` is used to provide stubs for:
- Rust integration tests,
- C examples,
- C++ integration tests, as proved by the code excerpt below.

```cmake
if(CASS_BUILD_INTEGRATION_TESTS OR CASS_BUILD_EXAMPLES)
  set(CMAKE_Rust_FLAGS "${CMAKE_Rust_FLAGS} --cfg cpp_integration_testing")
endif()
```

Note that function sets in both files **should** have empty intersection. Here you correctly noticed that `cass_cluster_set_host_listener_callback` was present in both files, which was incorrect and should have led to linker error. Perhaps error was not emitted because there are no C++ integration tests that call that function.

So the rule of thumb is to put stubs in `integration.rs` if they're useful to anything other than C++ integration tests, else leave it in `testing_unimplemented.cpp`.

scylla-rust-wrapper/src/host_listener.rs
wprzytula: Hmmm? What about the inspector side? How could we provide `&mut HostListener` to anyone if it was Arc'ed?

scylla-rust-wrapper/src/testing/integration.rs
Lorak-mmk: In that case, do we need testing_unimplemented.cpp? Maybe integration.rs is enough on its own?

scylla-rust-wrapper/src/host_listener.rs
Lorak-mmk: Right, that would require changing it to Box.
The `Sync` should be possible to remove, right? I'm not saying we should remove it.

wprzytula: `Arc<T>` is `Send` **iff** `T` is `Send + Sync`.
https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send-for-Arc%3CT,+A%3E

So we can't remove `Sync` from `HostListener` if we want to pass `Arc<HostListener>` to the tokio task with `ClusterWorker`.

scylla-rust-wrapper/src/testing/integration.rs
wprzytula: Probably yes, `testing_unimplemented.cpp` predates `integration.rs`, that's why it's even there.

scylla-rust-wrapper/src/host_listener.rs
wprzytula: https://github.com/scylladb/cpp-rs-driver/pull/398

scylla-rust-wrapper/src/host_listener.rs
Lorak-mmk: Ok, makes sense.

Why are we using Arc instead of Box?

wprzytula: 1. To make user able of keeping an instance of `HostListener` for their own use, most likely for reading the log if implemented similarly to one in our tests.
2. Not to break API in the future if we decide that after all for some reason we need to have `HostListener` accessible in more than a single place in the driver.

scylla-rust-wrapper/src/cluster.rs
wprzytula: Done, and added some docstrings.

scylla-rust-wrapper/src/host_listener.rs
Lorak-mmk: Ok

scylla-rust-wrapper/src/testing/integration.rs
Lorak-mmk: Let's not investigate that here for too long. Could you open an issue?

Lorak-mmk (approved):

scylla-rust-wrapper/src/testing/integration.rs
wprzytula: https://github.com/scylladb/cpp-rs-driver/issues/400

PR #406, opened 2025-12-15
Make `make run-integration-test` do not try to update rust

### LINKED ISSUE #401: Makefile is frustrating to use by developers
Current makefile tries to do both environment setup, and tasks (like building / running tests). This is great for CI, but not good for developers.
When  I exec `make run-test-integration-scylla`, I want makefile to build and run the integration tests. I don't want it to update Rust version on my machine (by calling rustup update), or update apt cache (I am on Fedora).

Imo those 2 things should be split. If there is a want to have all code in Makefile, this is fine, but such env-setup targets should never be called automatically.

dkropachev:
It is frustrating, we should not do it.

Fixes: https://github.com/scylladb/cpp-rs-driver/issues/401
## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,11 +259,11 @@ build-examples:
 		cmake -DCASS_BUILD_INTEGRATION_TESTS=off -DCASS_BUILD_EXAMPLES=on -DCMAKE_BUILD_TYPE=Release .. && (make -j 4 || make);\
 	}
 
-_update-rust-tooling:
+update-rust-tooling:
 	@echo "Run rustup update"
 	@rustup update stable
 
-check-cargo: install-cargo-if-missing _update-rust-tooling
+check-cargo: install-cargo-if-missing
 	@echo "Running \"cargo check\" in ./scylla-rust-wrapper"
 	@cd ${CURRENT_DIR}/scylla-rust-wrapper; cargo check --all-targets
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -315,7 +323,7 @@ download-ccm-cassandra-image: install-ccm-if-missing
 	@ccm create ccm_1 -i 127.0.1. -n 3:0 -v "${CASSANDRA_VERSION}" --config-dir=/tmp/download-cassandra.ccm
 	@rm -rf /tmp/download-cassandra.ccm
 
-run-test-integration-scylla: prepare-integration-test download-ccm-scylla-image
+run-test-integration-scylla: .prepare-environment-update-aio-max-nr prepare-integration-test
```
Lorak-mmk: Do I understand correctly that it will still call `prepare-integration-test`, which is something that prepares the environment? What it does:
- `prepare-environment-install-libc` which calls `dpkg` 
- `update-apt-cache-if-needed` which calls `apt-get`
- `install-valgrind-if-missing` which calls `apt`

It is fine to check if required tools are installed, and error out if not, but lets not implicitly install them. Installing in separate target is fine. Can you remove `prepare-integration-test` from user-facing targets (such as `run-test-integration-scylla`)?

Lorak-mmk (changes requested):

Makefile
dkropachev: Done, please check.

Lorak-mmk (approved):

Lorak-mmk:
Its better now, at least integration tests will be runnable without problems.

PR #412, opened 2026-02-02
Fix index out of bounds panic when map collection has odd items

### LINKED ISSUE #411: Index out of bounds error when iterating collections with 3 elements
## Bug Description

When using the cpp-rs-driver to read collections (lists, sets, maps) with 3 elements, an index out of bounds panic occurs in the collection iteration logic.

## Error Message

```
index out of bounds: the len is 3 but the index is 3
```

Location: `src/cql_types/collection.rs:164:49`

## Root Cause

This appears to be an off-by-one error in the collection parsing/iteration logic. When iterating through collection elements, the code accesses an index equal to the length (3) instead of the valid range `[0, 1, 2]` (i.e., `0..length`).

## How to Reproduce

The error occurs when reading CQL collection types with exactly 3 elements. We encountered this while running a benchmarking workload with collection types including:

- `list<int>` with 3 elements (e.g., `[1, 2, 3]`)
- `set<int>` with 3 elements
- `frozen<list<int>>` with 3 elements
- Nested collections like `frozen<list<frozen<list<int>>>>`

Example CQL schema that triggers the bug:

```cql
CREATE TABLE collections (
    pk bigint PRIMARY KEY,
    col_list_int list<int>,
    col_frozen_list_int frozen<list<int>>,
    col_list_of_lists frozen<list<frozen<list<int>>>>
);

INSERT INTO collections (pk, col_list_int, col_frozen_list_int, col_list_of_lists)
VALUES (1, [1, 2, 3], [1, 2, 3], [[1, 2], [3, 4], [5, 6]]);

SELECT * FROM collections WHERE pk = 1;
```

The error occurs during the SELECT operation when the driver tries to parse the collection values from the result.

## Environment

- cpp-rs-driver (latest)
- Used via Latte benchmarking tool's driver adapter system

## Expected Behavior

Collections with any number of elements should be iterated correctly without panicking.

## Additional Context

This was discovered while testing driver adapters for the [Latte](https://github.com/scylladb/latte) benchmarking tool. The adapter code correctly passes collection data, but the underlying cpp-rs-driver library panics when parsing the response.

The specific test workload includes various collection types (lists, sets, maps) with different nesting levels, all using 2-3 elements per collection.

dkropachev:
## Summary
- Fixes index out of bounds panic in map collection to `CassCqlValue` conversion when a map has an odd number of items (missing value in a key-value pair)
- Changes `From<&CassCollection>` to `TryFrom<&CassCollection>` for `CassCqlValue`, returning `CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE` instead of panicking
- Updates `binding.rs` to use `try_into()` to propagate the error to callers
- Logs an error via `tracing::error!` when an odd-item map is encountered
- Adds `Vec::with_capacity` for map pair allocation
- Adds unit test covering both odd (error) and even (success) item counts

Fixes #411

## Test plan
- [x] Run `cargo test collection::` to verify all collection tests pass
- [x] Run `make check-cargo-clippy check-cargo-fmt` to verify linting passes

### DIFF
--- a/scylla-rust-wrapper/src/cql_types/collection.rs
+++ b/scylla-rust-wrapper/src/cql_types/collection.rs
@@ -148,34 +148,41 @@ impl CassCollection {
     }
 }
 
-impl From<&CassCollection> for CassCqlValue {
-    fn from(collection: &CassCollection) -> Self {
+impl TryFrom<&CassCollection> for CassCqlValue {
+    type Error = CassError;
+
+    fn try_from(collection: &CassCollection) -> Result<Self, Self::Error> {
         // FIXME: validate that collection items are correct
         let data_type = collection.data_type.clone();
         match collection.collection_type {
-            CollectionType::List => CassCqlValue::List {
+            CollectionType::List => Ok(CassCqlValue::List {
                 data_type,
                 values: collection.items.clone(),
-            },
+            }),
             CollectionType::Map => {
-                let mut grouped_items = Vec::new();
-                // FIXME: validate even number of items
-                for i in (0..collection.items.len()).step_by(2) {
-                    let key = collection.items[i].clone();
-                    let value = collection.items[i + 1].clone();
-
-                    grouped_items.push((key, value));
+                let (pairs, remainder) = collection.items.as_chunks::<2>();
+                if !remainder.is_empty() {
+                    tracing::error!(
+                        "Map collection has an odd number of items ({}). Maps must have an even number of items (key-value pairs).",
+                        collection.items.len()
+                    );
+                    return Err(CassError::CASS_ERROR_LIB_INVALID_ITEM_COUNT);
                 }
+                let grouped_items = pairs
+                    .iter()
+                    .cloned()
+                    .map(|[key, value]| (key, value))
+                    .collect();
 
-                CassCqlValue::Map {
+                Ok(CassCqlValue::Map {
                     data_type,
                     values: grouped_items,
-                }
+                })
             }
-            CollectionType::Set => CassCqlValue::Set {
+            CollectionType::Set => Ok(CassCqlValue::Set {
                 data_type,
                 values: collection.items.clone(),
-            },
+            }),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


scylla-rust-wrapper/src/cql_types/collection.rs
```diff
@@ -157,12 +157,21 @@ impl From<&CassCollection> for CassCqlValue {
...
+                if items_len % 2 != 0 {
+                    tracing::warn!(
+                        "Map collection has an odd number of items ({}). The last unpaired key will be ignored.",
+                        items_len
+                    );
+                }
+
+                let pair_count = items_len / 2;
+                let mut grouped_items = Vec::with_capacity(pair_count);
+                for i in (0..pair_count * 2).step_by(2) {
```
Lorak-mmk: If map has odd number of items, it is definitely a malformed value.

We should not ignore it - error should be returned from such conversions.

src/testing.cpp
Lorak-mmk, line 1: Changes in `src/testing.cpp`, `src/testing.hpp` and `src/testing_rust_impls.h` seem unrelated to this PR.

Lorak-mmk (changes requested):

src/testing.cpp
dkropachev: it was my attempt to make cicd green

scylla-rust-wrapper/src/cql_types/collection.rs
```diff
@@ -580,4 +591,83 @@ mod tests {
...
+            let result: Result<CassCqlValue, CassError> = collection_ref.try_into();
+            assert!(
+                result.is_err(),
+                "Should return an error for odd number of items"
+            );
+            assert_eq!(
+                result.unwrap_err(),
+                CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE,
+                "Should return CASS_ERROR_LIB_INVALID_VALUE_TYPE"
+            );
```
wprzytula: :wrench: Let's use `assert_matches`:
```rust
assert_matches::assert_matches!(result, Err(CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE));
```

wprzytula (changes requested):

scylla-rust-wrapper/src/cql_types/collection.rs
dkropachev: done

dkropachev:
Regarding the review comment about returning an error instead of ignoring odd-item maps:

**Done.** Changed the implementation to return an error:

1. Changed `impl From<&CassCollection>` to `impl TryFrom<&CassCollection>` with `type Error = CassError`
2. Added validation that returns `Err(CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE)` when the map has an odd number of items
3. Updated `binding.rs` to use `try_into().map(Some)` to properly propagate the error
4. Added unit test that verifies the error is returned for odd-item maps

wprzytula (approved):

wprzytula:
@dkropachev Please update the cover letter.

dkropachev:
> @dkropachev Please update the cover letter.

done

dkropachev:
@Lorak-mmk , could you please review it ?

dkropachev:
Just in case, cicd failure not related to this PR.

scylla-rust-wrapper/src/cql_types/collection.rs
```diff
@@ -147,34 +147,45 @@ impl CassCollection {
...
-                // FIXME: validate even number of items
-                for i in (0..collection.items.len()).step_by(2) {
+                let items_len = collection.items.len();
+                // Maps require key-value pairs, so must have an even number of items.
+                if items_len % 2 != 0 {
+                    tracing::error!(
+                        "Map collection has an odd number of items ({}). Maps must have an even number of items (key-value pairs).",
+                        items_len
+                    );
+                    return Err(CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE);
```
Lorak-mmk: Maybe a better error would be `CASS_ERROR_LIB_INVALID_ITEM_COUNT`?

Lorak-mmk: This is also what original cpp-driver does:
```
CassError AbstractData::set(size_t index, const Collection* value) {
  CASS_CHECK_INDEX_AND_TYPE(index, value);
  if (value->type() == CASS_COLLECTION_TYPE_MAP && value->items().size() % 2 != 0) {
    return CASS_ERROR_LIB_INVALID_ITEM_COUNT;
  }
  elements_[index] = value;
  return CASS_OK;
}
```

scylla-rust-wrapper/src/cql_types/collection.rs
```diff
@@ -147,34 +147,45 @@ impl CassCollection {
...
+                }
+
+                let pair_count = items_len / 2;
+                let mut grouped_items = Vec::with_capacity(pair_count);
+                for i in (0..pair_count * 2).step_by(2) {
```
Lorak-mmk: This thing can be written in a much more idiomatic way, without index-based slice access.
```rust
            CollectionType::Map => {
                let (pairs, remainder) = collection.items.as_chunks::<2>();
                if !remainder.is_empty() {
                    tracing::error!(
                        "Map collection has an odd number of items ({}). Maps must have an even number of items (key-value pairs).",
                        collection.items.len()
                    );
                    return Err(CassError::CASS_ERROR_LIB_INVALID_VALUE_TYPE);
                }
                let grouped_items = pairs
                    .into_iter()
                    .cloned()
                    .map(|[key, value]| (key, value))
                    .collect();

                Ok(CassCqlValue::Map {
                    data_type,
                    values: grouped_items,
                })
            }
```

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/cql_types/collection.rs
dkropachev: done

scylla-rust-wrapper/src/cql_types/collection.rs
dkropachev: good call, done.

Lorak-mmk: Not done, you did something else. Why didn't you just use the snippet above, and instead done something with `.chunks_exact(2)`? Your version still needs index-based access to each group, and you still need to perform the modulo check manually.

dkropachev: Your code throws:
```
 error: current MSRV (Minimum Supported Rust Version) is `1.85.0` but this item is stable since `1.88.0`
     --> src/cql_types/collection.rs:162:59
      |
  162 |                 let (pairs, remainder) = collection.items.as_chunks::<2>();
      |                                                           ^^^^^^^^^^^^^^^^
```

That is why i change adopted it to `1.85.0` style.

Lorak-mmk: It is fine to bump our MSRV to 1.88, let's do that.

Lorak-mmk (approved):

wprzytula (approved):

PR #414, opened 2026-02-02
Fix flaky unit tests due to port conflicts

### LINKED ISSUE #413: Flaky unit tests failing due to port conflicts (Address already in use)
## Description

Three unit tests are consistently failing due to network port conflicts ("Address already in use"). These tests attempt to bind to specific ports that may already be in use on the system.

## Failing Tests

1. `runtime::tests::runtime_is_never_dropped_in_async_context`
2. `session::tests::session_clones_and_freezes_exec_profiles_mapping`
3. `session::tests::session_resolves_exec_profile_on_first_query`

## Error Messages

```
thread '<unnamed>' panicked at src/testing/utils.rs:136:10:
called `Result::unwrap()` on an `Err` value: DriverConnectionAttempt(127.0.16.147:9042, Os { code: 98, kind: AddrInUse, message: "Address already in use" })
```

```
thread 'runtime::tests::runtime_is_never_dropped_in_async_context' panicked at rusty-fork-0.3.0/src/fork_test.rs:158:9:
child exited unsuccessfully with signal: 6 (SIGABRT) (core dumped)
```

## How to Reproduce

```bash
make run-test-unit
```

Or:

```bash
cd scylla-rust-wrapper
RUSTFLAGS="--cfg cpp_rust_unstable --cfg cpp_integration_testing" cargo test
```

## Expected Behavior

Tests should either:
- Use dynamic port allocation to avoid conflicts
- Properly handle port-in-use errors with retries or fallback ports
- Be marked as requiring specific test environment setup

## Environment

- Tests fail on both the `master` branch and feature branches
- Failure is consistent across multiple runs
- Appears to be related to hardcoded IP addresses in the 127.0.16.x range

dkropachev:
## Summary
- Add retry logic to `test_with_one_proxy()` to handle "Address already in use" errors
- Proxy binding now attempts up to 10 different addresses before failing
- Logs retry attempts for debugging visibility

## Test plan
- Run unit tests multiple times to verify the fix reduces flakiness: `make run-test-unit`
- The fix handles race conditions where `scylla_proxy::get_exclusive_local_address()` returns addresses that are already bound

Fixes #413

### DIFF
--- a/scylla-rust-wrapper/src/testing/utils.rs
+++ b/scylla-rust-wrapper/src/testing/utils.rs
@@ -1,6 +1,6 @@
 use scylla_proxy::{
-    Condition, Node, Proxy, Reaction as _, RequestFrame, RequestOpcode, RequestReaction,
-    RequestRule, ResponseFrame, RunningProxy,
+    Condition, DoorkeeperError, Node, Proxy, Reaction as _, RequestFrame, RequestOpcode,
+    RequestReaction, RequestRule, ResponseFrame, RunningProxy,
 };
 use std::{
     collections::HashMap,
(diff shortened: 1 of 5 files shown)


scylla-rust-wrapper/src/testing/utils.rs
```diff
@@ -144,11 +144,67 @@ pub(crate) async fn test_with_one_proxy_at_ip(
...
+                    .expect("Test thread panicked");
+
+                let _ = proxy.finish().await;
+                return;
+            }
+            Err(e) => {
+                // Check if this is an address-in-use error
+                let is_addr_in_use = e.to_string().contains("Address already in use")
+                    || e.to_string().contains("AddrInUse");
+
```
Lorak-mmk: This is not the correct way to check errors. We (and Rust in general) use structured errors specifically to enable such error handling. 
What should be done: 
- Match on `e`, which is `DoorkeeperError`. Check which variant /variants are used in the situation you are trying to fix. It will probably be one of / many of `Listen`, `SocketCreate`, `SocketBind`.
- All of the above store `io::Error`. Call its `.kind()` method, and check that the result is [AddrInUse](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.AddrInUse)

scylla-rust-wrapper/src/testing/utils.rs
```diff
@@ -241,7 +241,7 @@ pub(crate) async fn test_with_one_proxy(
 /// It is also possible to specify a timeout which is applied to the test in
 /// the block, by adding the following line straight before the test:
 ///
-/// ```rust,no_compile
+/// ```ignore
```
Lorak-mmk: iirc `ignore` will skip checking the block at all, and will remove syntax highlighting from the block.  
rustdoc book says its almost never a right thing to do: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

Maybe a better fix is to use no_run, and make the examples compilable?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/testing/utils.rs
dkropachev: is it better now ?

scylla-rust-wrapper/src/testing/utils.rs
dkropachev: great point, fixed

scylla-rust-wrapper/src/testing/utils.rs
Lorak-mmk: Yes, great

dkropachev:
@Lorak-mmk , @wprzytula , can we merge it ?

wprzytula:
> @Lorak-mmk , @wprzytula , can we merge it ?

Not sure yet. Can it wait until I'm back?

dkropachev:
> > @Lorak-mmk , @wprzytula , can we merge it ?
> 
> Not sure yet. Can it wait until I'm back?

Sure, no need to do anything while you are on PTO

Lorak-mmk (changes requested):
Last two commits are modifying the older commits (they address review comments). Instead please modify the original commits instead of adding new ones.

Apart from that the PR looks fine to me.

dkropachev:
@Lorak-mmk , @wprzytula , done, please take a look.

dkropachev:
CICD failures not related to this PR, it is a heartbeat test that fails

Lorak-mmk (changes requested):
Commit `CI: Replace fedora-41 with fedora-43 in package build matrix` modifies `scylla-rust-wrapper/src/testing/utils.rs`.
Code changes probably belong to first commit. Doc comment changes should be in a separate commit. So, in other words my previous feedback was not really addressed. The fixups are now not in a separate commit, but in an unrelated commit.

dkropachev:
> Commit `CI: Replace fedora-41 with fedora-43 in package build matrix` modifies `scylla-rust-wrapper/src/testing/utils.rs`. Code changes probably belong to first commit. Doc comment changes should be in a separate commit. So, in other words my previous feedback was not really addressed. The fixups are now not in a separate commit, but in an unrelated commit.

Indeed, i pushed something wrong, check again please.

Lorak-mmk (approved):
Looks good to me, can be merged when CI finishes.

PR #417, opened 2026-02-16
Stabilize integration tests: fix flaky heartbeat test

dkropachev:
## Summary
- **test_heartbeat**: Replace flaky connection-count assertion with direct heartbeat timeout log check. The original test's `total_connections` metric could fluctuate due to reconnection attempts, and `execute_async` calls could interfere with idle-connection heartbeat detection.

## Pre-review checklist

- [x] I have enabled appropriate tests in `Makefile` in `{SCYLLA,CASSANDRA}_(NO_VALGRIND_)TEST_FILTER`.

### DIFF
--- a/tests/src/integration/tests/test_heartbeat.cpp
+++ b/tests/src/integration/tests/test_heartbeat.cpp
@@ -78,7 +78,7 @@ CASSANDRA_INTEGRATION_TEST_F(HeartbeatTests, HeartbeatDisabled) {
  *
  * @since 2.1.0
  * @jira_ticket CPP-152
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
@dkropachev I think this is no longer relevant after https://github.com/scylladb/cpp-rs-driver/pull/424 right? I'm closing, feel free to reopen if I'm mistaken.

dkropachev:
> @dkropachev I think this is no longer relevant after https://github.com/scylladb/cpp-rs-driver/pull/424 right? I'm closing, feel free to reopen if I'm mistaken.

Correct, it should be good now

PR #418, opened 2026-02-23
Bump rust driver to 1.5.0 and bump other dependencies

wprzytula:
Rust Driver is bumped to 1.5.0 (the newest released).
Other dependencies are bumped using `cargo update`.

This prepares for the 1.0.0 release.

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/Cargo.toml
+++ b/scylla-rust-wrapper/Cargo.toml
@@ -11,7 +11,7 @@ license = "MIT OR Apache-2.0"
 rust-version = "1.85"
 
 [dependencies]
-scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "4f02969823f7b07b0d6b007a11863fe4aac95821", features = [
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver.git", rev = "v1.5.0", features = [
     "openssl-010",
     "metrics",
     "unstable-cpp-rs",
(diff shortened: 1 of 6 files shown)


scylla-rust-wrapper/Cargo.toml
Lorak-mmk: Don't you need to change version here as well?

scylla-rust-wrapper/src/session.rs
```diff
@@ -984,11 +984,11 @@ mod tests {
...
-                    Condition::RequestOpcode(RequestOpcode::Query)
-                        .or(Condition::RequestOpcode(RequestOpcode::Batch))
-                        .and(Condition::BodyContainsCaseInsensitive(Box::new(
-                            *b"INSERT INTO system.",
-                        ))),
+                    Condition::any([
+                        Condition::RequestOpcode(RequestOpcode::Query),
+                        Condition::RequestOpcode(RequestOpcode::Batch),
+                    ])
+                    .and(Condition::not(Condition::ConnectionRegisteredAnyEvent)),
```
Lorak-mmk, line 991: Why `*b"INSERT INTO system.",` was removed? It is not something that CC would do.

scylla-rust-wrapper/tests/integration/session.rs
```diff
@@ -80,13 +80,14 @@ fn retry_policy_on_statement_and_batch_is_handled_properly_rules()
...
-                    *b"SELECT host_id FROM system.",
-                )))
-                // this 1 differentiates Fallthrough and Default retry policies.
-                .and(Condition::TrueForLimitedTimes(1)),
+            Condition::any([
+                Condition::RequestOpcode(RequestOpcode::Query),
+                Condition::RequestOpcode(RequestOpcode::Batch),
+                Condition::RequestOpcode(RequestOpcode::Prepare),
+            ])
+            .and(Condition::not(Condition::ConnectionRegisteredAnyEvent))
```
Lorak-mmk, line 88: ditto for `b"SELECT host_id FROM system.",`

scylla-rust-wrapper/src/session.rs
wprzytula: The logic there was just to rule out the CC. On non-control connections, only `"INSERT INTO system."` statement is executed. Now that we have `Condition::ConnectionRegisteredAnyEvent`, it makes more sense (and it's more robust) to use it instead.

Lorak-mmk (approved):

PR #423, opened 2026-02-25
IT: fix flaky ExecutionProfileTest.RequestTimeout test

### LINKED ISSUE #309: ExecutionProfileTest.Integration_Cassandra_RequestTimeout can fail
```
/home/runner/work/cpp-rust-driver/cpp-rust-driver/tests/src/integration/tests/test_exec_profile.cpp:279: Failure
      Expected: CASS_ERROR_LIB_REQUEST_TIMED_OUT
      Which is: 16777230
To be equal to: result.error_code()
      Which is: 33563136
[  FAILED  ] ExecutionProfileTest.Integration_Cassandra_RequestTimeout (1362 ms)
```

dkropachev:
## Summary
- Increase sleep time from 2ms to 10ms in the `ExecutionProfileTest.RequestTimeout` test to ensure the 1ms request timeout reliably fires before the server can respond
- The previous 2ms margin was too tight, causing intermittent failures where the server returned `CASS_ERROR_SERVER_INVALID_QUERY` (0x2002200) before the client-side timeout triggered

Fixes #309

### DIFF
--- a/tests/src/integration/tests/test_exec_profile.cpp
+++ b/tests/src/integration/tests/test_exec_profile.cpp
@@ -271,13 +271,13 @@ CASSANDRA_INTEGRATION_TEST_F(ExecutionProfileTest, RequestTimeout) {
   Batch batch;
   batch.add(statement);
   batch.set_execution_profile("request_timeout");
-  batch.set_sleep_time(2); // Simulate >=2ms latency
+  batch.set_sleep_time(10); // Simulate >=10ms latency (well above 1ms timeout)
   result = session_.execute(batch, false);
   ASSERT_EQ(CASS_ERROR_LIB_REQUEST_TIMED_OUT, result.error_code());
 
   // Execute a simple query with assigned profile (should timeout)
   statement.set_execution_profile("request_timeout");
-  statement.set_sleep_time(2); // Simulate >=2ms latency
+  statement.set_sleep_time(10); // Simulate >=10ms latency (well above 1ms timeout)
   result = session_.execute(statement, false);
   ASSERT_EQ(CASS_ERROR_LIB_REQUEST_TIMED_OUT, result.error_code());
 }


wprzytula (approved):

tests/src/integration/tests/test_exec_profile.cpp
```diff
@@ -271,13 +271,13 @@ CASSANDRA_INTEGRATION_TEST_F(ExecutionProfileTest, RequestTimeout) {
...
-  batch.set_sleep_time(2); // Simulate >=2ms latency
+  batch.set_sleep_time(100); // Simulate >=100ms latency (well above 1ms timeout)
   result = session_.execute(batch, false);
   ASSERT_EQ(CASS_ERROR_LIB_REQUEST_TIMED_OUT, result.error_code());
 
   // Execute a simple query with assigned profile (should timeout)
   statement.set_execution_profile("request_timeout");
-  statement.set_sleep_time(2); // Simulate >=2ms latency
+  statement.set_sleep_time(100); // Simulate >=100ms latency (well above 1ms timeout)
   result = session_.execute(statement, false);
```
Lorak-mmk, line 281: Do we need this much sleep? If timeout is 1ms, and test was flaky at 2ms sleep, then 10ms should be more than enough, right?

wprzytula: This sleep doesn't cause the delay in test IIUC @Lorak-mmk, because the driver times out the request quicker.

wprzytula: Or I'm wrong, because the sleep is synchronous in the driver...

wprzytula: Yep, I'm wrong. Indeed sleep of 10ms should suffice.

Lorak-mmk: It uses our `SleepingHistoryListener`, which uses `std::thread::sleep` in `log_attempt_start`. So yes, it will delay the test.

dkropachev: dropped to 10

Lorak-mmk (approved):

PR #429, opened 2026-02-26
Optimization for request result

### LINKED ISSUE #244: Don't use `Yoke` to represent self-borrowing in the first row of `CassRowsResult`
Lazy deserialization [PR](https://github.com/scylladb/cpp-rust-driver/pull/236) introduce one (probably) needless `Arc` allocation for `CassRowsResultSharedData`.

## Why this `Arc` is introduced?
There is currently no way to represent a self-borrowing relationship in safe Rust. This is why, we employed a [`Yoke`](https://docs.rs/yoke/latest/yoke/struct.Yoke.html) solution, which is safe.

TLDR; the `Arc` is required, because of the `C: StableDeref` constraint [here](https://docs.rs/yoke/latest/yoke/struct.Yoke.html#impl-Yoke%3CY,+C%3E).

However, we have some additional type and ownership guarantees that we cannot provide to the `yoke` crate (these guarantees cannot be expressed using the types that the crate provides). We could probably remove the `yoke` dependency and implement the self-borrowing ourselves, basing on the aforementioned guarantees. This, however, would require us to introduce some `unsafe` code which would need to be justified and well-documented.

Ref (comment/discussion): https://github.com/scylladb/cpp-rust-driver/pull/236#discussion_r2033546622

Lorak-mmk:
This PR performs 2 optimizations related to request results.

## CassRow allocation

Allocation for `Vec` in `CassRow` is reused when advancing `CassResultIterator`. Before we always dropped old `CassRow` and created new one, resulting in 1 `free` + 1 `malloc` call.
Now we take the previous `Vec`, clear it, and reuse it.
I verified with the benchmark (second commit) that allocations are really gone, and they are! Now we have no obligatory allocations per-row, only per-request.
There was no visible performance improvement in my simple test. This is most likely because libc probably optimizes a pattern of `free` + `malloc` with the same size, making them really cheap.
This may not be the case for other allocators, or in more complex scenarios (like multi-threaded programs), so I still think this opt makes sense

## Removed Arc for shared data in CassRowsResult

See the issue about this: https://github.com/scylladb/cpp-rs-driver/issues/244
This is solved by making the self-referential struct manually, without Yoke.
Unfortunately I see no way to apply it. I also see no good way to make the resulting code much more secure, or easy to understand.
I guess this is a tradeoff - allocation gone, in exchange for more dangerous and complex code.

I'll gladly accept suggestions for how to make it safer, better contained, simpler etc.
If reviewers think this tradeoff makes no sense, I can also drop this part of PR, and close the issue as `wontfix`.

Fixes: https://github.com/scylladb/cpp-rs-driver/issues/244

## Pre-review checklist

### DIFF
--- a/examples/result_iteration_bench/result_iteration_bench.c
+++ b/examples/result_iteration_bench/result_iteration_bench.c
@@ -1,0 +1,435 @@
+/*
+  This is free and unencumbered software released into the public domain.
+
+  Anyone is free to copy, modify, publish, use, compile, sell, or
+  distribute this software, either in source code form or as a compiled
+  binary, for any purpose, commercial or non-commercial, and by any
+  means.
+
+  In jurisdictions that recognize copyright laws, the author or authors
+  of this software dedicate any and all copyright interest in the
+  software to the public domain. We make this dedication for the benefit
+  of the public at large and to the detriment of our heirs and
+  successors. We intend this dedication to be an overt act of
+  relinquishment in perpetuity of all present and future rights to this
+  software under copyright law.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+  EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+  MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+  IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+  OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+  ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+  OTHER DEALINGS IN THE SOFTWARE.
+
+  For more information, please refer to <http://unlicense.org/>
+*/
+
+/*
+ * Benchmark for result iteration performance.
+ *
+ * Creates a table with 10 columns, inserts 10000 rows, then performs
+ * an unpaged SELECT * and measures how long it takes to iterate over
+ * all rows in the result set. The iteration is repeated multiple times
+ * to get stable measurements.
+ */
+
+#define _GNU_SOURCE
+#include <assert.h>
+#include <dlfcn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include "cassandra.h"
+
+/*
+ * Allocation tracking via interposition.
+ * We override malloc/realloc/free to count calls, forwarding
+ * to the real implementations via dlsym(RTLD_NEXT, ...).
+ */
+
+static size_t g_malloc_count = 0;
+static size_t g_realloc_count = 0;
+static size_t g_free_count = 0;
+
+typedef void* (*real_malloc_t)(size_t);
+typedef void* (*real_realloc_t)(void*, size_t);
+typedef void (*real_free_t)(void*);
+
+static real_malloc_t real_malloc = NULL;
+static real_realloc_t real_realloc = NULL;
+static real_free_t real_free = NULL;
+
+/* Small static buffer to bootstrap dlsym (which may call malloc internally). */
+static char bootstrap_buf[4096];
+static size_t bootstrap_used = 0;
+static int resolving = 0;
+
+static void ensure_real_funcs(void) {
+  if (real_malloc != NULL) return;
+  resolving = 1;
+  real_malloc = (real_malloc_t)dlsym(RTLD_NEXT, "malloc");
+  real_realloc = (real_realloc_t)dlsym(RTLD_NEXT, "realloc");
+  real_free = (real_free_t)dlsym(RTLD_NEXT, "free");
+  resolving = 0;
+}
+
+void* malloc(size_t size) {
+  if (resolving) {
+    /* dlsym is resolving symbols and called malloc - use static buffer. */
+    size_t aligned = (size + 7) & ~(size_t)7; /* align to 8 */
+    if (bootstrap_used + aligned > sizeof(bootstrap_buf)) {
+      return NULL;
+    }
+    void* ptr = bootstrap_buf + bootstrap_used;
+    bootstrap_used += aligned;
+    return ptr;
+  }
+  ensure_real_funcs();
+  g_malloc_count++;
+  return real_malloc(size);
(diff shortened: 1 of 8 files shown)


scylla-rust-wrapper/src/query_result.rs
```diff
@@ -179,8 +179,16 @@ impl<'result> CassRow<'result> {
...
     ) -> Result<Self, DeserializationError> {
-        let mut columns = Vec::with_capacity(row.columns.columns_remaining());
+        let mut columns = match old_row {
+            Some(row) => {
+                let mut old_vec = row.columns;
+                old_vec.clear();
+                old_vec
+            }
+            None => Vec::with_capacity(row.columns.columns_remaining()),
+        };
```
wprzytula, line 296: ♻️ After the change, `row` is once used as `CassRow` and once as `CassRawRow`, in different `match` branches. Please rename one to `raw_row`.

examples/result_iteration_bench/result_iteration_bench.c
```diff
@@ -0,0 +1,418 @@
...
+  real_free = (real_free_t)dlsym(RTLD_NEXT, "free");
+  resolving = 0;
+}
+
+void* malloc(size_t size) {
+  if (resolving) {
+    /* dlsym is resolving symbols and called malloc - use static buffer. */
+    void* ptr = bootstrap_buf + bootstrap_used;
+    bootstrap_used += (size + 7) & ~(size_t)7; /* align to 8 */
+    return ptr;
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk WDYT?

examples/result_iteration_bench/result_iteration_bench.c
```diff
@@ -0,0 +1,418 @@
...
+void* realloc(void* ptr, size_t size) {
+  if (resolving) {
+    void* new_ptr = bootstrap_buf + bootstrap_used;
+    bootstrap_used += (size + 7) & ~(size_t)7;
+    return new_ptr;
+  }
+  ensure_real_funcs();
+  g_realloc_count++;
+  return real_realloc(ptr, size);
+}
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk WDYT?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -38,9 +38,49 @@ pub(crate) enum CassResultKind {
...
+    /// SAFETY: borrows from `shared_data`. See [`SelfBorrowingFirstRow`] docs.
+    first_row: Option<SelfBorrowingFirstRow>,
+}
+
+impl Drop for CassRowsResult {
+    fn drop(&mut self) {
+        // Drop first_row before shared_data, because it borrows from shared_data.
+        if let Some(row) = self.first_row.take() {
+            drop(ManuallyDrop::into_inner(row.0));
+        }
```
wprzytula: 🔧 Pattern matching is more readable.
```rust
        if let Some(SelfBorrowingFirstRow(row)) = self.first_row.take() {
            drop(ManuallyDrop::into_inner(row));
        }
```

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -38,9 +38,49 @@ pub(crate) enum CassResultKind {
...
+///
+/// # Safety invariants
+///
+/// 1. `CassRowsResult` is always stored inside `Arc<CassResult>`,
+///    so the address of `CassRowsResultSharedData` is stable.
+/// 2. `CassRowsResult::Drop` drops this field before `shared_data`.
+/// 3. Access is only through [`row()`](Self::row), which shortens the
+///    lifetime to `&self`, preventing the reference from escaping.
+#[derive(Debug)]
+struct SelfBorrowingFirstRow(ManuallyDrop<CassRow<'static>>);
```
wprzytula, line 69: ❓ Why is `ManuallyDrop` required here?

scylla-rust-wrapper/src/query_result.rs
```diff
@@ -125,6 +182,48 @@ impl CassResult {
...
+                // This erases the lifetime, which is safe because:
+                // 1. shared_data lives inside Arc<CassResult> (stable address).
+                // 2. CassRowsResult::Drop drops first_row before shared_data.
+                // 3. SelfBorrowingFirstRow::row() shortens the lifetime back
+                //    on access, preventing the reference from escaping.
+                let row_static: CassRow<'static> = unsafe { std::mem::transmute(row) };
+                Some(SelfBorrowingFirstRow(ManuallyDrop::new(row_static)))
+            }
+            None => None,
+        };
```
wprzytula: ♻️ Use `.map()` instead of the verbose `match` here.

wprzytula (changes requested):

scylla-rust-wrapper/src/query_result.rs
Lorak-mmk: In `CassRowsResult` you have 2 fields: `shared_data: CassRowsResultSharedData`, and `first_row: Option<SelfBorrowingFirstRow>` that borrows from `shared_data` (in the way that borrow checked doesn't understand).
When dropping `CassRowsResult`, if you drop `shared_data` first, then references in `first_row` would point to freed data, which, as far as I understand, is immediate ub even if we don't use those references.

We could put `first_row` on top of the struct, because fields are dropped in order, but this deemed fragile by Rustonomicon, and ManuallyDrop is recommended instead: https://doc.rust-lang.org/nomicon/dropck.html?highlight=ManuallyDrop#a-related-side-note-about-drop-order

Lorak-mmk:
Addressed review comments.

Lorak-mmk:
I don't see the point of spending my time on perfecting code of this benchmark. It works, it shows the fix reduces the allocations. 
I can keep it as-is, or remove it from this PR, @wprzytula lmk what I should do.

wprzytula:
> I don't see the point of spending my time on perfecting code of this benchmark. It works, it shows the fix reduces the allocations. 
> I can keep it as-is, or remove it from this PR, @wprzytula lmk what I should do.

It's good enough for a benchmark.

scylla-rust-wrapper/src/query_result.rs
wprzytula: > We could put `first_row` on top of the struct, because fields are dropped in order, but this deemed fragile by Rustonomicon, and ManuallyDrop is recommended instead: https://doc.rust-lang.org/nomicon/dropck.html?highlight=ManuallyDrop#a-related-side-note-about-drop-order

This is what I was missing. Thanks.

wprzytula (approved):

PR #433, opened 2026-03-12
fix: missing whitespace in api docs

### LINKED ISSUE #430: API docs: No space between argument type and name, and return type / name of function
Image

I think the title and screenshot are pretty explanatory by themselves. 
In the frist declaration the function parameter looks like `CassBatchTypetype` and not `CassBatchType type`.
This first declaration is also missing a space between `CassBatch*` and `cass_batch_new`.

This is a problem in all signatures in API docs. cc @dgarcia360

dgarcia360:
Closes https://github.com/scylladb/cpp-rs-driver/issues/430


wprzytula (approved):
Now looks good!

PR #434, opened 2026-04-01
fix spelling for OpenSSL in cmake

MaikeR-TV:
Consistently use the same writing style for cmake module name, as used for the corresponding cmake variables.
Thus the cmake code will also work with generated cmake snippets from conan.

Renaming FindOpenSSL.cmake to FindOPENSSL.cmake accordingly.

Fixes: "Unable to Locate OpenSSL: Third party build step will be performed" when openssl is provided as conan dependency

## Pre-review checklist


dkropachev (approved):

dkropachev:
@MaikeR-TV , it needs to be fixed in other places too

MaikeR-TV:
I've fixed the failure for Linux by renaming the FindOpenSSL.cmake to FindOPENSSL.cmake accordingly.
macOS seemed to be having other issues.

dkropachev (changes requested):
it needs to be updated in other places too

MaikeR-TV:
> it needs to be updated in other places too

You mean mentions of the FindOPENSSL.cmake file [here](https://github.com/scylladb/cpp-rs-driver/blob/master/cmake/FindOpenSSL.cmake#L5) and [here](https://github.com/scylladb/cpp-rs-driver/blob/master/cmake/ExternalProject-OpenSSL.cmake#L78)?
Or something else I missed?

dkropachev:
@MaikeR-TV , thanks for bring it up and taking your time to fix it, i will take your changes and have some more on top of them in https://github.com/scylladb/cpp-rs-driver/pull/435

PR #437, opened 2026-04-01
ci: install Rust toolchain in macOS packages job

### LINKED ISSUE #436: CI: macOS packages build fails — Rust toolchain missing from macos-15-intel runner
## Summary

The `build-packages / macOS packages` CI job fails because the `macos-15-intel` GitHub Actions runner image no longer has Rust pre-installed.

## Error

```
CMake Error at CMakeLists.txt:14 (enable_language):
  Experimental Rust support is not enabled.

CMake Error: CMAKE_Rust_COMPILER not set, after EnableLanguage
```

The failure occurs at the very first `cmake` configure step, before any project code is reached.

## Reproduction

The failure is visible on PR #435 ([run log](https://github.com/scylladb/cpp-rs-driver/actions/runs/23854313334/job/69542889836)) and also affected the last master push run (run ID 23190177631, March 17 — macOS job was cancelled).

PR #432 (merged March 17) passed on an older runner image (`macos-15/20250313`), but the current image (`macos-15/20260330`) no longer includes Rust.

## Root Cause

The macOS job in `.github/workflows/build-cpack-packages.yml` only installs GNU make:

```yaml
macos:
  name: macOS packages
  runs-on: macos-15-intel
  steps:
    - uses: actions/checkout@v4
    - name: Install GNU make
      run: brew install make
    - name: Build and test macOS packages (PKG + DMG)
      run: gmake test-package-macos
```

Unlike the Linux RPM job, it does not explicitly install the Rust toolchain. It relied on Rust being bundled in the runner image, which is no longer the case.

## Suggested Fix

Add a Rust installation step before the build, similar to what the Linux RPM job does:

```yaml
- name: Install Rust toolchain
  run: |
    curl --proto "=https" --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
    echo "$HOME/.cargo/bin" >> $GITHUB_PATH
```

## Affected Jobs

- `build-packages / macOS packages` (build-cpack-packages.yml)

All other CI jobs (Linux DEB, Linux RPM, Windows, build+lint+unit tests) are unaffected.

dkropachev:
## Summary

Fix the macOS CI packages build failure by installing Rust explicitly and bypassing CMake's experimental Rust feature gate on CMake 4.0+.

## Context

The `build-packages / macOS packages` CI job began failing after the GitHub Actions runner image was updated from `macos-15/20250313` to `macos-15/20260330`. Two separate issues were uncovered:

1. **Missing Rust toolchain**: The new runner image no longer bundles Rust pre-installed. The Linux RPM job already handles this by installing via `rustup`, but the macOS job relied on it being pre-installed.

2. **CMake 4.0+ experimental Rust gate**: The macOS runner ships CMake 4.3.1 (via Homebrew), which includes built-in experimental Rust language support. When `enable_language(Rust)` is called, CMake checks `CMAKE_EXPERIMENTAL_RUST` *before* loading any custom modules. Without the correct version-specific UUID, CMake emits a fatal error. Since the UUID changes with every CMake release, hardcoding it is not sustainable.

## Changes

- **`.github/workflows/build-cpack-packages.yml`**: Added an `Install Rust toolchain` step to the macOS job using `rustup`, consistent with the Linux RPM job.
- **`CMakeLists.txt`**: Added a version guard around `enable_language(Rust)`. On CMake >= 4.0, the project now calls `find_package(Rust)` + `include(CMakeRustInformation)` directly instead of going through `enable_language(Rust)`. This is functionally identical because the project never uses CMake's native Rust compilation rules -- all Rust compilation is driven by Cargo through `CMakeCargo.cmake`. On CMake < 4.0, the existing `enable_language(Rust)` path is preserved.

## Testing

- `make check` passes locally (clang-format, cargo check, cargo clippy, cargo fmt)
- Full cmake configure + cargo build verified locally with CMake 4.2.0
- The fix will be validated by the CI pipeline on this PR itself

## Links

- Fixes: #436
- Failed run log: https://github.com/scylladb/cpp-rs-driver/actions/runs/23854313334/job/69542889836

### DIFF
--- a/.github/workflows/build-cpack-packages.yml
+++ b/.github/workflows/build-cpack-packages.yml
@@ -79,21 +79,14 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: Install Rust toolchain
-        run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Build and test RPM packages
-        run: |
-          source "$HOME/.cargo/env"
-          make test-package-rpm-native SCYLLA_HOST=scylla SKIP_DOCKER_COMPOSE=1
+        run: make test-package-rpm-native SCYLLA_HOST=scylla SKIP_DOCKER_COMPOSE=1
 
       - name: Collect artifacts
         if: inputs.save-artifacts
-        run: |
-          source "$HOME/.cargo/env"
-          make collect-package-artifacts
+        run: make collect-package-artifacts
 
       - uses: actions/upload-artifact@v4
         if: inputs.save-artifacts
@@ -108,10 +101,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Install Rust toolchain
-        run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Install GNU make
         run: brew install make
--- a/.github/workflows/build-cpack-packages.yml
+++ b/.github/workflows/build-cpack-packages.yml
@@ -108,6 +108,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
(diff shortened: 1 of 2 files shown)


.github/workflows/build-cpack-packages.yml
```diff
@@ -108,6 +108,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+
```
Lorak-mmk: We could use https://github.com/actions-rust-lang/setup-rust-toolchain which also sets up caching of Rust compilations. We already use it in Rust Driver.

Lorak-mmk (approved):

.github/workflows/build-cpack-packages.yml
dkropachev: Done

PR #445, opened 2026-04-29
Build system fixes - part 1

wprzytula:
This is the first part of the effort to make the build system more understandable, better maintainable, and usable.

This PR focuses on improving the CMakeCargo bridge. Individual commits describe the steps taken.
A lot of comments and helpful schemas are added.

## Pre-review checklist


scylla-rust-wrapper/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ endif()
 if(CASS_BUILD_STATIC)
   create_copy($<TARGET_FILE:scylla_cpp_driver> ${INSTALL_NAME_STATIC})
   add_library(scylla-cpp-driver_static STATIC IMPORTED GLOBAL)
-  add_dependencies(scylla-cpp-driver_static scylla-cpp-driver_static_target)
+  add_dependencies(scylla-cpp-driver_static ${INSTALL_NAME_STATIC}_copy)
   set_target_properties(scylla-cpp-driver_static PROPERTIES IMPORTED_LOCATION ${CMAKE_BINARY_DIR}/${INSTALL_NAME_STATIC})
 endif()
```
Lorak-mmk: Commit: "build: fix nonexistent dependency targets for imported libraries"

Where does this `_copy` comes from?

cmake/CMakeCargo.cmake
Lorak-mmk: Commit: " build: add comments to CMakeCargo.cmake explaining cargo_build function"

Do we need Android and Ios support? I'm pretty sure we don't and so this can be removed. If you did that in another commit, disregard that comment.

Lorak-mmk: Same commit, same comment. I don't think we need IOS support.

cmake/CMakeCargo.cmake
```diff
@@ -77,8 +123,20 @@ function(cargo_build)
     list(APPEND CARGO_ARGS "--profile" "dev")
   endif()
```
Lorak-mmk, line 215: Huh, I wanted to suggest that `debug` can be renamed to `dev` and then this ifology becomes just  something like `list(APPEND CARGO_ARGS "--profile ${LIB_BUILD_TYPE}")`, but it looks like Cargo uses `debug` folder for `dev` profile. I wonder what are the historical reasons for this weirdness.

cmake/CMakeCargo.cmake
```diff
@@ -77,8 +123,20 @@ function(cargo_build)
...
+  # The custom command runs `cargo build` whenever any .rs source file
+  # changes. The custom targets provide named handles that other CMake
+  # targets can depend on.
+  #
+  # We then wrap the raw cargo outputs in IMPORTED library targets so the
+  # rest of the CMake build can link against them using standard
+  # target_link_libraries() calls.
+
+  # Glob all Rust sources so the command re-runs when any of them change.
   file(GLOB_RECURSE LIB_SOURCES "*.rs")
```
Lorak-mmk: Maybe `Cargo.toml` and `Cargo.lock` should also be included?

cmake/CMakeCargo.cmake
```diff
@@ -1,3 +1,104 @@
...
+#  (3)   scylla_cpp_driver /              IMPORTED targets wrapping the raw cargo
+#         scylla_cpp_driver_shared        output paths. Exist solely so that
+#         (IMPORTED libraries)            create_copy can use $<TARGET_FILE:…>
+#                                         generator expressions to get the path.
+#
+#  (4)   _copy targets                    Copy + rename from Cargo's underscore
+#         (custom targets, ALL)           naming (libscylla_cpp_driver.so) to
+#                                         conventional C library naming
+#                                         (libscylla-cpp-driver.so.X.Y.Z) in
+#                                         the build root.
```
Lorak-mmk, line 92: Commit: "build: add dependency diagrams to CMakeCargo.cmake"

It sounds a bit inelegant that we need this `_copy`. I hope this is the layer that gets removed.

cmake/CMakeCargo.cmake
Lorak-mmk, line 1: Commit: build: simplify cargo_build by removing intermediate IMPORTED targets

I wonder what this change gives us. It is not really eliminating the layer, but changing it from the target to a variable with atypical scoping.
Why is it better?

Lorak-mmk (changes requested):
Great work!

cmake/CMakeCargo.cmake
wprzytula: Yep, this is a `debug` profile, yet it's turned on with `--dev` CLI option.

cmake/CMakeCargo.cmake
wprzytula: I'll happily remove those platforms :))

scylla-rust-wrapper/CMakeLists.txt
wprzytula: I think you answered yourself in a later comment. If not, let me know.

cmake/CMakeCargo.cmake
Lorak-mmk: Let's remove them

cmake/CMakeCargo.cmake
```diff
@@ -1,103 +1,257 @@
...
-      endif()
+  if(IOS)
+    # cargo-lipo handles universal (fat) binary builds on iOS.
+    set(CARGO_ARGS "lipo")
+  else()
+    set(CARGO_ARGS "rustc" "--crate-type" "${CARGO_CRATE_TYPE}")
+    list(APPEND CARGO_ARGS "--target" ${LIB_TARGET})
+    if (CMAKE_VERBOSE_MAKEFILE)
+      list(APPEND CARGO_ARGS "--verbose")
     endif()
```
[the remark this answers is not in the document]
wprzytula: Claude Opus has a different opinion, backed by my local tests:

> The reviewer is incorrect. --crate-type is a first-class cargo rustc option, not a rustc flag that needs -- separation. It was stabilized in Rust 1.64 (https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html) (September 2022) — see the cargo rustc docs (https://doc.rust-lang.org/cargo/commands/cargo-rustc.html#option-cargo-rustc--crate-type).

cmake/CMakeCargo.cmake
wprzytula: Agreed, reverted.

cmake/CMakeCargo.cmake
wprzytula: Opus explains this well:

> The _copy layer exists because Cargo outputs artifacts with underscored names (libscylla_cpp_driver.a) into its own target directory (build/scylla-rust-wrapper/x86_64-unknown-linux-gnu/release/), while the project needs hyphenated names (libscylla-cpp-driver_static.a) in the build root (build/). So the copy does two things: rename and relocate.
> Could we eliminate it? Let me think about the alternatives:
> 1. Tell Cargo to output with the right name — not possible. Cargo derives the artifact name from the crate name in Cargo.toml, and crate names can't contain hyphens.
> 2. Tell Cargo to output to the build root — we could set CARGO_TARGET_DIR to the build root, but then the artifact would still be at build/x86_64-unknown-linux-gnu/release/libscylla_cpp_driver.a, not build/libscylla-cpp-driver_static.a. Still need a rename.
> 3. Point IMPORTED targets directly at Cargo's output — skip the copy entirely, and have the final IMPORTED targets (scylla-cpp-driver_static, scylla-cpp-driver) point directly at the files inside Cargo's target dir. The symlinks for .so → .so.X.Y.Z would also point there. Install rules would rename on install.

> Option 3 is interesting. The copy layer currently exists so that build/libscylla-cpp-driver.so.X.Y.Z and its symlinks (.so.X, .so) live in a predictable location. If we skip the copy:
> - The IMPORTED targets would point into build/scylla-rust-wrapper/x86_64-unknown-linux-gnu/release/ — a platform-dependent path, but that's fine since it's only used internally during the build.
> - Symlinks would be created pointing at the file inside Cargo's output dir.
> - Install rules already handle the rename (RENAME argument or destination filename).

> The main downside: the build directory would no longer have the conveniently-named libscylla-cpp-driver.so at the top level for users who build but don't install. Some users (and CI scripts) may depend on finding the library there.

Lorak-mmk (approved):

cmake/CMakeCargo.cmake
```diff
@@ -1,103 +1,262 @@
...
+
+  add_custom_command(
+    OUTPUT ${LIB_OUTPUTS}
+    COMMAND ${CARGO_ENV_COMMAND} ${CARGO_EXECUTABLE} ARGS ${CARGO_ARGS}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS ${LIB_SOURCES}
+    COMMENT "running cargo rustc --crate-type ${CARGO_CRATE_TYPE}"
+  )
+
+  add_custom_target(${CARGO_NAME}_${CARGO_CRATE_TYPE}_target ALL DEPENDS ${LIB_FILE})
```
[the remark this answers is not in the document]
wprzytula: Fixed in the newest commit.

Lorak-mmk (approved):

Issue #301, opened 2025-05-14
Improve the API for cstring - rust string conversion and vice-versa


muzarski:
While we cannot make use of `FFI` trait API (we don't have any assumptions about the origin of the `char*` pointers provided by the user), we could probably utilize `CassPtr` to some extent. Thanks to that, we would be handling null pointers on a type level (represented as `None`).

I don't like current API for multiple reasons. One of them is that

```rust
pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static str> {
    if ptr.is_null() {
        return None;
    }
    std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) }).ok()
}
```

does not distinguish two scenarios:
1. pointer is null
2. provided bytes are non-utf8

In both of these scenarios, we return `None`.

egorsavkin-plb:
> std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) }).ok()

Oh, I've run into this line of code crashing when I passed a string view to `cass_row_get_column_by_name_n` . Same call on original cpp driver is working fine.

wprzytula:
> > std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) }).ok()
> 
> Oh, I've run into this line of code crashing when I passed a string view to `cass_row_get_column_by_name_n` . Same call on original cpp driver is working fine.

That's worrying.

1. Can you provide the calling code?
2. How did the code crash? What did you observe?
3. Any ideas why this crashed here and didn't crash on original cpp driver?

egorsavkin-plb:
I found out, it was a mistake on my side - I messed up with tying up the iterators and this was the consequence of it.

> > std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) }).ok()
> 
> Oh, I've run into this line of code crashing when I passed a string view to `cass_row_get_column_by_name_n` . Same call on original cpp driver is working fine.

PR #448, opened 2026-05-10
Fix handling of NULL strings in user-facing APIs 

### LINKED ISSUE #446: Fix `cass_collection_append_bytes()` crashing on nullptr
With the new driver, the function `cass_collection_append_bytes()` is crashing if a nullptr is passed for an empty element in a collection.
The old driver did not crash with the same code and use case.

wprzytula:
Fixes: #446
Ref: #301

#446 revealed issues with our handling of NULL string pointers in user-facing APIs. Together with Claude Opus, I examined vulnerable spots in the code:
1. Fixed UB in `binding.rs` - variations of the originally reported issue.
2. Fixed UB/panics in `cass_row_get_column_by_name[_n]` upon null name provided.
3. Fixed `ptr_to_cstr` to return `None` on NULL pointers instead of UB. This made behaviour consistent with `ptr_to_cstr_n`.
4. Refactored `ptr_to_cstr[_n]` to return `Result<&'static str, PtrToStrError>`, with `PtrToStrError` differentiating between null pointer provided and invalid UTF-8 encoding. Migrated all callers carefully.

### Further plans

I'm going to address #301 fully soon, by introducing borrowed `CassStr*` FFI types: #449

## Pre-review checklist

- ~~[ ] I have enabled appropriate tests in `Makefile` in `{SCYLLA,CASSANDRA}_(NO_VALGRIND_)TEST_FILTER`.

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -5,22 +5,41 @@ use std::marker::PhantomData;
 use std::os::raw::c_char;
 use std::ptr::NonNull;
 use std::sync::{Arc, Weak};
+use thiserror::Error;
 
-pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
-    if ptr.is_null() {
-        return None;
-    }
-    unsafe { CStr::from_ptr(ptr) }.to_str().ok()
+/// Error type for `ptr_to_cstr` and `ptr_to_cstr_n`, distinguishing
+/// a null pointer from invalid UTF-8 input.
+#[derive(Debug, Error)]
+pub enum PtrToStrError {
+    #[error("null string pointer")]
+    NullPointer,
+    #[error("invalid UTF-8: {0}")]
+    InvalidUtf8(std::str::Utf8Error),
 }
 
-pub unsafe fn ptr_to_cstr_n(ptr: *const c_char, size: size_t) -> Option<&'static str> {
+pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Result<&'static str, PtrToStrError> {
     if ptr.is_null() {
-        return None;
+        return Err(PtrToStrError::NullPointer);
     }
-    std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) }).ok()
+    unsafe { CStr::from_ptr(ptr) }
+        .to_str()
+        .map_err(PtrToStrError::InvalidUtf8)
 }
 
-pub(crate) unsafe fn arr_to_cstr<const N: usize>(arr: &[c_char]) -> Option<&'static str> {
+pub unsafe fn ptr_to_cstr_n(
+    ptr: *const c_char,
+    size: size_t,
+) -> Result<&'static str, PtrToStrError> {
+    if ptr.is_null() {
+        return Err(PtrToStrError::NullPointer);
+    }
+    std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) })
+        .map_err(PtrToStrError::InvalidUtf8)
+}
+
+pub(crate) unsafe fn arr_to_cstr<const N: usize>(
+    arr: &[c_char],
+) -> Result<&'static str, PtrToStrError> {
     let null_char = '\0' as c_char;
     let end_index = arr[..N].iter().position(|c| c == &null_char).unwrap_or(N);
     unsafe { ptr_to_cstr_n(arr.as_ptr(), end_index as size_t) }
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -7,6 +7,9 @@ use std::ptr::NonNull;
 use std::sync::{Arc, Weak};
 
 pub unsafe fn ptr_to_cstr(ptr: *const c_char) -> Option<&'static str> {
+    if ptr.is_null() {
+        return None;
+    }
     unsafe { CStr::from_ptr(ptr) }.to_str().ok()
 }
 
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -328,8 +328,15 @@ macro_rules! invoke_binder_maker_macro_with_type {
             $this,
             $consume_v,
             $fn,
-            |v, v_size, scale| {
+            |v: *const cass_byte_t, v_size, scale| {
                 use scylla::value::CqlDecimal;
+                if v.is_null() {
+                    if v_size != 0 {
+                        return Err(CassError::CASS_ERROR_LIB_BAD_PARAMS);
+                    }
+                    // cpp-driver treats NULL+0 as empty varint.
+                    return Ok(Some(Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(&[], scale))));
+                }
                 let varint = unsafe { std::slice::from_raw_parts(v, v_size as usize) };
                 Ok(Some(Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(varint, scale))))
             },
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -266,7 +266,15 @@ macro_rules! invoke_binder_maker_macro_with_type {
             $this,
             $consume_v,
             $fn,
-            |v, v_size| {
+            |v: *const cass_byte_t, v_size| {
+                if v.is_null() {
+                    if v_size != 0 {
+                        return Err(CassError::CASS_ERROR_LIB_BAD_PARAMS);
+                    } else {
+                        // cpp-driver treats NULL+0 as empty blob, not CQL NULL.
+                        return Ok(Some(Blob(vec![])));
+                    }
+                }
                 let v_vec = unsafe { std::slice::from_raw_parts(v, v_size as usize) }.to_vec();
                 Ok(Some(Blob(v_vec)))
             },
--- a/scylla-rust-wrapper/src/binding.rs
+++ b/scylla-rust-wrapper/src/binding.rs
@@ -88,20 +88,26 @@ macro_rules! make_name_binder {
                 tracing::error!("Provided null pointer to {}!", stringify!($fn_by_name));
                 return CassError::CASS_ERROR_LIB_BAD_PARAMS;
             };
-            // NULL or empty name is clearly a programmer error. The
-            // cpp-driver accidentally accepts both (treating NULL as
-            // an empty string via SAFE_STRLEN), but we purposefully
-            // diverge and return an error early instead of silently
-            // binding to a bogus empty-named parameter.
-            if name.is_null() {
-                tracing::error!("Provided null name pointer to {}!", stringify!($fn_by_name));
-                return CassError::CASS_ERROR_LIB_BAD_PARAMS;
-            }
-            let name = unsafe { ptr_to_cstr(name) }.unwrap();
-            if name.is_empty() {
-                tracing::error!("Provided empty name to {}!", stringify!($fn_by_name));
-                return CassError::CASS_ERROR_LIB_BAD_PARAMS;
-            }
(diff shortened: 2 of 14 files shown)


scylla-rust-wrapper/src/argconv.rs
```diff
@@ -6,18 +6,37 @@ use std::os::raw::c_char;
...
+pub unsafe fn ptr_to_cstr_n(
+    ptr: *const c_char,
+    size: size_t,
+) -> Result<&'static str, PtrToStrError> {
+    if ptr.is_null() {
+        return Err(PtrToStrError::NullPointer);
+    }
+    std::str::from_utf8(unsafe { std::slice::from_raw_parts(ptr as *const u8, size as usize) })
+        .map_err(PtrToStrError::InvalidUtf8)
+}
```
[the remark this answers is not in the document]
wprzytula: This will be fixed in a follow-up.

scylla-rust-wrapper/src/session.rs
```diff
@@ -581,12 +588,18 @@ pub unsafe extern "C" fn cass_session_prepare_n(
...
         // It seems to be intended (for some weird reason, why not save a round-trip???)
         // to receive a server error in such case (CASS_ERROR_SERVER_SYNTAX_ERROR).
         // There is a test for this: `NullStringApiArgsTest.Integration_Cassandra_PrepareNullQuery`.
-        .unwrap_or_default();
+        Err(PtrToStrError::NullPointer) => "",
+        Err(PtrToStrError::InvalidUtf8(_)) => {
+            tracing::error!("Provided non-UTF8 query to cass_session_prepare(_n)!");
+            return ArcFFI::null();
+        }
+    };
```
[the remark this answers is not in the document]
wprzytula: This is not important enough IMO to act.

scylla-rust-wrapper/src/binding.rs
```diff
@@ -252,6 +267,13 @@ macro_rules! invoke_binder_maker_macro_with_type {
             $consume_v,
             $fn,
             |v, v_size| {
+                if (v as *const ()).is_null() {
+                    if v_size != 0 {
+                        return Err(CassError::CASS_ERROR_LIB_BAD_PARAMS);
+                    } else {
+                        return Ok(None);
+                    }
+                }
```
Lorak-mmk, line 315: The branches for `string` and `string_n` return `Some(Text(String::new())))`, but this returns `None`.
Why the difference? Is this difference also present in cpp-driver?
Will this `None` be serializes as cql null?

This is not obvious, but it is not mentioned in the commit message at all. Commit: handle NULL pointer in string and string_n value binders

scylla-rust-wrapper/src/binding.rs
```diff
@@ -266,12 +266,13 @@ macro_rules! invoke_binder_maker_macro_with_type {
...
-            |v, v_size| {
-                if (v as *const ()).is_null() {
+            |v: *const cass_byte_t, v_size| {
+                if v.is_null() {
                     if v_size != 0 {
                         return Err(CassError::CASS_ERROR_LIB_BAD_PARAMS);
                     } else {
-                        return Ok(None);
+                        // cpp-driver treats NULL+0 as empty blob, not CQL NULL.
+                        return Ok(Some(Blob(vec![])));
```
Lorak-mmk, line 313: Commit: bytes binder NULL+0 should produce empty blob, not CQL NULL

Why the change of type in `as` cast?
Also why does this commit even exists? Previous commit introduces the wrong semantics, then this one fixes it. Why not introduce the correct semantics immediately?

scylla-rust-wrapper/src/binding.rs
```diff
@@ -88,7 +88,20 @@ macro_rules! make_name_binder {
...
+            // binding to a bogus empty-named parameter.
+            if name.is_null() {
+                tracing::error!("Provided null name pointer to {}!", stringify!($fn_by_name));
+                return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+            }
             let name = unsafe { ptr_to_cstr(name) }.unwrap();
+            if name.is_empty() {
+                tracing::error!("Provided empty name to {}!", stringify!($fn_by_name));
+                return CassError::CASS_ERROR_LIB_BAD_PARAMS;
+            }
```
Lorak-mmk: Not a fan of this duplication. Hopefully the abstractions and changes you introduce later will fix it.

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Great. 
Even better would be returning Err on wrong utf-8 and None on null. But again, I'm guessing that your later changes address this.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -481,9 +480,19 @@ pub unsafe extern "C" fn cass_cluster_set_application_name_n(
...
+            tracing::error!(
+                "Provided null app name pointer to cass_cluster_set_application_name(_n)!"
+            );
+            return;
+        }
+        Err(PtrToStrError::InvalidUtf8(_)) => {
+            tracing::error!("Provided non-UTF8 app name to cass_cluster_set_application_name(_n)!");
+            return;
+        }
+    };
```
Lorak-mmk, line 495: cass_cluster_set_application_name_n

I wonder if this is the proper way of handling this. It will fail mostly silently - only signal will be an entry in the logs.
Perhaps a panic is warranted here?

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -510,9 +519,21 @@ pub unsafe extern "C" fn cass_cluster_set_application_version_n(
...
+            );
+            return;
+        }
+        Err(PtrToStrError::InvalidUtf8(_)) => {
+            tracing::error!(
+                "Provided non-UTF8 app version to cass_cluster_set_application_version(_n)!"
+            );
+            return;
+        }
+    };
```
Lorak-mmk, line 536: ditto

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -862,20 +883,21 @@ pub unsafe extern "C" fn cass_cluster_set_local_address_n(
...
                     return CassError::CASS_ERROR_LIB_BAD_PARAMS;
                 }
             },
-            None => {
-                tracing::error!("Provided non-utf8 ip string to cass_cluster_set_local_address_n!");
+            Err(PtrToStrError::NullPointer) => None,
+            Err(PtrToStrError::InvalidUtf8(_)) => {
+                tracing::error!("Provided non-UTF8 ip string to cass_cluster_set_local_address_n!");
                 return CassError::CASS_ERROR_LIB_BAD_PARAMS;
             }
```
Lorak-mmk, line 895: Now that we have the new ptr_to_cstr_n you can get rid of the initial `if` I think.
To do that, add a branch with `Ok(s) if s.len() == 0` that also returns None.

scylla-rust-wrapper/src/cluster.rs
```diff
@@ -956,9 +978,28 @@ pub unsafe extern "C" fn cass_cluster_set_credentials_n(
...
+        Ok(s) => s,
+        Err(PtrToStrError::NullPointer) => {
+            tracing::error!("Provided null password pointer to cass_cluster_set_credentials(_n)!");
+            return;
+        }
+        Err(PtrToStrError::InvalidUtf8(_)) => {
+            tracing::error!("Provided non-UTF8 password to cass_cluster_set_credentials(_n)!");
+            return;
+        }
+    };
```
Lorak-mmk, line 1000: ditto about silent failure.
Btw does username / password need to always be utf-8? Perhaps we should accept arbitrary bytes here.

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/binding.rs
Lorak-mmk: I see now that it does, great!

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: I see now that it does, great!

scylla-rust-wrapper/src/binding.rs
wprzytula: > Commit: bytes binder NULL+0 should produce empty blob, not CQL NULL
> 
> Why the change of type in `as` cast?

I added explicit type annotation. Before, the type was inferred. Without this annotation, after my changes compiler would yell at me that it can't infer.

> Also why does this commit even exists? Previous commit introduces the wrong semantics, then this one fixes it. Why not introduce the correct semantics immediately?

The PR begins with two commits:
1. `handle NULL pointer in string and string_n value binders`
2. `bytes binder NULL+0 should produce empty blob, not CQL NULL`

Commit 1 touches `string` value binders.
Commit 2 touches `bytes` value binders.
They're fully independent. I don't see how 2 fixes 1.

scylla-rust-wrapper/src/binding.rs
wprzytula: Aha, a rebase artifact. This commit shouldn't touch handling of `bytes`, just `string`.

scylla-rust-wrapper/src/binding.rs
Lorak-mmk: Commit one touches `bytes` as well. It starts handling null for bytes, by treating it as CQL null.

scylla-rust-wrapper/src/cluster.rs
wprzytula: I don't think that failure in setting application name warrants a panic. It's not critical.

scylla-rust-wrapper/src/cluster.rs
wprzytula: ditto

scylla-rust-wrapper/src/cluster.rs
wprzytula: > ditto about silent failure.

Here I _could_ agree to panic, but OTOH the user will anyway notice authentication failure that follows, so the failure is by no means silent. Also, issuing an `error!` level message is the proper way in this driver to signal error conditions, especially in APIs that do not allow returning an error code.

> Btw does username / password need to always be utf-8? Perhaps we should accept arbitrary bytes here.

Rust Driver doesn't allow for arbitrary bytes, so it's not a thing to be discussed here, rather in Rust Driver. Anyway, not a blocker for this PR.

Lorak-mmk: `error!` is a good way to explain what happened, because we have just error codes, not rich nested error types. It is not a good way to inform that a failure happened at all.

wprzytula:
Addressed some review comments, responded to others.

scylla-rust-wrapper/src/cluster.rs
wprzytula: So you really advise to panic (abort) here?

Lorak-mmk: I think we should not have silent failures if it's possible to avoid them. Panic is the only way to do that here. Let's panic

scylla-rust-wrapper/src/cluster.rs
Lorak-mmk: Ok

scylla-rust-wrapper/src/cluster.rs
wprzytula: It's a bigger refactor IMO, because there are much more `void` setters codewide which can fail. I could push a commit that makes this one credentials setter panic on error, but this won't be a complete solution for the issue.
Do you want me to push this commit or do we handle this separately?

wprzytula: Note that this function could already "silently" (according to your definition) fail before this PR, if null cluster pointer was provided.

Lorak-mmk: Ok, let's address silent failures separately - it looks like we need more discussion about various cases. Could you open an issue?

wprzytula: https://github.com/scylladb/cpp-rs-driver/issues/453

Lorak-mmk (approved):

PR #449, opened 2026-05-12
Introduce `CassStr*` abstractions

### LINKED ISSUE #301: Improve the API for cstring - rust string conversion and vice-versa

wprzytula:
Fixes: #301
Based on: #448

-----

## What's done

This PR introduces FFI abstractions for C string types:
- `CassStrNulTerminated` - represents a nul-terminated C string passed by a pointer;
- `CassStrLenDelimited` - represents a string slice passed by a pointer, paired with
- `CassStrLen` - represents length of `CassStrLenDelimited`. I decided to make a newtype for this to increase type safety.

All abstractions are `repr(transparent)`, and they appear in FFI functions' signatures. They provide convenient constructors and conversion methods.

#### Type safety

It's now harder to misuse one of string pointer as another.

#### Borrow checks

Both `CassStrNulTerminated` and `CassStrLenDelimited` carry a lifetime, which prevents UAFs.

#### Self-documenting code

The code is much more explicit and clear in what it does with strings.

The names are a bit verbose, but this is in line with our FFI approach (see long names of `CassPtr` abstractions, such as `CassBorrowedSharedPtr`).

-----

## Pre-review checklist

### DIFF
--- a/scylla-rust-wrapper/src/argconv.rs
+++ b/scylla-rust-wrapper/src/argconv.rs
@@ -7,7 +7,7 @@ use std::ptr::NonNull;
 use std::sync::{Arc, Weak};
 use thiserror::Error;
 
-/// Error type for `ptr_to_cstr` and `ptr_to_cstr_n`, distinguishing
+/// Error type for C string pointer conversions, distinguishing
 /// a null pointer from invalid UTF-8 input.
 #[derive(Debug, Error)]
 pub enum PtrToStrError {
@@ -37,12 +37,13 @@ pub unsafe fn ptr_to_cstr_n(
         .map_err(PtrToStrError::InvalidUtf8)
 }
 
-pub(crate) unsafe fn arr_to_cstr<const N: usize>(
-    arr: &[c_char],
-) -> Result<&'static str, PtrToStrError> {
+pub(crate) unsafe fn arr_to_cstr<const N: usize>(arr: &[c_char]) -> Result<&str, PtrToStrError> {
     let null_char = '\0' as c_char;
     let end_index = arr[..N].iter().position(|c| c == &null_char).unwrap_or(N);
-    unsafe { ptr_to_cstr_n(arr.as_ptr(), end_index as size_t) }
+    unsafe {
+        CassStrLenDelimited::from_raw(arr.as_ptr())
+            .to_str(CassStrLen::from_raw(end_index as size_t))
+    }
 }
 
 pub(crate) fn str_to_arr<const N: usize>(s: &str) -> [c_char; N] {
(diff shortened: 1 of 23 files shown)


scylla-rust-wrapper/src/argconv.rs
```diff
@@ -37,12 +37,13 @@ pub unsafe fn ptr_to_cstr_n(
...
-    arr: &[c_char],
-) -> Result<&'static str, PtrToStrError> {
+pub(crate) unsafe fn arr_to_cstr<const N: usize>(arr: &[c_char]) -> Result<&str, PtrToStrError> {
     let null_char = '\0' as c_char;
     let end_index = arr[..N].iter().position(|c| c == &null_char).unwrap_or(N);
-    unsafe { ptr_to_cstr_n(arr.as_ptr(), end_index as size_t) }
+    unsafe {
+        CassStrLenDelimited::from_raw(arr.as_ptr())
+            .to_str(CassStrLen::from_raw(end_index as size_t))
+    }
```
Lorak-mmk: Not very related to this PR, but wow I hate this function so much - both before and after your changes.

The name suggests it converts the array, but it receives a slice.
There is no length check on this slice: if it is shorter than N and has no null bytes, then UB will happen.

The only use case for this function is in our default log callback, `stderr_log_callback`.
It receives `CassLogMessage`, which stores the message inline as an array. This array is guaranteed to be null-terminated, so I see no reason not to use our existing string conversion primitives.

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -76,6 +77,178 @@ pub(crate) unsafe fn strlen(ptr: *const c_char) -> size_t {
...
+                CassStrLen::from_raw(0),
+            );
+        }
+        let len = unsafe { strlen(self.ptr) };
+        (
+            CassStrLenDelimited::from_raw(self.ptr),
+            CassStrLen::from_raw(len),
+        )
+    }
+}
```
Lorak-mmk, line 150: Why do you need to define a new struct for this, instead of having type alias like `type CassStrNulTerminated<'a> = CassPtr<'a, c_char, (Shared, CConst)>`?
The additional methods could still be implemented on the `impl` block for such specific type I think.

If there is a good reason to not do that, can you describe it in the comment?

scylla-rust-wrapper/src/argconv.rs
```diff
@@ -76,6 +77,178 @@ pub(crate) unsafe fn strlen(ptr: *const c_char) -> size_t {
...
+///
+/// Must be paired with a [`CassStrLen`] parameter to convert to `&str`
+/// via [`CassStrLenDelimited::to_str`].
+#[derive(Clone, Copy)]
+#[repr(transparent)]
+pub struct CassStrLenDelimited<'a> {
+    ptr: *const c_char,
+    _phantom: PhantomData<&'a c_char>,
+}
+
```
Lorak-mmk, line 176: ditto

scylla-rust-wrapper/src/testing/utils.rs
```diff
@@ -31,13 +31,16 @@ macro_rules! assert_cass_error_eq {
...
+        #[allow(unused_unsafe)] // Sometimes the macro is expanded in an `unsafe` context.
+        unsafe {
+            assert_eq!(
+                ___x,
+                $error,
+                "expected \"{}\", instead got \"{}\"",
+                ptr_to_cstr(cass_error_desc($error)).unwrap(),
+                ptr_to_cstr(cass_error_desc(___x)).unwrap()
+            );
+        }
```
Lorak-mmk, line 48: Commit: "future: remove no longer needed unsafe blocks"

If you don't require caller to use unsafe, doesn't it allow the possibility of undefined behavior?

For example, if `$error` is an invalid poiner?

Lorak-mmk (changes requested):

scylla-rust-wrapper/src/argconv.rs
wprzytula: A string being null-terminated is an additional, strong property that neither `*const c_char` nor `CassPtr<'_, c_char,m _>` provide. Strong typing here provides clear information about that property and prevents misuse. Are you convinced now?

scylla-rust-wrapper/src/argconv.rs
wprzytula: This newtype is added to work in tandem with `CassStrNulTerminated`, whose rationale I've already provided in response to another comment of yours.

I really like strong newtypes, they carry additional semantic information in their types.

scylla-rust-wrapper/src/testing/utils.rs
wprzytula: Macros don't have a well-established `unsafe` semantics like functions. If a function requires special care when calling, it should be marked as `unsafe`. We could follow the same requirement with macros, but note that error message when forgetting to enclose the macro invocation in the `unsafe` block will not be very clear - it will mention the expanded code, not even the macro's name.

All in all, it's not clear what to do. I did the necessary changes to calm clippy down.

Lorak-mmk: Can we calm clippy down some other way?
Calling this macro safely requires some invariants, so imo it should be enclosed in unsafe.

scylla-rust-wrapper/src/argconv.rs
Lorak-mmk: Right, I did not think of that sorry.
WDYT about making it a newtype around CassPtr, to avoid basically duplicating it?

wprzytula: This makes sense of course, I'll do that.

scylla-rust-wrapper/src/testing/utils.rs
wprzytula: > Can we calm clippy down some other way?

Yes, we can add `#[allow(unused_unsafe)]` in the macro body.

Lorak-mmk: Please do that.

wprzytula:
Rebased on master.

wprzytula:
Made the new `CassStr*` abstractions newtypes over `CassBorrowedSharedPtr`.

scylla-rust-wrapper/src/testing/utils.rs
wprzytula: Removed the commit altogether.

wprzytula:
Deleted the commit that introduced the `unsafe` blocks in the macros.

scylla-rust-wrapper/src/logging.rs
```diff
@@ -88,7 +87,7 @@ pub(crate) unsafe extern "C" fn stderr_log_callback(
         unsafe { CassStrNulTerminated::from_raw(message.file).to_str() }
             .expect("file is set to a null-terminated Rust string by on_event"),
         message.line,
-        unsafe { arr_to_cstr::<{ CASS_LOG_MAX_MESSAGE_SIZE }>(&message.message) }
+        unsafe { CassStrNulTerminated::from_raw(message.message.as_ptr()).to_str() }
             .expect("message is populated from a Rust String via str_to_arr"),
     )
```
Lorak-mmk, line 92: Nit: message in `expect` should also be adjusted,

Lorak-mmk (approved):

scylla-rust-wrapper/src/logging.rs
wprzytula: How? It's still up-to-date. `str_to_arr` is still used; we removed `arr_to_cstr`.

Lorak-mmk: Ah right, nvm then

PR #452, opened 2026-05-18
fixup: dll name mismatch on windows

MaikeR-TV:
Since cargo_build produces scylla_cpp_driver.dll and .lib on shared build, when the "INSTALL_NAME"s are changed to scylla-cpp-driver, this will lead to "dll not found" errors when trying to run the final executable. While after the build, we could rename the dll again to circumvent the startup issue, this leaves a Conan CMakeDeps warning

fixes: `-- DLL library not found, creating UNKNOWN IMPORTED target, searched for: scylla-cpp-driver`
fixes: `dll not found` on startup

## Pre-review checklist


scylla-rust-wrapper/CMakeLists.txt
wprzytula, line 1: The naming for libraries with the legacy cpp-driver used hyphens. As we wanted to preserve compatibility, cpp-rs-driver uses hyphens too.

The problem is, cargo outputs dynamic libraries normalized to underscores (not hyphens). Then we copy-rename the artifacts to ones with hyphens.

For Linux, this works because .SONAME is set correctly to hyphens. See `readelf` output:
`0x000000000000000e (SONAME)             Library soname: [libscylla-cpp-driver.so.1]`

Windows doesn't have an equivalent of `.SONAME`, and the name the linker uses seems to be determined by the original cargo output (with underscores), not by the copied target (with hyphens).

Your fix breaks the assumption that the artifacts use hyphens, which is most likely a deviation from cpp-driver. Also, it introduces discrepancy between Windows (underscores) and Linux (hyphens).

wprzytula: @MaikeR-TV WDYT?

MaikeR-TV: I understand that having differences between different OS and also between the old vs. new driver is problematic.
For us though, the rename seems to break the conan-cmake integration flow into our apps.

Since cargo doesn't support names with hyphens, we'll need to find another solution.
Could make this change configurable to allow the name to be overwritten.

wprzytula: We're considering breaking the backwards naming compatibility by changing the library name for all platforms to `scylla` or `scylladb`, similarly to the original driver's name (`cassandra`). So no hyphens nor underscores.
Do you think this will affect users badly? Will switching the library name be mostly seamless?

MaikeR-TV: For us it should work just fine with a different name as long as the correct name is set in cmake so it can be resolved by conan.

wprzytula: #464 This describes what I'm going to do to fix this problem.

MaikeR-TV: Thank you.
I will close this PR.

PR #456, opened 2026-05-21
packaging: fix dev package dependency resolution

### LINKED ISSUE #454: scylla_cpp_driver-dev_1.0.0_amd64.deb doesn't include the actual .so library
The devel version of the release package (at least DEB one) installs a symlink but not the actual .so library:

```
$ sudo dpkg -L scylla-cpp-driver-dev
/usr
/usr/include
/usr/include/cassandra.h
/usr/include/cassandra_deleted_functions.h
/usr/lib64
/usr/lib64/libscylla-cpp-driver_static.a
/usr/lib64/pkgconfig
/usr/lib64/pkgconfig/scylla-cpp-driver.pc
/usr/lib64/pkgconfig/scylla-cpp-driver_static.pc
/usr/lib64/libscylla-cpp-driver.so
```
`/usr/lib64/libscylla-cpp-driver.so` is a symlink:

```
$ ll /usr/lib64/libscylla-cpp-driver.so
lrwxrwxrwx 1 root root 29 Feb 24 15:46 /usr/lib64/libscylla-cpp-driver.so -> libscylla-cpp-driver.so.1.0.0*
```

This naturally makes it difficult to link against ;)

dkropachev:
## Summary
- add the runtime package dependency to the DEB/RPM dev packages so the unversioned linker symlink and pkg-config files do not install without the shared library
- exercise dependency resolution in the smoke tests by installing the dev package through a temporary local repository and letting apt/dnf resolve the runtime package

Fixes: https://github.com/scylladb/cpp-rs-driver/issues/454
## Testing
- `make test-package-deb`
- `make test-package-rpm`

## Notes
- macOS packaging was checked separately; it uses split component packages as well, but the current productbuild/DragNDrop flow does not encode equivalent package-manager dependency metadata.

## Checklist


packaging/smoke-test-app/Makefile
```diff
@@ -183,7 +217,15 @@ install-driver-dev-deb:
...
-	${INSTALL_PACKAGE_SCRIPT}
+	${VALIDATE_SINGLE_PACKAGE}
+	${MAYBE_SUDO}
+	runtime_packages=(${MAKEFILE_PATH}/../../build/scylla_cpp_driver_*.deb)
+	if [ "$${#runtime_packages[@]}" -ne 1 ]; then echo "No scylla-rs-driver runtime DEB package found"; exit 1; fi
+	${PREPARE_TEMP_DEB_REPO}
+	# Install the dev package from the temporary repo and exercise apt's resolver.
+	$$SUDO apt-get -o Acquire::Languages=none -o Dir::Etc::sourcelist="$$repo_dir/sources.list" -o Dir::Etc::sourceparts=- -o Dir::State::lists="$$repo_dir/lists" install -y scylla-cpp-driver-dev
+	dpkg -s scylla-cpp-driver >/dev/null
+	dpkg -s scylla-cpp-driver-dev >/dev/null
```
wprzytula, line 233: 💭 Previously, `INSTALL_PACKAGE_SCRIPT` handled all cases: {dev,runtime} \crossprod {rpm,deb,msi,pkg}. Now it's only used for runtime cases, while dev cases use their own inline logic.
❓ Can we extract the common part for dev cases similarly to runtime?

packaging/smoke-test-app/Makefile
wprzytula, line 1: 💭 It's hard for me to navigate in this file and grasp it, because there is no Big Picture view what it achieves and how; subroutines do not have documentation; the are few comments. Please document the file in a new commit, adding both higher-granularity comments and a Big Picture view.

packaging/smoke-test-app/Makefile
```diff
@@ -42,11 +42,11 @@ endif
 define VALIDATE_SINGLE_PACKAGE
 	set -euo pipefail
```
wprzytula, line 43: commit `packaging: test dependency resolution for dev packages`:
🔧 There's no commit message that would explain in detail what was done before and what is done now.

wprzytula (changes requested):

packaging/smoke-test-app/Makefile
dkropachev: Addressed by factoring the shared throw-away repo bootstrap into `PREPARE_TEMP_REPO`. I kept the DEB and RPM helpers separate because the metadata generation and install commands are still package-manager-specific, and collapsing those would not buy much beyond a larger abstraction.

packaging/smoke-test-app/Makefile
dkropachev: Addressed. I added a file-level overview at the top of `packaging/smoke-test-app/Makefile` and step comments around the temporary repo flow so the purpose of the harness and the reason for the throw-away repos are explicit.

packaging/smoke-test-app/Makefile
dkropachev: Addressed. I rewrote the second commit message to `packaging: test dev package dependency resolution via temp repos` and added a body that explains the temporary apt/dnf repositories and the smoke-test build-dir cleanup.

packaging/smoke-test-app/Makefile
```diff
@@ -183,7 +222,15 @@ install-driver-dev-deb:
 	@shopt -s nullglob
 	packages=(${MAKEFILE_PATH}/../../build/scylla_cpp_driver-dev_*.deb)
 	package_name="scylla-rs-driver dev DEB"
```
wprzytula: 💭 I know it's a preexisting thing, but why is the package named `scylla-rs-driver`? Shouldn't this be `scylla-cpp-driver` or similar?

packaging/smoke-test-app/Makefile
```diff
@@ -12,6 +12,14 @@ DRIVER_BUILD_DIR ?= ${MAKEFILE_PATH}/../../build
 DRIVER_INSTALL_PATH ?= C:/Program Files/ScyllaDB/Scylla CPP Driver
 SMOKE_TEST_INSTALL_PATH ?= C:/Program Files/ScyllaDB/Scylla CPP Driver Smoke Test
 
+# Smoke-test packaging flow:
+# - build the smoke app package from this tree
+# - install the driver runtime or dev artifact under test
+# - install the smoke app package and run it against ScyllaDB
+#
+# The dev-package helpers create throw-away local repositories so apt/dnf
+# resolve the runtime package from the packages produced by this build.
```
wprzytula: ♻️ commit `packaging: test dev package dependency resolution via temp repos` is still a bag of loosely-related (or fully unrelated) changes. See the description:

> Build throw-away apt/dnf repositories from the local runtime and dev artifacts so the smoke tests exercise package-manager dependency resolution instead of installing both files directly. Clear the smoke-test build directory before rerunning package targets so caches from earlier container runs do not leak across tests. Harden non-Windows smoke-test cluster startup with retries and explicit container health checks before running the smoke binary.

I can see at least 3 distinct things:
1. > Build throw-away apt/dnf repositories from the local runtime and dev artifacts so the smoke tests exercise package-manager dependency resolution instead of installing both files directly.
2. > Clear the smoke-test build directory before rerunning package targets so caches from earlier container runs do not leak across tests.
3. > Harden non-Windows smoke-test cluster startup with retries and explicit container health checks before running the smoke binary.

IMO each should come in a separate commit. Now it's hard to grasp each change, because they are mixed. Once again @dkropachev, I urge you to keep changes as separate as possible. Otherwise, both code quality and review quality deteriorate.

wprzytula: BTW, Opus is quite good in splitting changes into commits. I ask it to suggest commit split and its suggestions are mostly reasonable.

wprzytula: Just ask it to document each change thorougly in a commit message, **and make sure you fully agree with the explanation**.

wprzytula (changes requested):

packaging/smoke-test-app/Makefile
dkropachev: Done, please take a look

packaging/smoke-test-app/Makefile
wprzytula: You marked this as resolved, but no answer nor code changes.

docs/source/topics/installation.md
```diff
@@ -15,34 +15,33 @@ To install the dependencies:
...
 
 ```bash
-sudo dnf install -y scylla-cpp-driver-<VERSION>.rpm
+sudo dnf install -y ./scylla_cpp_driver_<VERSION>_x86_64.rpm
 ```
 
-When developing against the driver you'll also want to install the development
-package and the debug symbols:
+When developing against the driver you'll also want to install the runtime and
+development packages:
```
wprzytula:
```suggestion
When developing against the driver you'll also want to install the development package:
```

packaging/smoke-test-app/Makefile
```diff
@@ -42,11 +42,11 @@ endif
...
 	if [ "$${#packages[@]}" -eq 0 ]; then
-		echo "No ${package_name} package found"
+		echo "No $${package_name} package found"
 		exit 1
 	fi
 	if [ "$${#packages[@]}" -gt 1 ]; then
-		echo "Found more than one file for ${package_name}: $${packages[@]}"
+		echo "Found more than one file for $${package_name}: $${packages[@]}"
 		exit 1
 	fi
```
wprzytula, line 51: ♻️ These are fixes to existing code's flaws, not something connected to the commit's topic, right? `packaging: test dev package dependency resolution via temp repos`

Then this should come in a separate commit, **with an explanation**.

packaging/smoke-test-app/Makefile
wprzytula: You extracted point 3 to another PR. Points 1 and 2 are still packed in one commit. I ask you again to split them into at least 2 commits.
IIUC, point 2 (clearing the test dirs) can be introduced before point 1 (testing via apt/dnf repos).

wprzytula (changes requested):

packaging/smoke-test-app/Makefile
dkropachev: Done.

packaging/smoke-test-app/Makefile
dkropachev: done

docs/source/topics/installation.md
wprzytula: Why is this set as resolved? It's not fixed.

packaging/smoke-test-app/Makefile
wprzytula, line 1: commit `packaging: escape package_name diagnostics` - commit message says:
> Make expanded ${package_name} before recipe execution, so validation errors could omit the package label. Escape it as a shell variable.

I can see the escaping here (second sentence). What is the first sentence about?

dkropachev: > commit `packaging: escape package_name diagnostics` - commit message says:
> 
> > Make expanded ${package_name} before recipe execution, so validation errors could omit the package label. Escape it as a shell variable.
> 
> I can see the escaping here (second sentence). What is the first sentence about?

`${var}` is expanded by make as a Makefile variable. `$${var}` makes `$$` become a literal `$`, so `${var}` is passed to the shell and expanded there.

wprzytula (approved):
Thanks!

PR #458, opened 2026-05-22
packaging: harden smoke-test cluster startup

dkropachev:
## Summary
- harden the non-Windows smoke-test cluster startup path used by package tests
- retry `docker compose up`, wait for the Scylla container to become healthy, and dump compose state/logs when startup fails
- keep compose cleanup non-fatal so transient teardown errors do not hide the real startup failure

## Testing
- not run locally on macOS in this branch split; logic was originally validated on the combined branch via `make test-package-linux` and by the succeeding macOS package CI job after the hardening was introduced

## Checklist
- [ ] I added appropriate `Fixes:` annotations to PR description.

### DIFF
--- a/packaging/smoke-test-app/Makefile
+++ b/packaging/smoke-test-app/Makefile
@@ -410,6 +410,7 @@ test-app-package: .prepare-for-test
 		"$$smoke_bin" $(SCYLLA_HOST);\
 	else\
 		dc="";\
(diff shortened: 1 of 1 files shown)


wprzytula (changes requested):
I doubt this is needed at all. See https://github.com/docker/compose/pull/8777#issue-1022875399. If a healthcheck condition is defined (which is the case in our `docker-compose.yml`), `docker compose up --wait` waits until the service is healthy.

PR #461, opened 2026-05-25
Gtest IT: log also to stderr, not only to file

wprzytula:
## Motivation

[A recent CI run](https://github.com/scylladb/cpp-rs-driver/actions/runs/26300848077/job/77426773254?pr=456) experienced a panic inside the Rust Driver. I couldn't debug it, because no logs were printed.

#### Destination

Before, the logs were only output to a file. Now, they are additionally printed to stderr.

#### Format

1. Timestamp display was improved, made in line with Rust tracing subscriber's one.
2. Log level is now coloured (IFF stderr points to a terminal). It uses the same palette as, again, Rust tracing subscriber.

### CI

1. `RUST_LOG=trace` is now passed, so the logs are visible in the CI run console.
2. `RUST_BACKTRACE=full` is now passed, which will aid in debugging. 

## Pre-review checklist

### DIFF
--- a/tests/src/integration/logger.cpp
+++ b/tests/src/integration/logger.cpp
@@ -32,11 +32,13 @@ using namespace test::driver;
 #ifdef _WIN32
 #define FILE_MODE 0
 #define LOCALTIME(tm, time) localtime_s(tm, time)
+#define GMTIME(tm, time) gmtime_s(tm, time)
 #include <io.h>
 #define IS_STDERR_TTY() _isatty(_fileno(stderr))
 #else
 #define FILE_MODE S_IRWXU | S_IRWXG | S_IROTH
 #define LOCALTIME(tm, time) localtime_r(time, tm)
+#define GMTIME(tm, time) gmtime_r(time, tm)
 #include <unistd.h>
 #define IS_STDERR_TTY() isatty(STDERR_FILENO)
 #endif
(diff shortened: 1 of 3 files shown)


wprzytula:
CI failures will hopefully be fixed with https://github.com/scylladb/scylla-rust-driver/pull/1644.

tests/src/integration/logger.cpp
```diff
@@ -125,7 +125,11 @@ void Logger::log(const CassLogMessage* log, void* data) {
   // Create the formatted log message and output to the file
   std::string severity = cass_log_level_string(log->severity);
   logger->output_ << date.str() << " " << time.str() << " " << severity << ": " << message << " ("
-                  << log->file << ":" << log->line << ") " << std::endl;
+                   << log->file << ":" << log->line << ") " << std::endl;
+
+  // Also log to stderr so that CI output captures driver logs in real-time.
+  std::cerr << log->time_ms << " [" << severity << "] ("
+            << log->file << ":" << log->line << ") " << message << std::endl;
```
Lorak-mmk, line 132: Is there a reason why the format is different between the two?

tests/src/integration/logger.cpp
```diff
@@ -147,7 +149,14 @@ void Logger::log(const CassLogMessage* log, void* data) {
...
+  cass_uint64_t const epoch_secs = log->time_ms / 1000;
+  unsigned short const ms = log->time_ms % 1000;
+  time_t const raw = static_cast<time_t>(epoch_secs);
+  struct tm utc;
+  GMTIME(&utc, &raw);
+  char ts[24];
+  strftime(ts, sizeof(ts), "%Y-%m-%dT%H:%M:%S", &utc);
+  std::cerr << ts << "." << std::setfill('0') << std::setw(3) << ms << "Z" << " [" << color << severity << reset << "] "
             << dim << "(" << log->file << ":" << log->line << ")" << reset
             << " " << message << std::endl;
```
Lorak-mmk, line 162: ditto, why only adjust the cerr output and not both?

Lorak-mmk (approved):

tests/src/integration/logger.cpp
wprzytula: I wanted to keep the original format untouched, not to break parsers etc.;
the console format is new, so nothing to break.

Lorak-mmk: Ok

PR #466, opened 2026-05-28
chore: update openssl and openssl-sys in Cargo.toml, update Cargo.lock

MaikeR-TV:
We want to update our OpenSSL to v4 (new API, no deprecations) and this requires newer versions for the rust openssl and openssl-sys crates.

Fixes: Cannot build driver with OpenSSL 4 (new API, no deprecations)

### DIFF
--- a/scylla-rust-wrapper/Cargo.lock
+++ b/scylla-rust-wrapper/Cargo.lock
@@ -787,15 +787,14 @@ checksum = "42f5e15c9953c5e4ccceeb2e7382a716482c34515315f7b03532b8b4e8393d2d"
 
 [[package]]
 name = "openssl"
-version = "0.10.75"
+version = "0.10.80"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08838db121398ad17ab8531ce9de97b244589089e290a384c900cb9ff7434328"
+checksum = "a45fa2aa886c42762255da344f0a0d313e254066c46aad76f300c3d3da62d967"
 dependencies = [
  "bitflags 2.11.0",
  "cfg-if",
  "foreign-types",
  "libc",
- "once_cell",
  "openssl-macros",
  "openssl-sys",
 ]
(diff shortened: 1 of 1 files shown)


scylla-rust-wrapper/Cargo.toml
```diff
@@ -25,8 +25,8 @@ rand = "0.8.4"
 num-traits = "0.2"
 num-derive = "0.3"
 libc = "0.2.108"
-openssl-sys = "0.9.75"
-openssl = "0.10.32"
+openssl-sys = "0.9.114"
+openssl = "0.10.78"
```
wprzytula: ❓ Is there any reason to bump versions here?

Cargo.toml specifies minimal versions of dependencies, while Cargo.lock chooses the actual used versions. As you can see, we already used newer versions (`openssl 0.10.75`, `openssl-sys 0.9.111`) than Cargo.toml specified (`openssl 0.10.32`, `openssl-sys 0.9.75`)

MaikeR-TV: From our research, minimal version of openssl is openssl-v0.10.78 and for openssl-sys it's openssl-sys-v0.9.114 in order to support OpenSSL 4.
So the minimal version in this Cargo.toml should reflect that.

wprzytula: I'm not opposed. I was just curious why update the `Cargo.toml` for that, when `Cargo.lock` update would achieve the same.

wprzytula: cc @Lorak-mmk

wprzytula (approved):

scylla-rust-wrapper/Cargo.toml
wprzytula: Cargo.toml should, in general, specify the lowest viable dependency versions for the application to work correctly; Cargo.lock should specify the actual used versions.

Now, there may be users that want to use an older `openssl` crate in their driver build. For that, they'll specify a lower dep version in the Cargo.lock; they know they can safely do that as long as Cargo.toml allows for that.

wprzytula: @MaikeR-TV WDYT?

MaikeR-TV: I was wondering about the failing checks on the PR build.

You're suggesting to only update the lock file with this merge request and leave the toml as the update is not necessary in general, depending on the OpenSSL version that consumers use?
I can update the PR.

wprzytula: > I was wondering about the failing checks on the PR build.

They are unrelated to this PR, I'm working on fixing them.

> You're suggesting to only update the lock file with this merge request and leave the toml as the update is not necessary in general, depending on the OpenSSL version that consumers use? I can update the PR.

Indeed. Please update it.

wprzytula (approved):

PR #468, opened 2026-06-02
future: Fix double callback execution

Lorak-mmk:
`ResolvableFuture::set_callback` first set a callback under lock, released the lock, and then checked if result is set. If it is, then it invoked the callback.
The other path for callback calling is in future execution: when setting the result (under lock) new_from_future also reads a callback. If it exsits, then it is called.
the following race is possible:
- `set_callback` sets callback, releases lock
- `new_from_future` produces result, sees the callback, calls it
- `set_callback` checks for the result, sees it exists, calls the callback.

This results in double callback execution, potentially causing UB and other issues.
This commit fixes that: in `set_callback`, while still under lock, we check if callback should be called. Then outside of lock we call it if needed. It is important to call the callback outside the lock, to avoid various deadlocks.

Fixes: https://scylladb.atlassian.net/browse/CUSTOMER-409

## Pre-review checklist

- [ ] ~~I have implemented Rust unit tests for the features/changes introduced.~~ - no clue how to test that

### DIFF
--- a/scylla-rust-wrapper/src/future.rs
+++ b/scylla-rust-wrapper/src/future.rs
@@ -424,7 +424,7 @@ impl ResolvableFuture {
         // Check if the callback is already set (in such case we must error out).
         // If it is not set, we store the callback in the state, so that no different
         // callback can be set.
-        {
+        let should_invoke_callback = {
             let mut lock = self.state.lock().unwrap();
             if lock.callback.is_some() {
                 // Another callback has been already set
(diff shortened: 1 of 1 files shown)


wprzytula (approved):

wprzytula:
Can you add a test that with high probability causes a race with the unfixed code?

Lorak-mmk:
I would love to, but I don't really have any idea how to write such test.

PR #489, opened 2026-08-05
Add TLS testing via CCM

wprzytula:
## Background / Motivation

Someone noticed that our TLS verification default did not match what we
document. `include/cassandra.h` claimed:

```
 * <b>Default:</b> CASS_SSL_VERIFY_PEER_CERT
```

while `cass_ssl_new_no_lib_init()` actually did:

```rust
SSL_CTX_set_verify(ssl_context, CASS_SSL_VERIFY_NONE, None);
```

`CASS_SSL_VERIFY_NONE` is `0x00`, which happens to be `SSL_VERIFY_NONE` too, so the effective default was **no peer verification whatsoever** — the exact opposite of what the header promised, and a security-relevant discrepancy: a
user who read the docs and did not call `cass_ssl_set_verify_flags()` believed the server certificate was being validated when it was not.

Looking into it, I found the more fundamental problem: **none of this was tested**. There was no coverage of TLS at all on the Rust side — no test would have caught the wrong default, and no test would catch us breaking it again.
`cass_ssl_set_verify_flags()` was similarly untested, and it showed: unknown flag values silently fell through to a catch-all arm that enabled `SSL_VERIFY_PEER`, so passing garbage quietly changed the security settings.

This PR does one thing: **adds real end-to-end TLS tests**.
A follow up PR will use them to validate the fixed defaults and the flag handling, and finally implement correctly the one flag (`CASS_SSL_VERIFY_PEER_CERT`) we did not respect before due to Rust Driver's behavioural change (turning on hostname verification in OpenSSL - https://github.com/scylladb/scylla-rust-driver/pull/1491).

### Appetizers

- added docstring and docs checkboxes to the PR template. They were missing.
- Makefile: bumped ScyllaDB version to 2026.2.2. This also reduce testing time via CCM enormously, because the previous version was not a full SemVer release - it was missing patch number - which caused a web request to S3 on each ccm operation (see https://github.com/scylladb/scylla-ccm/issues/771 for more details).

### Updated docs to mention TLS verification's actual default

There was a discrepancy between the documentation and the actual default value of the SSL verification option in the `cassandra.h` header file. The documentation stated that the default was `CASS_SSL_VERIFY_PEER_CERT`, but in reality, it was set to `CASS_SSL_VERIFY_NONE`.
Note that this change is only to align the documentation with the actual behavior of the code. The default value of the SSL verification option should likely be changed to enable verification, but that requires further changes, so is left for a follow-up.

### Bumped Rust Driver

- ran `cargo update`.
- got new Rust Driver's fixes.
- got access to recently extracted `scylla-ccm-bridge`, which we will use for writing CCM tests in Rust.

### Tests written

End-to-end TLS tests that drive the driver through its **C API**, exactly as a C/C++ consumer would, against a real TLS-enabled ScyllaDB cluster. They are ports of the Rust Driver's `ccm::tls` tests.

The clusters are managed from Rust via the **`scylla-ccm-bridge` crate from the Rust Driver**, added as a dev-dependency. That lets us write CCM tests in Rust rather than going through the C++ CCM harness, which makes them far easier to maintain. Certificates are generated on the fly with `rcgen`, so there is nothing checked in and nothing to rotate: the harness creates a CA, issues a per-node server certificate, installs it via `ccm <node> updateconf`, and hands the CA to the test so it can build a matching client trust store.

One practical note for anyone running these locally: pass a **fully-qualified** cluster version. `scylla-ccm` re-resolves the version on *every* `ccm` invocation, and a partial one such as `release:2026.2` makes it list an S3 bucket and sleep for a random 0–5 seconds each time, which dominates the runtime (~31 s vs ~8 s for the whole suite). The default in the Makefile is fully qualified, and the harness warns if `SCYLLA_TEST_CLUSTER` is not.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ ifndef CCM_COMMIT_ID
 endif
 
 ifndef SCYLLA_VERSION
-	SCYLLA_VERSION := release:2025.3
+	SCYLLA_VERSION := release:2026.2.2
 endif
 
 ifndef CASSANDRA_VERSION
--- a/Makefile
+++ b/Makefile
@@ -509,7 +509,10 @@ endif
 
 run-test-unit: install-cargo-if-missing
 	@cd ${CURRENT_DIR}/scylla-rust-wrapper
-	RUSTFLAGS="${FULL_RUSTFLAGS}" cargo test
+	@# The `ccm` tests require a running CCM environment and a real cluster,
+	@# so they are excluded here. They are run as part of the integration
+	@# test targets instead (see `run-test-integration-scylla`).
+	RUSTFLAGS="${FULL_RUSTFLAGS}" cargo test -- --skip ccm
 
 # Currently not used.
 CQLSH := cqlsh
--- a/Makefile
+++ b/Makefile
@@ -495,6 +495,18 @@ endif
 	valgrind --error-exitcode=123 --leak-check=full --errors-for-leak-kinds=definite build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${SCYLLA_TEST_FILTER}"
 	@echo "Running timeout sensitive tests on scylla ${SCYLLA_VERSION}"
 	build/cassandra-integration-tests --scylla --version=${SCYLLA_VERSION} --category=CASSANDRA --verbose=ccm --gtest_filter="${SCYLLA_NO_VALGRIND_TEST_FILTER}"
+	@echo "Running Rust CCM integration tests on scylla ${SCYLLA_VERSION}"
+	@cd ${CURRENT_DIR}/scylla-rust-wrapper
+	@# These tests start real, TLS-enabled clusters via CCM and drive the
+	@# driver through its C API. `SCYLLA_TEST_CLUSTER` selects the CCM version
+	@# (see the scylla-ccm-bridge crate). Ignored tests (documenting not-yet
+	@# implemented behavior) are intentionally not run.
+	@#
+	@# Prefer a fully-qualified SCYLLA_VERSION (e.g. release:2025.3.8 rather
+	@# than release:2025.3): scylla-ccm re-resolves the version on every `ccm`
+	@# invocation, and a partial one forces it to list an S3 bucket and sleep
+	@# for a random 0-5 seconds each time, which dominates the test runtime.
+	SCYLLA_TEST_CLUSTER="${SCYLLA_VERSION}" CCM_ROOT_DIR=/tmp/ccm-rust RUSTFLAGS="${FULL_RUSTFLAGS}" cargo test --test integration ccm
 
 run-test-integration-cassandra: install-java8-if-missing
 ifdef DONT_REBUILD_INTEGRATION_BIN
--- a/scylla-rust-wrapper/tests/integration/ccm/tls.rs
+++ b/scylla-rust-wrapper/tests/integration/ccm/tls.rs
@@ -7,6 +7,7 @@
 
 use std::ffi::CString;
 use std::net::{IpAddr, Ipv4Addr};
+use std::os::raw::c_char;
 use std::sync::Arc;
 
 use futures::StreamExt as _;
@@ -31,7 +32,8 @@ use scylladb::api::session::{
     cass_session_new,
 };
 use scylladb::api::ssl::{
-    cass_ssl_add_trusted_cert, cass_ssl_free, cass_ssl_new, cass_ssl_set_verify_flags,
+    cass_ssl_add_trusted_cert, cass_ssl_free, cass_ssl_new, cass_ssl_set_cert,
+    cass_ssl_set_private_key, cass_ssl_set_verify_flags,
 };
 use scylladb::api::statement::{cass_statement_free, cass_statement_new};
 use scylladb::argconv::CassStrNulTerminated;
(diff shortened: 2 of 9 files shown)


scylla-rust-wrapper/tests/integration/ccm/tls.rs
```diff
@@ -0,0 +1,336 @@
...
+                    .await
+                    .unwrap();
+
+                // scylla-ccm cannot handle `ccm updateconf` after a per-node
+                // `ccm <node> updateconf`, so all cluster-wide configuration
+                // (including anything from `cluster_config`) must be applied
+                // before configuring the individual nodes below.
+                // See: https://github.com/scylladb/scylla-ccm/issues/686
+                let mut cluster = cluster_config(cluster).await;
+
```
Lorak-mmk: You are linking to a closed issue. Is it still relevant?  You also mentioned this problem in the commit message.

scylla-rust-wrapper/tests/integration/ccm/tls.rs
Lorak-mmk, line 1: Why are those tests using tokio? Tokio is an internal detail from the users perspective, I don't see a good reason for `async` here. Especially that you don't use concurrency, and have to use `spawn_blocking`.

Lorak-mmk (changes requested):

scylla-rust-wrapper/tests/integration/ccm/tls.rs
wprzytula: Idk if it's still relevant. I experienced a race and Opus attributed the race occurrence to that issue. I can remove the mentions of the issue from the code and from the commit message. The way the test is written works and this is what really matters.

scylla-rust-wrapper/tests/integration/ccm/tls.rs
wprzytula: `scylla-ccm-bridge` exposes async API. This means I need some executor to run. tokio makes sense as it's already a dependency. I'm not sure if `scylla-ccm-bridge` uses some tokio internals or not; if not, we can consider using another executor here, e.g. [futures::executor](https://docs.rs/futures/latest/futures/executor/index.html).

scylla-rust-wrapper/tests/integration/ccm/tls.rs
```diff
@@ -0,0 +1,529 @@
...
+    let cert_file_path = node.node_dir().join("db.cert");
+    let mut cert_file = File::create_new(&cert_file_path).await.unwrap();
+    cert_file.write_all(cert.pem().as_bytes()).await.unwrap();
+
+    let key_file_path = node.node_dir().join("db.key");
+    let mut key_file = File::create_new(&key_file_path).await.unwrap();
+    key_file
+        .write_all(key.serialize_pem().as_bytes())
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
wprzytula: False. `tokio::fs::File` passes writes straight to OS `std::fs::File`. Only `BufWriter` does internal user-space buffering. `tokio::fs::File::flush()` is noop.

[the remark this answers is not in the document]
wprzytula: See: https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#1104

scylla-rust-wrapper/tests/integration/ccm/tls.rs
Lorak-mmk: Ah, ok. I forgot about async functions from ccm. In that case tokio makes sense.

scylla-rust-wrapper/tests/integration/ccm/tls.rs
Lorak-mmk: Let's remove the mention of this issue from comment and commit message. Using a fixed issue as a rationale for doing something is just confusing for current and future readers.

wprzytula: Actually, the commit message mentions a separate issue which is real (executing `ccm node updateconf` concurrently not working correctly), so I won't change the commit message.
I'll remove the comment from the code.

scylla-rust-wrapper/tests/integration/ccm/tls.rs
```diff
@@ -0,0 +1,524 @@
...
+const CASS_SSL_VERIFY_PEER_IDENTITY: i32 = 0x02;
+
+fn cluster_3_nodes() -> ClusterOptions {
+    ClusterOptions {
+        name: "cluster_tls_3_node".to_string(),
+        version: CLUSTER_VERSION.clone(),
+        nodes_per_dc: vec![3],
+        ..ClusterOptions::default()
+    }
+}
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk Do you know if same cluster name will cause conflicts in CCM?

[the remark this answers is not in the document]
Lorak-mmk: There should be no problem. We are using a separate temp config dir for each cluster.

Makefile
```diff
@@ -177,7 +177,7 @@ ifndef CCM_COMMIT_ID
 endif
 
 ifndef SCYLLA_VERSION
-	SCYLLA_VERSION := release:2025.3
+	SCYLLA_VERSION := release:2026.2.2
 endif
```
Lorak-mmk, line 182: I invite you to configure Renovate to bump this automatically.

Lorak-mmk (approved):

PR #499, opened 2026-09-04
Allow changing log level on the fly

### LINKED ISSUE #115: Logging: allow setting log level more than once
After #111 was merged, setting the logging level by calling `cass_log_set_level` implied that a global tracing subscriber is set for the duration of the entire program. As the default global subscriber can be set only once, later attempts to set the logging level are ignored. Although the logging configuration should not be changed after a session is established, this limitation is too restrictive, so perhaps it is possible to avoid it by setting a default subscriber for each thread before they start an execution:
https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start

### LINKED ISSUE #498: Improve setting log level
## Problem Statement

Log level can be configured via `cass_log_set_level`. It currently has the following shortcomings:
1. It works only once. After it's called once, the global tracing subscriber is configured and can't be altered.
2. It must be called before any function that can issue any logs.
3. (a bug more than a feature IMO) It follows `RUST_LOG` env var, because it uses `EnvFilter`. The driver is for C/C++, so it shouldn't react to such a foreign env var contents.

This is a serious pitfall that users can fall into. For instance, run-time log level modifications are not possible, though the API can mislead to believe it's supported.

## Suggested Solution

1. Switch the subscriber implementation to allow run-time log level modifications.
2. No longer use `EnvFilter` to stop reacting to `RUST_LOG`.

wprzytula:
Fixes: #498
Fixes: #115

# What

`cass_log_set_level` had two big limitations, both stemming from its implementation: it *installed* a whole new global `tracing` subscriber, with the requested level baked into that subscriber's `EnvFilter`.

1. It had to be called before anything that could log. Until the first call there was no subscriber at all, so every earlier event was dropped on the floor and the user's log callback never saw it.
2. It had to be called at most once. `tracing::subscriber::set_global_default` succeeds only once per process, and its error was swallowed with `.unwrap_or(())`, so every later call silently did nothing - the level was frozen for the rest of the program.

Both are lifted here. The subscriber is now installed exactly once, independently of `cass_log_set_level`, and its filter is a *reloadable* `LevelFilter` whose handle lives in a static. Setting the level is then an ordinary mutation, valid at any point of the driver's lifetime and repeatable.

Two related bugs are fixed on the way, and both were called out in #498:

3. `CASS_LOG_DISABLED` never disabled anything. It only logged "Logging is disabled!" and returned. It appeared to work when it happened to be the very first call - no subscriber was installed then, so nothing was logged anyway - but after any other level had been set, logging just kept flowing at that level.
4. The driver reacted to `RUST_LOG`. `EnvFilter::from_default_env()` meant a C/C++ application's log level could be silently overridden by an environment variable of a foreign ecosystem.

### Lazy initialization from entry points, not a library constructor

To lift limitation 1, the subscriber must be up before the first log event. The obvious tool is a `#[ctor]`-style library constructor running at `dlopen` time, but I chose a `LazyLock` forced from `init_logging()`, called at the top of the entry points that can plausibly be the first `cass_*` function an application calls: `cass_cluster_new`, `cass_session_new`, `cass_ssl_new_no_lib_init` (and thus `cass_ssl_new`) and `cass_log_set_callback`. Reasons:

- no new dependency;
- the crate is also built as a `staticlib`, where a linker is free to garbage-collect an `.init_array` entry that nothing references;
- it composes with the crate's own unit tests, which install their own subscriber via `setup_tracing()`. A constructor would always win that race and silently take logging away from every existing test.

`init_logging()` is a `LazyLock::force` - one atomic load after the first call.

### `cass_log_set_callback` initializes logging too

An application that only installs a callback and relies on the default level now receives events. Previously it received nothing at all, because nothing had installed a subscriber.

### A foreign global subscriber is respected

If a subscriber is already installed globally - by the application embedding the driver, or by a unit test - the reload handle is `None` and `cass_log_set_level` becomes a no-op, rather than pretending to reconfigure logging that isn't the driver's to reconfigure.

### `RUST_LOG` support is dropped deliberately

Point 3 of #498. The driver's own subscriber filters with a plain `LevelFilter`. The `env-filter` feature is still enabled because the unit tests' `setup_tracing()` uses it.

### The default level is now actually WARN

`include/cassandra.h` has always documented `CASS_LOG_WARN` as the default. Until now that was only true in the sense that nothing was logged at all before the first `cass_log_set_level` call. Now `DEFAULT_LOG_LEVEL_FILTER` makes it true.

## Testing

- A `rusty_fork` unit test in `logging.rs` covering both lifted limitations: an event emitted before `cass_log_set_level` was ever called reaches the callback, and the level then follows repeated calls - lowered, raised, disabled, enabled back. Events are emitted through closures, so every level change is exercised against callsites already used under the previous level; this makes the test cover invalidation of `tracing`'s per-callsite interest cache, which would otherwise keep a callsite filtered out after a level change. A fresh process is needed because the subscriber is global and installable once, hence   `rusty_fork`.
- `LoggingTests.SetLevelIsMutable` in `test_logging.cpp`. `cass_log_set_level` emits a DEBUG event confirming the level it just set, which makes it its own log source: the callback fires iff the new level admits DEBUG. That is enough to test the whole sequence without a running cluster. Before the fix it fails at the very first level change.
- Both are picked up by CI with no filter changes: `LoggingTests.*` is already listed as a wildcard in both `SCYLLA_TEST_FILTER` and `CASSANDRA_TEST_FILTER`, and the unit test is not a `ccm` test, so `make run-test-unit` runs it.

## Drive-by

`LoggingTests.Callback` had a data race, in a separate commit: the driver emits log events from its own background threads, so the `bool` flag its callback sets was written from those threads while the test thread read it, with no synchronisation. It is now atomic, as is the new test's counter.

## Pre-review checklist

- [x] I added relevant tests for new features and bug fixes.
- [x] All commits compile, pass static checks and pass test.
- ~~[ ] I have provided docstrings for the public items that I want to introduce.~~ (no new public items)
- [x] I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/scylla-rust-wrapper/src/logging.rs
+++ b/scylla-rust-wrapper/src/logging.rs
@@ -271,3 +271,101 @@ pub unsafe extern "C" fn cass_log_get_callback_and_data(
         *data_out = logger.data;
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::os::raw::c_void;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    use rusty_fork::rusty_fork_test;
+    use tracing::{error, info, warn};
+
+    use crate::argconv::{CConst, CassBorrowedSharedPtr};
+    use crate::cass_log_types::{CassLogLevel, CassLogMessage};
+
+    use super::{cass_log_set_callback, cass_log_set_level};
+
+    /// Counts the log events that made it through the driver's filter
+    /// to the user-provided log callback.
+    #[derive(Default)]
+    struct EventCounter(AtomicUsize);
+
+    impl EventCounter {
+        /// Returns the number of events captured since the previous call.
+        fn take(&self) -> usize {
+            self.0.swap(0, Ordering::SeqCst)
+        }
+    }
+
+    unsafe extern "C" fn counting_log_callback(
+        _message: CassBorrowedSharedPtr<CassLogMessage, CConst>,
+        data: *mut c_void,
+    ) {
+        let counter = unsafe { &*(data as *const EventCounter) };
+        counter.0.fetch_add(1, Ordering::SeqCst);
+    }
+
+    rusty_fork_test! {
+        #[test]
+        /// Verifies that the log level can be set at any point of the driver's
+        /// lifetime, any number of times.
+        ///
+        /// This is run with rusty_fork in order to have a fresh process. The
+        /// tracing subscriber is global and can only be installed once, so the
+        /// test must neither share a process with other tests (which install
+        /// their own subscriber via `setup_tracing`) nor with another run of
+        /// itself. For the same reason the test must not call `setup_tracing`.
+        fn log_level_is_mutable() {
+            let counter = EventCounter::default();
+            let data = std::ptr::from_ref(&counter).cast::<c_void>().cast_mut();
+
+            // Each of these emits from a single tracing callsite, no matter how
+            // many times it is called. Reusing them across log level changes is
+            // what makes this test cover the invalidation of tracing's interest
+            // cache, which memoizes per-callsite whether anyone is interested
+            // in its events.
+            let emit_info = || info!("An INFO event");
+            let emit_warn = || warn!("A WARN event");
+            let emit_error = || error!("An ERROR event");
+
+            // An event emitted before any `cass_log_set_level` call already
+            // reaches the callback, at the default (WARN) level...
+            unsafe { cass_log_set_callback(Some(counting_log_callback), data) };
+            emit_warn();
+            assert_eq!(counter.take(), 1);
+
+            // ...and one below that level does not.
+            emit_info();
+            assert_eq!(counter.take(), 0);
+
+            // Lowering the level admits more events, including from callsites
+            // whose events were already filtered out above.
+            unsafe { cass_log_set_level(CassLogLevel::CASS_LOG_TRACE) };
+            counter.take(); // Discard the driver's own log about the change.
+            emit_info();
+            assert_eq!(counter.take(), 1);
+
+            // Raising it filters them out again - the level is mutable both
+            // ways, also for callsites that were admitted a moment ago.
+            unsafe { cass_log_set_level(CassLogLevel::CASS_LOG_ERROR) };
+            counter.take();
+            emit_info();
+            emit_warn();
+            assert_eq!(counter.take(), 0);
+            emit_error();
+            assert_eq!(counter.take(), 1);
+
+            // Logging can be disabled entirely...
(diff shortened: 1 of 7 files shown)


scylla-rust-wrapper/src/logging.rs
```diff
@@ -271,3 +271,101 @@ pub unsafe extern "C" fn cass_log_get_callback_and_data(
...
+            emit_error();
+            assert_eq!(counter.take(), 0);
+
+            // ...and enabled back again.
+            unsafe { cass_log_set_level(CassLogLevel::CASS_LOG_WARN) };
+            counter.take();
+            emit_warn();
+            assert_eq!(counter.take(), 1);
+        }
+    }
```
Lorak-mmk, line 481: In the driver, we have 2 main "categories" of logs / spans.
 - Connected to the user requests
 - Connected to long-running driver tasks (connection, connection pools, workers)
 
For user-request stuff, it doesn't matter a lot if log level changes are applied to already-running requests, or just new requests. What is important is that for new requests `span.is_enabled()` works correctly, because we guard allocating replica listing behind it. Can you add a test for that?

For long-running tasks everything seems more difficult. Spans will be long-lived. Will log-changing work in that case? Can you add a test for that?

wprzytula: I added two more unit tests illustrating that logs (_events_ in the `tracing` glossary) truly react to level changes.

Important: **span's enabledness is immutable**. We can do nothing about this limitation, which is inherent to `tracing`. This is not a problem if we follow the `tracing`'s design, which says that spans should be used for units of work, not for a potentially eternal process. Currently, only short-lived tasks in the Rust Driver are covered by spans, which agrees with the design.

Lorak-mmk: > This is not a problem if we follow the tracing's design, which says that spans should be used for units of work, not for a potentially eternal process. 

Could you point me to that? I was thinking we should have e.g. a span for ClusterWorker, another for MetadataWorker (with sub-span for single CC). Are you saying this is wrong? If so, how should events from such tasks be grouped together?

wprzytula: > > This is not a problem if we follow the tracing's design, which says that spans should be used for units of work, not for a potentially eternal process.
> 
> Could you point me to that?

It seems to be my misinterpretation.

> I was thinking we should have e.g. a span for ClusterWorker, another for MetadataWorker (with sub-span for single CC). Are you saying this is wrong? If so, how should events from such tasks be grouped together?

You're right that in the future we may want to emit such spans. Hence, the newest commit clearly mentions that alterations of the log level are considered experimental and might misbehave.

Lorak-mmk (approved):

